PR #4, opened 2019-07-18
Remove old GNUmakefiles

mathomp4:


mathomp4:
Note that I restored the GNUmakefiles for `make_bcs` and `make_tiles` as they are weird little sub-programs. Now, they seem to need MAPL source to do their thing, but this is knowledge not to lose for now.

PR #14, opened 2019-08-01
Bugfix/kgerheis/#13-fv_tracer2d-ksplt

### LINKED ISSUE #13: Incorrect comparison to global 'nsplt' instead of local ksplt(k) in fv_tracer_2d
I believe I found a bug in fv_tracer2d.F90 when filling the optional `dpA` in the subroutine `tracer_2d`. A comparison is made to the global `nsplt` when it should be compared to the local `ksplt(k)` value. This leads to `dpA` not being fully filled if `ksplt` for that level is less than `nsplt`

This:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/f5fd1e30f4c6ee376c6f799b32cb71e606335dd4/GEOSagcm_GridComp/GEOSsuperdyn_GridComp/FVdycoreCubed_GridComp/fvdycore/model/fv_tracer2d.F90#L546

Should be:
```fortran
if ( it /= ksplt(k)) then 
```

because if `ksplt(k) == 1` for that level, and `nsplt == 2`, `it` will never be 2 because the outer `if ( it .le. ksplt(k) )` will fail and therefore dpA won't be assigned.

kgerheiser:
Fixes #13 

Compare to ksplt(k) instead of nsplt so that dpA is properly assigned

### DIFF
--- a/GEOSagcm_GridComp/GEOSsuperdyn_GridComp/FVdycoreCubed_GridComp/fvdycore/model/fv_tracer2d.F90
+++ b/GEOSagcm_GridComp/GEOSsuperdyn_GridComp/FVdycoreCubed_GridComp/fvdycore/model/fv_tracer2d.F90
@@ -543,7 +543,7 @@ subroutine tracer_2d(q, dp1, mfx, mfy, cx, cy, gridstruct, bd, domain, npx, npy,
                enddo
             enddo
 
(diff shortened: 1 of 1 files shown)


mathomp4:
@wmputman Am I right in thinking this change will *not* affect the way GEOS utilizes this code?

PR #22, opened 2019-08-08
Updating Moist and FV3 for NH and GFDL-MP

wmputman:


mathomp4:
@wmputman Do your new changes to the FV3 `CMakeLists.txt` change the hydrostatic results? If so, I can't see how we can pull this into `develop` without making the default of the whole model non-hydrostatic. 

The two moist changes (if on a separate pull) look like they could be fairly safe as they seem to GFDL Moist only updates. But has anyone tested the effect of `-DUSE_COND -DMOIST_CAPPA` on hydrostatic runs?

mathomp4:
I've dismissed the @tclune approval as I'm not sure the FV3 `CMakeLists.txt` change is safe. @wmputman does that change (`-DUSE_COND -DMOIST_CAPPA`) affect the model when run in hydrostatic mode?

wmputman:
These changes are not are for the hydrostatic core. They need to wait until we always run NH.

PR #23, opened 2019-08-09
bugfix/narnold1/#32-fix-triangle-pdf

narnold1:
Fixed triangular pdf option.  Since triangular pdf isn't used by default, this should be 0-diff to develop.


mathomp4:
@narnold1 We'll wait for @wmputman to approve and then we'll try and get this in once 10.3.2 (Jason 3.2) is released.

PR #28, opened 2019-08-14
Fixes #27 - register grids from AdvCore

### LINKED ISSUE #27: AdvCore_GridComp does not register its grids
AdvCore needs to register its grid like DynCore does

kgerheiser:
Move register_grid_and_regridders to its own module to avoid duplication and call it from AdvCore

### DIFF
--- a/GEOSagcm_GridComp/GEOSsuperdyn_GridComp/FVdycoreCubed_GridComp/AdvCore_GridCompMod.F90
+++ b/GEOSagcm_GridComp/GEOSsuperdyn_GridComp/FVdycoreCubed_GridComp/AdvCore_GridCompMod.F90
@@ -67,6 +67,7 @@ module AdvCore_GridCompMod
 
       USE FV_StateMod,     only: AdvCoreTracers => T_TRACERS
       USE FV_StateMod,     only: FV_Atm
+      use CubeGridPrototype, only: register_grid_and_regridders
 
       implicit none
       private
@@ -316,6 +317,7 @@ contains
       !--------------------------------------------------
       if (.NOT. FV3_DynCoreIsRunning) then
          call fv_init2(FV_Atm, dt, grids_on_my_pe, p_split)
+         call register_grid_and_regridders()
       end if
 
       ! Ending with a Generic SetServices call is a MAPL requirement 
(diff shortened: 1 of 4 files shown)


mathomp4:
I think this is indeed a @bena-nasa request. :smile:

bena-nasa:
@sdrabenh This should be a trivial zero-diff change.

sdrabenh:
@kgerheiser please double check this code again. Maybe I am doing something incorrect, but every time I run your executable it seg faults.

tclune:
@sdrabenh Failures for you are odd given that this code only affects the advect core which I don't think is exercised in your usual tests.    But I agree first step is for @kgerheiser to double check a standard build on his end.  If that still works for him the SI team will look deeper into the issue.

sdrabenh:
I agree it is odd. I built this two different times and could never run it without seg fault. I have not had this issue with any other pull requests yet.

tclune:
Note: Whoever submits the pull request for this should also fix the various declarations (e.g., temp2d) that are declared with KIND=4.   This is not standard conforming.   E.g., with the  NAG Fortran compiler  KIND=1 is 32 bit real with the default settings and can be something rather weird with more aggressive debug flags.

kgerheiser:
I found the problem and fixed it. Somehow I missed part of the register_grid_and_regridders subroutine where it added the prototype to the GridManager.

sdrabenh:
@kgerheiser has the push been updated? I see the same 4 files but can't tell if your most recent changes are included. Also, can you please verify this is zero diff?

tclune:
@sdrabenh It can be a bit subtle to see the subsequent changes during an ongoing PR.   But you should see two "View Changes" boxes on the RHS of the thread above.     The second one was just before Kyle posted his claim of fixing the problem.   That one has changes in only one file.

tclune:
@kgerheiser   Now that we finally merged the change that puts the dycore into its own repository, you are going to need to redo this PR in 

https://github.com/GEOS-ESM/FVdycoreCubed_GridComp

or maybe even in 

https://github.com/GEOS-ESM/GFDL_atmos_cubed_sphere

depending on where the files are.

Sorry about that.   (Note that I did bring in your other hotfix though, so we should be out-of-the-woods on further complications.)

PR #32, opened 2019-08-15
Feature/xanadu fms mom6

sanAkel:
@aerorahul, @mathomp4 I added back whatever stuff was in the GuestOcean_GridComp so that it can work with both versions of MOM (5 and 6). 

@aerorahul, as we discussed, when you have your mods for CMakeLists.txt, we should have both MOM5 and MOM6, built and at run time the user can choose and set "OCEAN_MODEL_NAME: " to be "DATA_OCEAN" (default) or "MOM5" or "MOM6." 
** I went ahead to issue this PR, because in future, I would like to combine these different "models" and clean this stuff up! 

@mathomp4 With these changes, I don't see a reason, why the GuestOcean_GridComp would break for anyone! Could you please consider merging!


aerorahul:
@sanakel this PR will break mom5 compatibility and should not be merged.

sanAkel:
@aerorahul, May be... 

but once your mods for CMakeLists is done, IT SHOULD NOT!! 

If I need to wait for your mods, okay, fine, I'll close the PR.

sanAkel:
revisit after getting changes from @aerorahul

PR #42, opened 2019-08-19
Fixes #35. Applied the same logic as in the fix for Saulo. This might…

### LINKED ISSUE #35: Ordering inside MTR bundle
The fields out of the MTR bundle come in "wrong" order and this affects the Henry's law constants. This can be fixed easily by fetching field by name instead of by index

atrayano:
… be non-zero diff change, due to fixing the order fetches the correct Henry's law constants, and possibly different sumation order.

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/GEOS_MoistGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/GEOS_MoistGridComp.F90
@@ -6518,11 +6518,11 @@ contains
          ! Get the Kth Field from tracer bundle
          !-------------------------------------
 
-         call ESMF_FieldBundleGet(TR, K, FIELD, RC=STATUS)
+         NAME=QNAMES(K)
+
+         call ESMF_FieldBundleGet(TR, fieldName=NAME, Field=FIELD, RC=STATUS)
          VERIFY_(STATUS)
 
-         call ESMF_FieldGet(FIELD, name=NAME, RC=STATUS)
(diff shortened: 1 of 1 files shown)


sdrabenh:
@atrayano I found this was a zero-diff change using the default configuration

sdrabenh:
@tclune and @mathomp4, please remind me about these hot fixes. Do I merge this into develop too now that is was merged into master?

tclune:
Ideally, yes.   The alternative is that the original requester notices the approval and remembers to propagate to develop with another PR. 

Note: no testing should be required for these changes.    Yes in theory there could be some odd conflict between subsequent PR's that have gotten onto the develop branch, but that should be very rare.   And even then, that just means that the develop branch needs to be fixed.  If it is on master then it must get to develop (eventually).

@atrayano raised the issue that you may not be able to **merge** your own PR's.   I suspect it is just that you cannot **approve** your own PR's.    If you cannot merge, then we need a different option.  Otherwise my preference is that you immediately submit a PR and and find someone to do a hasty, half-hearted review.

sdrabenh:
@mathomp4 is there an easy way to recycle this pull request on github - for example just changing which branch (develop vs master) we are merging into?

tclune:
I don't think there is a way, and even if there was, it would probably be a bad way to do it, as in theory it could be multiple PR's that have accumulated and are now propagating down to develop.  

What you could do is reference the original issue # in the new pull request.   The actual commit in the original pull request will already be in the commit set of the new pull request, so there is not much point in referencing it by number.

I'd just mention in the subject line that it is propagating a hot fix from master - that should be enough info for forensics down the road.

I may be naive here.  Later we may decide a more formal process is needed for this.

PR #44, opened 2019-08-19
3 updates: (1) PEATLAND data processing, (2) global country codes and…

smahanam:
… US state codes, and (3) MODIS VISDF and NIRDF processing for CATCHCN


JustinPerket:
Some of these will be hard to test. Per conversations with @bena-nasa this week, neither `GEOSsurface_GridComp/Shared/Raster` nor `GEOSsurface_GridComp/Shared/Raster/src` will currently build in git because of the GNUmake files.

There also seems to be at least one non-functioning bit, and I have a couple questions on others. 
Perhaps the focus should be on getting `GEOSsurface_GridComp/Shared/Raster` before this merge. In the meantime, we can also discuss offline these changes.

PR #50, opened 2019-08-23
Fixes #49 split fvdycore repo

### LINKED ISSUE #49: Split FVdycoreCubed_GridComp into sepratate Git repo
This is to support running FV standalone in a virtual machine, but will also facilitate various CTM applications that do not need the full GEOS GCM.

tclune:
This should be a zero diff change - just changes the home for FV.


tclune:
Double checked and something is broken in the build.

PR #54, opened 2019-08-26
Fixes #53 - initialize uc and vc to 0 before assigning the non-halo part

### LINKED ISSUE #53: Crash in fv_computeMassFluxes_r8 due to uninitialized halo values
`fv_computeMassFluxes_r8` calls `compute_utvt` with `uc` and `vc` that have uninitialized halos which causes the code to crash. The solution is to set them to 0 before assigning the non-halo part, and the r4 version of this subroutine already does that.

kgerheiser:
The r4 version of the code already has this, so add it to the r8 version.


mathomp4:
I'm assuming this PR should be closed and re-opened in the new FVdycoreCubed_GridComp repository?

kgerheiser:
Yes

PR #83, opened 2019-09-19
Fixes #82. This change should enable reading netCDF increments during…

### LINKED ISSUE #82: Exact replay cannot read netCDF increments
Currently, the only supported type of increment files is 'binary'. We should add ability to read netCDF increments

atrayano:
… EXACT replay


atrayano:
@lltakacs this is the change to enable reading netCDF increments during EXACT replays

PR #117, opened 2019-10-11
should keep outFmt open for writing

weiyuan-jiang:


JustinPerket:
seems to be a trivial change to me

PR #92, opened 2019-09-27
After regridding land prognostics, we adjust GHCNT to ensure deep soi…

smahanam:
**PURPOSE**
It was noted that the current regridding tool has a problem. If porosity of at an output tile is different to the porosity of the input tile that provided restart prognostics, there would be an inconsistency in deep soil temperature and ground heat content at the very beginning. Scale_Catch.F90 was modified to ensure deep soil temperature is preserved by scaling GHTCNT to match changing porosity.

PR #137, opened 2019-11-13
Bugfix/sm make bcs update

JustinPerket:
@smahanam fixed make_bcs utilitiy to work with cmake.

Despite `bugfix` name, also adds a zero-diff feature.  **PEATMAP data processing**. Disabled by default with logical parameter process_peat=.false.  

Other changes: 

- values of pi changed to MAPL_PI
- 8 more climatological (fPAR, VISDF, NIRDF stats for CATCHCN) data files and new clsm/country_and _state_code.data file. 
- Now an extra column on soil_param.* files 
This branch produced zero-diff BCs with NLv3 (checked CF90atmos-CF90atmos and M36 SMAP grid)

This addresses https://github.com/GEOS-ESM/GEOSgcm/issues/42


GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/Raster/mkLatLonRaster.F90
```diff
@@ -1,4 +1,4 @@
-!   $Id$
+!   $Id:
```
JustinPerket:
```suggestion
```

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/Raster/mkMITAquaRaster.F90
```diff
@@ -1,6 +1,4 @@
-
-
-!  $Id$
+!  $Id:
```
JustinPerket:
```suggestion
```

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/Raster/mkMOMAquaRaster.F90
```diff
@@ -1,5 +1,4 @@
-
-!   $Id$
+!   $Id:
```
JustinPerket:
```suggestion
```

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/Raster/rasterize.H
```diff
@@ -1,5 +1,5 @@
 
-!  $Id$
+!  $Id: rasterize.H,v 1.3 2011/01/19 14:00:05 msuarez Exp $
```
JustinPerket:
```suggestion
```

JustinPerket:
As @smahanam points out, I commented on the issue page, and not this PR:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/92#issuecomment-553459018
> build fails with:
> 
> ```
> root/project/GEOSgcm/src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/Raster/zip.c:4:10: fatal error: zlib.h: No such file or directory
>  #include "zlib.h"
> ```
> 
> presumably an easy fix

mathomp4:
@JustinPerket @smahanam Interesting. Baselibs should definitely have `zlib.h` but its possible we don't handle C code as well as Fortran code. 

I'll try building myself and if that works, it points to something odd in the containers @pchakraborty uses, but I'm betting on CMake...

mathomp4:
Okay. I might know what's happening. I need to consult @tclune and @pchakraborty on the best way to solve it. There's a cheap-and-easy way to solve and a more correct way. Hmm.

JustinPerket:
interesting, thanks @mathomp4

mathomp4:
@JustinPerket @smahanam I think I have a fix. I'm going to push the CMakeLists.txt on your branch because, unfortunately, the only way to test this is to have CircleCI launch a new test.

smahanam:
I see @mathomp4 it is this automated CircleCI that has this issue. I just built this branch through GEOSldas too.

mathomp4:
I now found a bug with gfortran being strict pushing again.

mathomp4:
Oh dear. This next error I think means we need to use GCC 9 for the CI testing. @tclune might know more, but I think `findloc()` was a Fortran 2008 function that wasn't supported in GCC until 9.

smahanam:
My apologies if this has caused a problem. I can work around the code as well if it is more convenient.
I can replace it with  a loop.

mathomp4:
> My apologies if this has caused a problem. I can work around the code as well if it is more convenient.
> I can replace it with a loop.

@smahanam Let me talk with @tclune about this. `findloc()` is valid Fortran and it's only due to an old compiler this is happening. I like `findloc()` save for the fact it returns an array. Did some tests and it looks like it's as fast or faster than a brute force loop and definitely more readable.

mathomp4:
I think I have a fix for this. It'll first require some changes go into `ESMA_cmake` that needs @sdrabenh to then update `Develop.cfg`. After that, I can then add my changes here and that should let it pass. Phew. Getting close!

smahanam:
Thank you @mathomp4 ! As I understand we are not going backward, right? We don't have to go backward to accommodate FINDLOC, I can fix my code instead.

mathomp4:
No, the way it will be is CMake will test if a compiler supports FINDLOC(). If not, we’ll compile a new source file, findloc.F90, and then set an ifdef to use the findloc() in there. That new file just has a native brute force loop version of findloc() that Intel should never see.

smahanam:
@zhaobin74 @atrayano @yvikhlya 

I added Atanas, Bin Zhao and Yuri to this PR request since we all contribute to or use this Shared/Raster directory where make_bcs is run.

mathomp4:
Okay. I've pushed a new file: `findloc.F90` that has a *very limited* version of `findloc()`, the one-d `findloc(array,value)` case. Now if an older compiler doesn't support `findloc()` it will build and use this.

NOTE: To get this functionality, you need ESMA_cmake v1.0.11. @sdrabenh changed `Develop.cfg` in the `GEOSgcm` to use this. Hopefully the CircleCI will let this build now.

JustinPerket:
> Hopefully the CircleCI will let this build now.
looks like it does. Thanks @mathomp4 !

smahanam:
So CircleCI  passed.
Yes Thanks so much @mathomp4

JustinPerket:
I think this is ready for review then. Will contact @sdrabenh and @wmputman

tclune:
Bonus points go to the person that remembers to delete the CMake test (and test program) in some future era when all of the compilers support ```findloc()``.

mathomp4:
Per @smahanam, who discovered that my `findloc` was insanely incorrect in the sense that it always returned 0.

The odder question is *how* @smahanam found this out. Intel shouldn't have triggered this code since it supports the intrinsic `findloc()`. Odd.

PR #141, opened 2019-11-21
Fixes #135 - incorrect use of pointers

tclune:
Each of 3 distinct data structures were unintentionally sharing a 1D
array accross 33 Fortran pointers.  The fix was to directly allocate
each pointer and eliminate the fixed intermediary.

Because "33" was a magic number throughout the source, a new integer
parameter NUM_TAUA was introduced.

This component is not exercised by the usual tests and thus all that
has been verified is that the new version compiles.  But should be 0-diff for usual tests.


tclune:
ok - was not sure if the component was "that broken" in its current state.

atrayano:
Santha is correct - the component was broken since nobody implemented the skin layer scheme

On Dec 2, 2019, at 11:56 AM, Tom Clune <notifications@github.com<mailto:notifications@github.com>> wrote:


ok - was not sure if the component was "that broken" in its current state.

PR #162, opened 2019-12-19
clarified LONG_NAME of "snow depth" variable

gmao-rreichle:
Trying to correct a long-standing problem with users misunderstanding of the GEOS snow mass and snow depth output.   See "snow depth" entry in MERRA-2 FAQ:
[https://gmao.gsfc.nasa.gov/reanalysis/MERRA-2/FAQ/#](url)
This has been an issue for years, with multiple inquiries about "funny" MERRA-2 output.  This past summer, we narrowly avoided two peer-reviewed papers by ECMWF and Meteo-France showing very unhelpful MERRA-2 results.  Ask me if you want to know more.
By fixing the F90 src code, we have at least a chance of getting this critical information into new versions of GEOS file specs documents.
Similar edits may be good to have for snow depth over glaciated surfaces, which is why I added @lcandre2 and @zhaobin74 to the list of reviewers.  (Richard Cullather isn't on Github.)

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
@@ -1361,7 +1361,7 @@ module GEOS_SurfaceGridCompMod
     VERIFY_(STATUS)
 
      call MAPL_AddExportSpec(GC                             ,&
-        LONG_NAME          = 'snow_depth'                        ,&
+        LONG_NAME          = 'snow_depth_within_snow_covered_area_fraction'   ,&
         UNITS              = 'm'                                 ,&
         SHORT_NAME         = 'SNOWDP'                            ,&
         DIMS               = MAPL_DimsHorzOnly                   ,&
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
@@ -1872,7 +1872,7 @@ subroutine SetServices ( GC, RC )
   VERIFY_(STATUS)
 
   call MAPL_AddExportSpec(GC,                    &
-    LONG_NAME          = 'snow_depth_in_snow_covered_area' ,&
+    LONG_NAME          = 'snow_depth_within_snow_covered_area_fraction'   ,&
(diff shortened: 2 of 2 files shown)


wmputman:
Let's verify with Rob what impacts this has on file-specs in GEOS-FP before approving.

gmao-rreichle:
For the record, I'm adding an email from Matt to the discussion below.  

In response, without knowing much at all about CF conventions, I think Matt's email is head-on.  My only comment would be that the "snow depth" from the GEOS "lnd" Collection is the mean over "snow-on-land" only, and I don't know how that would be expressed in CF-compliant language.  (That is, the mean excludes the zero snow depth values over the snow-free land portion and also excludes non-land areas such as sea-ice, regardless of whether they are snow-covered.)

For now, I think we should still go ahead with the PR and not wait for the big switch to CF compliance.

---------------------------------------

**Snow Depth and CF Conventions**
From: Thompson, Matt
Thu 19-Dec-19 10:48 AM
To: Reichle, Rolf 

Rolf,

This is a random question, but your PR about the snow depth change made me go on a tangent to CF. I did that because one of my (perhaps "our") long standing wants is to convert GEOS to use CF Convention standard_names instead of using short_name everywhere (and where long_name is the actual human-readable description). 

So I wondered how would one make your change under CF? This is something we might have to think about soon-ish.

I think/hope I'm safe saying that our "snow_depth" would be "surface_snow_thickness" in the conventions table. But then the help for that says:
The surface called "surface" means the lower boundary of the atmosphere. Unless indicated in the cell_methods attribute, a quantity is assumed to apply to the whole area of each horizontal grid box. Previously, the qualifier where_type was used to specify that the quantity applies only to the part of the grid box of the named type. Names containing the where_type qualifier are deprecated and newly created data should use the cell_methods attribute to indicate the horizontal area to which the quantity applies.
Thus, your change made me go on the cell_methods tangent. I stared at:

[http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#statistics-applying-portions](url)

and wondered would this then need to have a cell_method of:

  surface_snow_thickness:cell_methods="area: METHOD where snow"

where "METHOD" is the right entry from here:

[http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#appendix-cell-methods](url)

mean, median, whatever the code does.

Does this track with your understanding? I'm just trying to think of all the bits and pieces Atanas/Ben/Tom might need to implement in MAPL in the future (if we go the CF route).

Matt

gmao-rreichle:
Should be ready for approval now.

Adding Rob's email below for the record:

Date: Tue, 24 Dec 2019 00:14:36
From: Rob Lucchesi
Subject: Re: action items re. mis-interpretation of MERRA-2 (and GEOS) snow output

Rolf, Bill,
[...]  Changing the long-name should not be a big deal.  Is the intent
to have this in for the current FPP going operational after the 1st of
the year?  I think it is reasonable to just mention the change in the
release notes for the system.
[...]
Thanks, Rob

PR #170, opened 2020-01-08
Feature/pnorris/#197 chou suarez longwave all sky no aerosol bugfix

dr0cloud:
Pete Colarco identified an issue with the all-sky no-aerosol fluxes from the Chou-Suarez longwave code. There is erroneous cloud contamination in OLR-OLRNA (aerosol forcing). Peter Norris has tracked down the bug is pushing this bugfix to the develop branch. It will be "zero-diff" in the sense that Chou-Suarez is no longer the default.


mathomp4:
Mistakenly from MAPL 2.0 codebase.

PR #171, opened 2020-01-08
pmn: bugfix for all-sky-no-aerosol longwave fluxes under Chou-Suarez

dr0cloud:
Fixes issue #172 
Pete Colarco identified an issue with the all-sky no-aerosol fluxes from the Chou-Suarez longwave code. There is erroneous cloud contamination in OLR-OLRNA (aerosol forcing). Peter Norris has tracked down the bug is pushing this bugfix to the develop branch. It will be "zero-diff" in the sense that Chou-Suarez is no longer the default.


dr0cloud:
Note - original issue (197) mistakenly put under MAPL. Transferred it to GEOSgcm and it got the current new issue number (172). Unfortunately the branch name references the old issue number. Oh well...

PR #187, opened 2020-01-23
Online calculation of lightning flash rate for GEOSCHEMchem

christophkeller:
Compute the lightning flash rate (LFR) online following the approach used by GEOSCHEMchem_GridComp. The LFR is calculated in MOIST and then made available to other components as a new export field (LFR_GCC). This update does not touch the original LFR calculation in MOIST (using Dale Allen's method). 

This pull request is contingent on [adding the LFR module to GMAO_Shared](https://github.com/GEOS-ESM/GMAO_Shared/pull/69). It only adds a new diagnostics to MOIST, hence should be a zero-diff change.


GEOSagcm_GridComp/GEOSphysics_GridComp/GEOS_PhysicsGridComp.F90
```diff
@@ -1244,7 +1244,7 @@ subroutine SetServices ( GC, RC )
     VERIFY_(STATUS)
 
     call MAPL_AddConnectivity ( GC,                                &
-         SHORT_NAME  = (/'TS'/),                                   &
+         SHORT_NAME  = (/'TS', 'LWI'/),                            &
```
mathomp4: @christophkeller I know from past experience this will cause a failure with GNU compilers. Can you push a change so that the two members of the array constructor have the same length:
```fortran
SHORT_NAME  = (/'TS ', 'LWI'/)
```

christophkeller: Done!

sdrabenh:
@christophkeller , I tested your pull requests and found they are 0-diff except for adding 'LWI' to moist. However, after discussing with @lltakacs , the LWI is a legacy field which has been phased out of history. Instead, the more accurate and recommended approach is to use the existing fractional areas:

                              'FRLAKE'    , 'SURFACE'       ,
                              'FRLAND'    , 'SURFACE'       ,
                              'FRLANDICE' , 'SURFACE'       ,
                              'FROCEAN'   , 'SURFACE'       ,

Would it be possible to adapt your code to use those existing variables and not to re-introduce a legacy variable?

christophkeller:
Removed LWI import from MOIST and instead calculate it from FROCEAN, FRLAKE, FRACI, and TS. The definition is identical to the original LWI formulation.

christophkeller:
Now that [PR69 in GMAO_Shared](https://github.com/GEOS-ESM/GMAO_Shared/pull/69) is closed, can we move forward with this pull request?
Thanks!

mathomp4:
@christophkeller The branch was out-of-date with `develop` so I clicked the update button. Saves @sdrabenh a step.

(And, I guess, let's me keep my lead on commits in the org 😄 )

sdrabenh:
@christophkeller thanks for bringing this PR to my attention again. I ran the 0-diff tests on your code, and it passed with the default settings. However, I noticed your code initializes in moist but uses the fractions from surface (FRLAKE, FRACI, etc). Perhaps @mathomp4 can correct me, but I believe the set services for moist are done before surface. Therefore, I presume all these fractions would be bootstrapped to zero, at least for the first model step. Is this the desired behavior? I have not tested activating your lightning module - have your tested running this under the new MAPL-2?

christophkeller:
@sdrabenh thanks for bringing this up, this is a good point. Wouldn't these fields then go into the moist import restart file, though?
If all fractions are zero (because they are bootstrapped), LWI would become 1.0 (= land) in the original definition. This is not ideal, since the flash rates over land are much higher than over the ocean. I just changed the parameterization - now using FRLAND - to make water (0.0) the default value. This should prevent unrealistically high global flash rates if the values are bootstrapped. Also, FRLAND is already in the moist import restart file, so there should be less of an issue here.

mathomp4:
@christophkeller have you made sure/tested this with the GEOS gcm_regress test? Does it pass a start/stop in time?

christophkeller:
Yes, the update passes time regression. As expected, FRACI gets added to the moist import restart file, hence zero-diff.

Issue #145, opened 2019-12-05
GEOS_LandGridComp.rc


smahanam:
Land models need a separate resource file to keep increasing land physics parameters,

PR #208, opened 2020-02-06
Feature/sm land gcrc 145

smahanam:
This PR deals with issue [145](https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/145)
There are related PRs in GEOSgcm_App and GEOSldas using the same branch name "SM_LandGCRC_145"


gmao-rreichle:
@smahanam @sdrabenh @weiyuan-jiang 
If I remember correctly, this GCM GridComp pull request is in preparation for a suite of individual pull requests by Sarith that will add new science features to the GCM GridComp (incl dealing w/ solid precip, runoff tuning, peatland physics, permafrost, etc).
Before we can go ahead and merge the pull request into the GCM GridComp, we need to run the suite of GEOSldas tests.
@weiyuan-jiang : Can you please run these tests, or support @smahanam in running the tests?

weiyuan-jiang:
I have merged develop into this branch and tested agains GEOSldas. it passed 0-diff test

gmao-rreichle:
It's good news that we passed the GEOSldas tests per @weiyuan-jiang 's comment.  I do have a few questions for @smahanam and @sdrabenh : 

1) @smahanam : The Plant-functional-type lookup table ("data" statements) in GEOS_CatchCNGridComp.F90 (new line 8974...) have changed.  How can this be zero-diff for CatchCN?

2) @sdrabenh : The "z0_formulation" default value in GEOS_CatchGridComp.F90 and GEOS_CatchCNGridComp.F90 is set to 2.  In the GEOS FP upgrade to f525_fp the default value should have been changed to 4.  Was this missed in the git tag?  Or possibly even in the ops CVS tag?  This might require input from @lltakacs .

3) @smahanam : In GEOS_CatchCNGridComp.F90, it looks like the interface to subroutine catch_calc_soil_moist() has changed (additional "bf1" and "bf2" in the calling statement).  The interface is unchanged everywhere else, incl lsm_routines.F90 where the subroutine lives.  Because of the optional arguments it's conceivable that everything compiles, but it's not obvious how running CatchCN could be zero-diff.  

I am a bit rushed today and am probably missing something, but please double-check the above.

smahanam:
AGCM does not perform zero-diff on CATCHCN, as far as I know. CATCHCN still still under development. I removed bf1, and bf2 from call catch_calc_

The default CATCH passed zero-diff tests as spelled out by the CI-Team.

gmao-rreichle:
@sdrabenh @smahanam @mathomp4 : I think it's time to add a CatchCN test case to the GCM suite of tests.  If this had been merged, it would have been extremely difficult to detect the bug in the catch_calc_soil_moist() interface (item 3 above and  commit [3e77b31)](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/208/commits/3e77b31f643a9e01b00a9c375fadb8e01a55336e) at some later time, possibly weeks or months removed from this merger.

smahanam:
I disagree @gmao-rreichle it is not yet the time to add a CATCHCN test case. After this PR, I need a few things to push, we can start from that version.

gmao-rreichle:
@smahanam : I had asked @weiyuan-jiang to "run the suite of GEOSldas tests", which should have included a CatchCN run.  It's still not clear to me how we could have passed zero-diff for CatchCN in these tests.  Perhaps only the SMAP test case was run.

gmao-rreichle:
@smahanam : You are reorganizing CatchCN in a major way from the version that has been used in publications.  What is the plan for science verification of these changes if we don't test CatchCN? If we start the baseline after merging all the changes, how do we know it's any good?

smahanam:
If GEOSldas could put together a document like this:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/wiki/Zero-diff-changes

I would have done GEOSldas zero-diff.  I did not know GEOSldas has implemented zero-diff rule for CATCHCN.  It is bit too early. I and @mathomp4 created a branch that has 2 versions of CATCHCN (CLM4 and CLM45). That would be a good starting point for test cases.

smahanam:
I checked @weiyuan-jiang , please run your GEOSldas zero-diff on CATCHCN with default settings. It looks like it will pass. Let me know.

gmao-rreichle:
I just confirmed that with @weiyuan-jiang that he had accidentally run the wrong test (using a different branch).  This explains why we didn't see a failure with the bug in the catch_calc_soil_moist() interface.  We need to test again.  

@smahanam : now that the interface bug is fixed, we still have the "data" statements to worry about.  Are you saying that the default settings will not engage these data statements, and that we should expect a zero-diff result when we run the test now?

smahanam:
"we still have the "data" statements to worry about. Are you saying that the default settings will not engage these data statements, and that we should expect a zero-diff result when we run the test now?" CORRECT

gmao-rreichle:
@sdrabenh @smahanam : The repeat of the test exposed major issues with the setup of the nightly tests.  At the very least, the reconfigured rc setup doesn't work out of the box with the GEOSldas nightly tests.  We will need to investigate.  The pull request should NOT be merged until further notice.

smahanam:
@weiyuan-jiang can we talk? @GEESgcm_GridComp science science modifications in this branch are zero-diff. I'll swing by your office.

gmao-rreichle:
@smahanam : Re. your comment

> If GEOSldas could put together a document like this:
> https://github.com/GEOS-ESM/GEOSgcm_GridComp/wiki/Zero-diff-changes

You are absolutely correct.  This was on Justin's list but didn't get done before he left.  Weiyuan will work with you to learn how to run the GEOSldas tests.  In the process, perhaps the two you can put together the corresponding wiki page for GEOSldas.

sdrabenh:
@gmao-rreichle regarding the "z0_formulation" default value in GEOS_CatchGridComp.F90 and GEOS_CatchCNGridComp.F90, those lines have not been changed since these files were initially added to git 2019-06-07. This is not something @GEOS-ESM/gcm-gatekeepers would have changed without input/PRs from the land group. Also, if any PRs are a work in progress, please mark them as draft with a DNA label and change them to official pull requests only when the land group has reached a consensus that it is finalized and ready.

smahanam:
@sdrabenh can't agree more !  I am closing this PR, that has become a practice for my PR.
@weiyuan-jiang : top priority, we need to sit together and come up with a plan and add to land protocol, alternatively, I can pass Randy's and my updates to you, you can pass to the next level by yourself.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOS_CatchCNGridComp.F90
```diff
@@ -7184,7 +7162,7 @@ max_dayl = (86400./MAPL_PI) * acos(fac)  ! maximum daylength (sec; function of l
           if(associated(CNNPP )) cnnpp  = 1.e-3*npp   * cnsum
           if(associated(CNGPP )) cngpp  = 1.e-3*gpp   * cnsum
           if(associated(CNSR  )) cnsr   = 1.e-3*sr    * cnsum
-          if(associated(CNNEE )) cnnee  = 1.e-3*nee   * cnsum
+          if(associated(CNNEE )) cnnee  = 1.e-3*nee   ! * cnsum
```
gmao-rreichle: Is this intentional? Or is it accidentally left over from some debug test?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOS_CatchCNGridComp.F90
```diff
@@ -7217,7 +7195,7 @@ max_dayl = (86400./MAPL_PI) * acos(fac)  ! maximum daylength (sec; function of l
           if(associated(CNNPP )) cnnpp  = 0.
           if(associated(CNGPP )) cngpp  = 0.
           if(associated(CNSR  )) cnsr   = 0.
-          if(associated(CNNEE )) cnnee  = 0.
+          if(associated(CNNEE )) cnnee  = 1.e-3*nee
```
gmao-rreichle: Same question again: Is this intentional? Or is it accidentally left over from some debug test?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOS_CatchCNGridComp.F90
smahanam: Yes the change is intentional. As part of this modification NEE is saved too. Atmospheric CO2 model wants to import CNNEE from CATCHCN at every time step for 2-way CO2 coupling. Before this modification, since the CN model is only called at DTCN, CNNEE was updated only at DTCN but to account for the rest of the time steps outside DTCN, NEE was multiplied by pnsum.  Say for e.g., computed NEEs are 1. and 2, AGCM DT = 15min, DTCN is 90min

Exported CNNEE before this modification
6,0,0,0,0,0,12,0,0,0,0,0
After this modification
1,1,1,1,1,1,2,2,2,2,2,2 (so atmospheric carbon model reads a  useful value from CATCHCN at every time step.
This fix was documented in CVS with this comment, I guess, this got documented here too.

gmao-rreichle: Sarith: Thanks for the explanation.  Two things, though: I don't think the github.com comments are necessarily sufficient documentation.  My suggestion would be to make a brief comment in the F90 file, which will be there regardless of the version control software.  Also, why not make the same change to the other variables (cnnpp, cngpp, ....)???  Shouldn't the same reasoning apply to the other variables?

smahanam: OK - I add a comment. 
Other variables: we sure can do that, that need never arose.

smahanam:
@weiyuan-jiang said the modifications passed GEOSldas nightly test for CATCH while one CATCHCN diagnostic field is different (CATCHCN model state passed
[#234](https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/234) : I actually had ignore IGNORE_VEG_HEIGHTS when I implemented GEOS_LandGrid.rc, as per @gmao-rreichle , I included later to this branch.

With some notes about NEE and I can remove IGNORE_VEG_HEIGHTS too, are we good to proceed? If we all are on the same page, I can open a PR. After addressing above. I have a backlog in a private branch, that awaits to see develop.

gmao-rreichle:
@smahanam : 
I just met with @weiyuan-jiang , and we think that feature/SM_LandGCRC_145 is now ok.
The solution is to keep a separate resource file for GEOSldas that contains many of the same parameters that are already in GEOS_LandGridComp.rc.  The two-file solution is the only way to maintain different defaults for the AGCM and GEOSldas, which we unfortunately have to have because the AGCM still uses the old land model as the default.

HOWEVER, because of the non-zero-diff CN test, we will wait for Scott to merge #224 and #229 before re-opening this pull request (# 208).
I will talk with Scott to find out what the schedule might be for merging #224 and #229.  

Re. IGNORE_VEG_HEIGHTS:  I'm a bit nervous about making this change before we have a better sense of whether any of the "new land" restarts might still contain the JPL veg heights.  This needs to be sorted out with Scott.  It's probably better to keep this as a separate issue and eventual pull request.  There's no urgency.

PR #213, opened 2020-02-14
Fixes #212. Update CircleCI for GridComp

### LINKED ISSUE #212: CircleCI is broken with MAPL 2.0
As with the fixture, we need to fix the CircleCI in GEOSgcm_GridComp

mathomp4:


mathomp4:
Okay. This seems to work now. One weird CMake hack I'm not proud of, but working is better than not.

PR #219, opened 2020-02-20
mv getids.H to mk_shared.F90 under land shared

weiyuan-jiang:
mv mk_LDASsaRestarts.F90 to GEOSldas App


smahanam:
GEOSland_GridComp/Shared is to keep shared modules for land surface models (CATCH and CATCHCN)
mk_restarts directory contains all regridding tools for land, ice, lake, saltwater and all kind of surface tiles. Those routines in getids.H are part of regridding. So I don't think GEOSland_GridComp/Shared is for the place for new module mk_shared.F90

Perhaps, we can move back mk_shared.F90 back to mk_restarts and  just rename to getids.F90.

My 2cs, other than that your changes to getids.H look good to me.

weiyuan-jiang:
That means we need to generate a module or a lib out of the mk_restarts. That is something I want to avoid

smahanam:
I understood your point, but ....sorry,  I still can't agree.
I'll leave it to you, anyways.

gmao-rreichle:
I don't quite understand why we shouldn't have a module or lib in ./mk_restarts.  Don't we have executables there?  If there's a good reason to avoid putting the new module into mk_restarts, shouldn't it live further  up in Surface/Shared?  Land/Shared is somewhere inbetween and not a good fit.

smahanam:
If @weiyuan-jiang really wants to avoid adding a new library, Surface/Shared more sense is more sense to me.

weiyuan-jiang:
Then maybe it is good time to re-org the codes. I have no idea why mk_restars is under catch_gridcomp  where there are ice, water , catchcn, lake,....

smahanam:
If you really want to think, the whole mk_restarts directory should move out of GEOScatch_GrdiComp and move to Surface/Shared
like wehave Surface/Shared/Raster for surface  BCs

gmao-rreichle:
I have no idea why mk_restarts lives in Catch, but presumably there are Catch-specific subroutines in the package.   Moving everything to Surface isn't a great idea, necessarily.  Presumably, each Component (Catch, Veg, etc) should have some component-specific code to generate restarts, and everything should be tied together by some programs in Surface/Shared that deal primarily with location (geographic) info.   
If you are serious about reorganizing, I recommend talking with Larry first.  He may have additional helpful insights.

smahanam:
Yes talk to @atrayano, @lltakacs , @gmao-jstassi  first
As I understand, moving mk_restarts to Surface/Shared is the right thing.

weiyuan-jiang:
The file name is renamed to mk_restarts_shared.F90 that says itself

gmao-rreichle:
Closing this pull request for now until we have got clearance from Scott et al to proceed.

smahanam:
@weiyuan

PR #224, opened 2020-03-04
zero-diff changes in re-gridding (re-tiling) restarts

gmao-rreichle:
GEOSldas needs to have access to "getids" subroutines, which have been put into a module.
Also contains some clean-up and renaming.
Addresses issue #225


sdrabenh:
Merging since regridding worked and was 0-diff

PR #226, opened 2020-03-05
pmn: now gets daylength from MAPL sun code abstractly so works also w…

dr0cloud:
CatchCN now gets daylength from MAPL sun code abstractly so works also with new precession including type of orbit

*Note* from @mathomp4: Requires a new version of MAPL (2.1) not yet released. Will fail build tests until brought in because of new interfaces

*Note 2* Not turned on by default even when it can build as it is protected by an `if(.true.)` for the current default code.


dr0cloud:
Adding @smahanam as a reviewer as he as probably the CatchCN person?

smahanam:
Thank you @dr0cloud ! 
So this only affects  MAPL_SunOrbitQuery - day length calculation. Since we can't build this without MAPL 2.1, I can't test your modification. Now that we are aware of the modification, I'll have it turned in on (.false.) once MAPL2.1 is available.

dr0cloud:
@smahanam Dont turn on till bill approves. Thanks Sarith. Peter.

gmao-rreichle:
@dr0cloud @sdrabenh :  Thanks for alerting Sarith, who let me know about this pull request.  I assume this will be non-zero diff for the CatchCN tests running with GEOSldas.  Would it be possible to phase the mergers such that pull request #224 goes in first and the present pull request goes in a day later?  This would verify #224 being zero-diff across all nightly tests (GCM & GEOSldas) before having a non-zero diff change.

smahanam:
@dr0cloud 's if(.true.) logic will ensure that the modification is zero-diff.

gmao-rreichle:
@smahanam @dr0cloud @sdrabenh : Sorry, I was in a hurry and didn't look closely enough.  I guess the if (.true.) logic is in preparation for a later non-zero diff change.  
Seeing Matt's note about the MAPL-2.1 requirement, it looks like this isn't quite ready to move.  
In any case, with the if (.true.) logic it's ok for GEOSldas.

dr0cloud:
Hi Rolf, I’ll look at this straight after monitoring meeting. Peter.

gmao-rreichle:
I spoke with @dr0cloud.  Here's what we agreed on:

The modification still needs verification with CatchmentCN, so @gmao-jkolassa should go ahead, run the new code (by changing "if (.true.)" to "if (.false.)" for a sufficiently long period to establish that the day length ("dayl") is ok.  It's possibly zero-diff, but more likely round-off diff.

Additional considerations:

- @dr0cloud remembers that some variables related to orbital calculations ("zs" and "zc"?) still need attention (and may be in the CatchCN restarts under some name).  This requires further investigation by @gmao-jkolassa .

- @dr0cloud has added a new system for orbital calculations into MAPL.  The code change under consideration in this pull request is designed to work with the old and new orbital systems.  After verifying the pull request with the old orbital system, @gmao-jkolassa should try turning on the new orbital system and see if that also works out.  This can be done by adding "ORBIT_ANAL2B: 2" (sp?) to the AGCM.rc file.  (The assumption is that the new orbital system is available in a recent tag of MAPL and can be used with GEOSldas.)

- @dr0cloud also added a new 'equation of time', which more accurately determines when the sun reaches its highest point.  This can be turned on with something like "EOT: TRUE" (sp?) in AGCM.rc.  It's probably best to turn on both the new orbit system and EOT together.

- @dr0cloud is available to help with the above configuration choices and to fix any problems in the orbital calcs that our verification may turn up.

- @dr0cloud will talk about some of the orbit calcs changes at the GMAO Theme Meeting next Thursday.

dr0cloud:
Talked to Rolf by phone. The next step is to test "as is" that it is bit zero-if. *Then* to change the if(.TRUE.) to an if(.FALSE.) in the day length calculation of the catchCN GridComp and recompile. The code should then be either bit zero-diff or near zero-diff with the old orbital system (the default), but with the new orbital system (adding EOT: .TRUE. and ORBIT_ANAL2B: .TRUE. to to AGCM.rc), it will be non-zero-diff. It should not make a big difference but the new orbital system with the equation of time is more accurate and should become the default in the future. I will respond to an issues that come up during testing. If all the testing goes smoothly, the if(.TRUE.) branch (the old code) should eventually be removed completely, because the else (.FALSE.) branch handles both the old and new orbital systems agnostically (i.e., in a much better coding practice (encapsulation) of not interacting with the innards of the orbital system from the outside). Peter.

dr0cloud:
Just saw Rolf's post. My previous post is very similar, except for two points: (a) turning on the new orbital system uses ORBIT_ANAL2B: .TRUE. (not 2), and (b) I suggested that the very first thing we do is make sure this PR is zero-diff out-of-the-box, but this is probably checked already in the PR approval process. Peter.

gmao-rreichle:
Thanks, @dr0cloud , for the clarification.  Agreed.

mathomp4:
> Just saw Rolf's post. My previous post is very similar, except for two points: (a) turning on the new orbital system uses ORBIT_ANAL2B: .TRUE. (not 2), and (b) I suggested that the very first thing we do is make sure this PR is zero-diff out-of-the-box, but this is probably checked already in the PR approval process. Peter.

All, I'm a bit late on this, but as a reminder these are *logical* options in `AGCM.rc` so you need those dots:
```
ORBIT_ANAL2B: .TRUE.
EOT: .TRUE.
```
I thank @dr0cloud for using the logicals as using 0/1 integers as a logical switch can lead to fun with GCC (as `if(integer)` is not allowed by Fortran strictly)

gmao-rreichle:
In an attempt to consolidate the offline email discussion that seems to touch only subgroups of those involved, here's my latest understanding:
1) Using the old orbit system, Jana's testing showed that the new daylength calc is within roundoff of the old daylength calc. My recommendation is to eventually merge the _new_ daylength calc as a non-zero-diff change for CatchCN.  (The merge is trivially 0-diff for Catchment.)
2) Matt merged the latest "develop" branch into "feature/pnorris/fixes-#205"
3) Jana's testing revealed that CatchCN does not run with the new orbit system.  Peter is fixing the issue.  My recommendation is to add this fix to the branch and just have one merge that switches to the new daylength calc _and_ has the capability to run the new orbit system (but keep the old orbit as default).

cc: @rdkoster , @gmao-jkolassa , @weiyuan-jiang , @mathomp4 , @dr0cloud

dr0cloud:
Rolf, Sorry for leaving you out of the discussion. Here’s what I just sent to Jana with a few mods:

Hi Jana,
I have pushed a fix to feature/pnorris/fixes-#205. It only changes the catchCN Grid Comp. Can you please get Matt’s help to update your test code to these changes.

After that, these are the next steps:

  1.  Run the test with the new orbital system (ORBIT_ANAL2B: and EOT: both .TRUE.) To do this, no changes to the code are needed, because the IF(.TRUE.) has now been changed to an IF(.FALSE.) meaning that the old catchCN code is replaced by the new catchCN code. In addition, my fix has now added an extra IF(.FALSE.) that stops the previously offending code from being run, since it is not needed under the new catchCN code. You should hopefully have a successful run with non-zero-diff but minimal changes.
  2.  If that works, rerun the test with the default heritage orbital system (i.e., without the EOT and ORBIT_ANAL2B in AGCM.rc). Again, no code changes needed. This should also be non-zero-diff but minimal changes.
  3.  If both these work, I want to do some final code cleanup, namely to completely remove both ugly old code (IF(.FALSE.)) branches, and remove the declarations with ‘Xpmn’ comments. The whole point of this issue was to remove this old code and make catchCN agnostic to internal orbital matters.
  4.  Once that change is pushed, we can issue a pull request (non-zero-diff but minimal changes).

Sorry to all for my original incompetence. Hopefully all is good from here on (??). If not I’ll fix it quickly.

gmao-rreichle:
_Adding copy of email from Jana with test report for future reference:_

From: Kolassa, Jana (GSFC-610.1)
Wed 20-May-20 4:01 PM

[...] I re-ran the GEOSldas Catchment-CN experiments using the updated branch feature/pnorris/fixes-#205. I ran two configurations:

(1) with new daylength calculation ('if (.false.)') and new orbit system switched off (EOT and ORBIT_ANAL2B set to .false.)
(2) with new daylength calculation ('if (.false.)') and new orbit system switched on (EOT and ORBIT_ANAL2B set to .true.)

Both configurations now run fine with Catchment-CN. Both configurations are not zero-diff, but within roundoff of the baseline. 

A previous experiment with the old daylength calculation ('if (.true.)') and old orbit system, that I ran as a sanity check, was zero-diff.   [...]

gmao-rreichle:
Next steps:
1) @dr0cloud to clean up the branch.  This cleanup should be trivially zero-diff from the tested version of the branch.
2) Merge after formal approval by @wmputman (zero-diff for Catch, non-zero-diff for CatchCN).

cc: @rdkoster @weiyuan-jiang @gmao-jkolassa @mathomp4

dr0cloud:
I have just pushed a cleaned up version.

dr0cloud:
Trivially zero-diff to test just performed.
Non-zero-diff the CatchCN
Zero-diff to Catch.

Issue #228, opened 2020-03-05
fix reading of Catch analysis increments into Catch Grid Comp


gmao-rreichle:
There is an error in the MAPL read statements used to read Catch analysis increments into the Catch GridComp.
See pull request #198

gmao-rreichle:
Addressed and merged with #229

PR #229, opened 2020-03-05
bug fix: reading of Catch analysis increments into Catch GridComp

gmao-rreichle:
addresses issue #228


gmao-rreichle:
Re-opening now that everything's moved to MAPL-2.  Ready for review.

PR #230, opened 2020-03-06
GF move ShallowCu contribution to mass flux and scale-dependent convection stochastics

sdrabenh:


mathomp4:
All,

This line:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/120e5e8187d03fc7064ae2ad7e550dec0b17d33e/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/GEOS_MoistGridComp.F90#L7738
is causing a failure to build with GNU because `if(integer)` is not allowed.

This either needs to be:
```fortran
if (STOCHASTIC_CNV /= 0) then
```
or `STOCHASTIC_CNV` needs to be a `LOGICAL` and the `MAPL_GetResource()` calls changed appropriately.

I'll make a PR into Scott's branch for the former as it's "simpler".

mathomp4:
PR #231 has been made to be pulled into `feature/sdrabenh/GFDL_stoch_CNV_MF`

sdrabenh:
Thanks @mathomp4 ! I meant to change that over the weekend but ran out of time. It built fine with Intel but I noticed that failure point with GNU.

PR #231, opened 2020-03-09
Fix for GNU for if(integer) in Moist GC

mathomp4:
GNU does not allow `if(integer)`


mathomp4:
Marked 0-diff as I believe it's zero-diff to what's in the feature branch. I don't know if that's zero-diff to GEOSgcm v10.10

PR #237, opened 2020-03-13
Fix logging error with ESMF in GCM

mathomp4:
ESMF was logging an error due to calling `ESMF_AlarmIsRinging` on an alarm that's only created if running replay (I think?). So we protect to make sure the alarm is created.

### DIFF
--- a/GEOS_GcmGridComp.F90
+++ b/GEOS_GcmGridComp.F90
@@ -1702,7 +1702,9 @@ contains
 
     ! Ensure Active PREDICTOR_STEP Alarm of OFF
     ! -----------------------------------------
-    IF(ESMF_AlarmIsRinging(PredictorIsActive)) CALL ESMF_AlarmRingerOff(PredictorIsActive, RC=STATUS)
+    if(ESMF_AlarmIsCreated(PredictorIsActive)) then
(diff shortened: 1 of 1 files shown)


mathomp4:
@sdrabenh This should be zero-diff with replay on and off, and I think I tested both, but it's probably good for you to test as well to double check me!

mathomp4:
@sdrabenh Were you able to test this bugfix? It should be zero diff.

sdrabenh:
@mathomp4 I'll try to get this tested today.

sdrabenh:
@mathomp4 it looks like this may have fixed a number of issues. For the first time during my 0-diff tests, all non-conventional history collections (modis, lidar, radar, etc) were also 0-diff for amip, replay, and 0-increment replays. That's great!

mathomp4:
> @mathomp4 it looks like this may have fixed a number of issues. For the first time during my 0-diff tests, all non-conventional history collections (modis, lidar, radar, etc) were also 0-diff for amip, replay, and 0-increment replays. That's great!

I'm definitely surprised it affected anything like that. I wonder why? Maybe some weird ESMF interaction??

PR #248, opened 2020-04-02
Bugfix/san akel/#116 conflicts mom5 mom6

sanAkel:
Responds to @tclune https://github.com/GEOS-ESM/GEOSgcm/issues/116#issuecomment-608038679

https://github.com/GEOS-ESM/GEOSgcm/issues/116


GEOSogcm_GridComp/GEOSocean_GridComp/GuestOcean_GridComp/GuestOcean_GridComp.F90
```diff
@@ -544,11 +544,11 @@ subroutine SetServices ( GC, RC )
             CHILD_ID   = OCN,                                         &
             RC=STATUS  )
        VERIFY_(STATUS)
-       call MAPL_AddExportSpec ( GC   ,                          &
-            SHORT_NAME = 'SSH',                                       &
-            CHILD_ID   = OCN,                                         &
-            RC=STATUS  )
-       VERIFY_(STATUS)
+!!$       call MAPL_AddExportSpec ( GC   ,                          &
```
tclune, line 547: Not really my purview, but I strongly recommend against commented out code.   Delete it if it is not needed any more.  (The repository will remember for you.)   If it needs a conditional, add a conditional.

yvikhlya: @tclune This code is actually needed, but MOM6 plug does not provide these exports, and we are trying to figure out how to get them from MOM6. I commented this just to make MOM6 plug going to test it as temporary solution (hack). This should not go into develop branch of course.

tclune: OK - I have added the "Contingent - DNA" tag for the time being then.    That tag will prevent accidental merge onto develop.  The tag should of course be manually removed when you are ready.

sanAkel: That's right! In the sense we would `want' these exports from MOM6 as well, but do not know (yet!) how to get them. Having said that (and the reason why I'm leaving a comment is because...) both GFDL - MOM6 developers and myself do not want to export full 3D fields. 

Reason being MOM6 hybrid coordinates that change from one time step to another renders any 3D field to be `meaningless' unless they are associated with a `fixed' vertical grid, which does not exist in the MOM6 world. 

In any case, we'll explore this issue with GFDL...

Externals.cfg
```diff
@@ -20,5 +20,12 @@ local_path = ./GEOSogcm_GridComp/GEOSocean_GridComp/GuestOcean_GridComp/MOM_GEOS
 tag = geos/v1.0.1
 protocol = git
 
+[mom6]
```
mathomp4, line 23: @sanAkel  A couple notes.

First, if this goes into production I *really* do not want a branch in this. It can make it impossible to duplicate results if a push goes into mom6 `geos` and no one knows about it. (I am still annoyed as all get out we have to point to a branch in GSW, but I'll allow that as we don't control it.)

Second, we'll want/need an equivalent section for the `components.yaml` in the GEOSgcm fixture. You can use the branch I provided to you as a template. Since mom6 has submodules, the mepo section is a bit unique.

sanAkel: @mathomp4 I get your 1st point and concern reg the `branch.' GFDL folks are willing to have our code/modifications on their branch, of course, we are the ones actively maintaining/testing it (that's what their doing with other groups such as NCAR, HYCOM). Rather than hashing it here on github, we can sort it elsewhere?? *I'll remember to offer a solution for GSW* please remind me if I forget!

Reg. your second point: components.yml, if we have a clean fix for above 1, would that not be enough? 

Because current components.yml reads:

> mom6:
  local: ./src/Components/@GEOSgcm_GridComp/GEOSogcm_GridComp/GEOSocean_GridComp/GuestOcean_GridComp/MOM6_GEOSPlug/@mom6
  remote: ../MOM6.git
  **branch: geos
  develop: geos**
  recurse_submodules: True

>

mathomp4: @sanAkel 

I had to do that in `components.yaml` to get things to work. Since the last tag on our mom6 fork (`geos/v0.9.1`) there are 4 commits to the `geos` branch:

https://github.com/GEOS-ESM/MOM6/compare/geos/v0.9.1...GEOS-ESM:geos

Since mom6 is not my repo, I didn't want to assume/presume to make a new tag on `geos` (`geos/v0.9.2`?) which has these commits since I didn't know if you or @aerorahul  had any other work needed before a new tag. 

Thus, I used the only code path that was workable which was to use the branch in full (since GSW moved out of mom6, you can't compile `geos/v0.9.1` with the GSW currently in GEOS being around.

If you are "satisifed" with the mom6 we currently have, you could make a new release/pre-release off of `geos` and then use that tag.

sanAkel:
https://github.com/GEOS-ESM/GEOSgcm/pull/172 --> close this issue.

PR #250, opened 2020-04-03
Bugfix/yury/tskinifix

### LINKED ISSUE #249: Coupled model crashes in GEOS_SurfGridComp
Due to a bug with TSKINI.

yvikhlya:
Fixes #249.


sdrabenh:
@yvikhlya I have no problem with this PR. However, please make sure that changing this name does not adversely impact the das @rtodling , land @gmao-rreichle , or aoil @sanAkel

yvikhlya:
This fix was made by @sanAkel, so I guess he knew what he did. It is zero diff in AMIP mode and should affect only the coupled model.

sanAkel:
That's right. This bug only showed up in coupled model configuration. So the fix only applies to *that* configuration, i.e., coupled. 

As far as I know, @rtodling , @gmao-rreichle do not use coupled model configuration; leave them off this, shall we?

sdrabenh:
> That's right. This bug only showed up in coupled model configuration. So the fix only applies to _that_ configuration, i.e., coupled.
> 
> As far as I know, @rtodling , @gmao-rreichle do not use coupled model configuration; leave them off this, shall we?

@sanAkel that sounds reasonable to me as long as this is only impacts things in coupled mode.

gmao-rreichle:
@sanAkel @yvikhlya @sdrabenh : Thanks for keeping land in the loop.  If anything changes in Surface GridComp or below, the offline land system (GEOSldas) is potentially impacted.   
This particular change is fine.

Issue #150, opened 2019-12-05
Catchment and CatchmentCN model : PRECIPFRAC


smahanam:
Introducing PRECIPFRAC parameter to deal with drizzle.

PR #256, opened 2020-04-08
PRECIPFRAC functionality for Catch and CatchCN 

gmao-rreichle:
Add functionality to unify and generalize tile area fraction on which precipitation falls for the purpose of interception loss and surface runoff.

Addresses issue #150 

*Requires* 0-diff testing for GCM *and* *GEOSldas*.  See [GEOSldas Wiki](https://github.com/GEOS-ESM/GEOSldas/wiki) for info on GEOSldas tests.


GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/catchmentCN.F90
```diff
@@ -1128,25 +1128,41 @@ SUBROUTINE CATCHCN (                                           &
...
 
 !**** UPDATE CANOPY INTERCEPTION; DETERMINE THROUGHFALL RATES.
-
-      CALL INTERC (                                                            &
-                   NCH, DTSTEP, TRAINLX, TRAINCX, SMELT,                       &
-                   SATCAP, SFRAC,BUG,                                          &
-                   CAPAC,                                                      &
-                   THRU                                                        &
-                  )
+      IF(SFRAC == 1.) THEN
```
gmao-rreichle, line 1131: I think this particular implementation using two separate subroutines is confusing in the long run.
It looks like PRECIPFRAC=.9999 would end up running with different physics because the FWETL and FWETC parameters in lsm_routines.F90 are chosen based on the floating point value of PRECIPFRAC.
That is, the results for PRECIPFRAC --> 1. would NOT converge to PRECIPFRAC=1.
Perhaps a better way to deal with this is to introduce a new integer or logical switch that selects between the "_1p" and "_2p" physics.  In all cases, floating point values for PRECIPFRAC, FWETL and FWETC need to be provided.  To get zero-diff, set the switch to "_1p" and provide the old values for PRECIPFRAC, FWETL and FWETC.

gmao-rreichle, line 1131: Why not use this opportunity and rename SFRAC to PRECIPFRAC throughout?  There's no need for two different names for what's essentially the same variable.

gmao-rreichle, line 1131: Add "ONLY:" to "use lsm_routines" and explicitly list the subroutines from the lsm_routines that are used here.  This will make it clearer where to find, e.g., "srunoff()"

gmao-rreichle, line 1131: Some of the above comments also apply to "catchment.F90".

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/Shared/lsm_routines.F90
```diff
@@ -258,11 +274,281 @@ SUBROUTINE INTERC (                                                      &
  100  CONTINUE
 !****
       RETURN
-      END SUBROUTINE INTERC
+      END SUBROUTINE INTERC_single_throughfall
+
+! -----------------------------------------------------------------------------
+
+      SUBROUTINE INTERC_ls_and_conv_throughfalls (                                                      &
```
gmao-rreichle, line 281: If I'm reading this correctly the implementation could be simplified considerably.
My understanding is that a single subroutine would suffice if FWETL and FWETC are provided as inputs to the subroutines and THRU, THRUL and THRUC are all returned as output (regardless of 1p or 2p).
In the 1p case, the THRU output is used.  In the 2p case, the THRUL and THRUC outputs are used.
There's at least a very good chance of getting 0-diff while avoiding having two nearly-identical copies of the INTERC subroutine.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/Shared/lsm_routines.F90
```diff
@@ -258,11 +274,281 @@ SUBROUTINE INTERC (                                                      &
...
+
+!**** - - - - - - - - - - - - - - - - - - - - - - - - - 
+
+      DO N=1,NCH
+
+         !**** Compute runoff from large-scale and convective storms separately:
+         
+         deficit=srfmx(n)-srfexc(n)
+         
+         srunl=AR1(n)*THRUL(n)
```
gmao-rreichle, line 508: This question applies to the old and new code: What happens when arsnow>0?  How does this work when some or all of the precip falls on snow?  Is this taken into account elsewhere?  If not, should it be addressed as part of these changes?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/catchmentCN.F90
gmao-rreichle, line 1131: Here's another question: Is it correct that even in the new code, only the products FWETL*PRECIPFRAC and FWETC*PRECIPFRAC are used?  If so, there's no need to have three input parameters.  Two would be sufficient.  We could call them PRECIPFRAC_L and PRECIPFRAC_C.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/Shared/lsm_routines.F90
```diff
@@ -258,11 +274,281 @@ SUBROUTINE INTERC (                                                      &
...
+
 !****
 !**** [ END INTERC ]
 !****
 !****
+!**** ===================================================
+!**** ///////////////////////////////////////////////////
+!**** ===================================================
+
+      SUBROUTINE SRUNOFF_single_throughfall (                                                  &
```
gmao-rreichle, line 410: It's not as easy to see if it would be possible to have only one subroutine srunoff() that can handle both the 1p and 2p cases.  If it's possible, it would probably be better.

PR #260, opened 2020-04-14
Fixes #258. Allow GNU to run

### LINKED ISSUE #258: Current develop branch doesn't run with GNU
The nightly tests last night died running with GNU. It seems to be due to the addition of #174 into `develop`.

Running with Debug build we see:
```
At line 589 of file /discover/nobackup/mathomp4/SystemTests/builds/AGCM_GNU/CURRENT/GEOSgcm/src/Components/@GEOSgcm_GridComp/GEOS_GcmGridComp.F90
Fortran runtime error: Different CHARACTER lengths (7/8) in array constructor
```
where we have:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/a48b33e1735c8ee2356a46b847236a52d6c0c610/GEOS_GcmGridComp.F90#L586-L595

I'll need to write a tester to see what's happening here between the two compilers.

mathomp4:
This code fix does seem to let GNU run. Testing with Intel in AGCM mode shows zero-diff but I might have royally screwed up the Coupled model. I'll ping @yvikhlya to test it out.

### DIFF
--- a/GEOS_GcmGridComp.F90
+++ b/GEOS_GcmGridComp.F90
@@ -585,13 +585,14 @@ contains
 
      do k=1, 33
       write(unit = suffix, fmt = '(i2.2)') k
-      call MAPL_TerminateImport( GC,                                &
-           SHORT_NAME = (/'TAUA_'//suffix,                          &
-                          'ASYMP_'//suffix,                         &
-                          'SSALB_'//suffix/),                       &
-           CHILD      = OGCM,                                       &
-           RC=STATUS  )
-       VERIFY_(STATUS)
+      call MAPL_TerminateImport( GC,           &
+         SHORT_NAME = [ character(len=(8)) ::  &
+            'TAUA_'//suffix,                   &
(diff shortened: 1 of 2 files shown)


mathomp4:
Note: It looks like there was a change in `develop` since I started this (#259). I updated the branch here. I don't think it should affect this PR that I didn't test with that. But @sdrabenh and @yvikhlya should of course check to make sure it passes their tests of working/zero-diff.

mathomp4:
NOTE: I've added the "Contingent - DNA" label only so that we make sure @yvikhlya says this is safe before we pull it in. If it says it tests okay, we can remove that label.

yvikhlya:
I would ping @atrayano  on this too. Atanas, let's both review these `if(do_obio/=0)`, just to see if they are in right places or need to be optimized and then I'll test it in coupled/dual ocean.

sdrabenh:
@yvikhlya is there anything left to do with this PR? You've already tested in couple mode, correct? Can we approve this now?

PR #261, opened 2020-04-14
Sync dev into master

sdrabenh:


yvikhlya:
@sdrabenh Let me test it in coupled mode.

yvikhlya:
@sdrabenh @mathomp4 Also, should this merge include [260](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/260), or should it be added later?

sdrabenh:
> @sdrabenh @mathomp4 Also, should this merge include [260](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/260), or should it be added later?

Ideally, yes, #260 should be included as well depending on how long the coupled tests will take to complete. I really need to sync develop into master and issue a tag tomorrow because other users have requested a tag reference on master (with the latest changes now in develop) so their systems can run. Unfortunately, this PR rolls in 10 other PRs - all of which have been approved already in develop.

yvikhlya:
@sdrabenh @mathomp4 Good news is that develop works in coupled mode (which is unusual). Let me check if we can merge Matt's branch into develop before merging everything  to master. I have a meeting now, but it should not take a lot of time after meeting. I don't need to run whole lot of tests, if it is zero diff with develop, it is good to go.

sdrabenh:
@yvikhlya you should be able to approve #260 and remove the DNA label. At that point I will merge it into develop. Then I'll update this PR, but then you have to approve this PR too.

yvikhlya:
@mathomp4 @sdrabenh Sorry it takes so long. System hiccups are horrible today. Still did not finish zero diff test with #260 , but looking at code I think everything is ok. Is there anything else I need to do to close this PR?

Issue #273, opened 2020-05-04
Update RRTMGP to Pincus Master as of 3 May 2020


dr0cloud:
Will be zero-diff because RRTMGP is not on by default ... supersedes #254

Issue #254, opened 2020-04-07
Add NRLSSI2 and LW Jacobian RRTMGP changes from Pincus DEVELOP branch as at 2020 Apr 3


dr0cloud:
Add NRLSSI2 and LW Jacobian RRTMGP changes from Rob Pincus rte-rrtmgp GitHub DEVELOP branch as at 2020 Apr 3. Will be zero diff only in the sense that RRTMGP is still default off.

PR #274, opened 2020-05-05
Feature/pnorris/#273 rrtmgp to master may2020

dr0cloud:
This brings RRTMGP code up to date with recent Pincus Master changes as of 3 May 2020.
It should be zero-diff because RRTMGP off by default. It solves issue #273 and supercedes issue #254 (an earlier set of changes).


dr0cloud:
Thanks guys! Peter.

PR #276, opened 2020-05-11
Convert ConvPar to submodules

mathomp4:
This commit pulls ConvPar into a submodule to try to prevent recompilation of the expensive Moist GC.


mathomp4:
Note. This is zero-diff as far as I can see. I suppose if @tclune is happy with naming, etc. and @sdrabenh and @srfreita approves/doesn't mind, there is no reason we couldn't put this in `develop`. It seemed to work with GCC and Intel.

mathomp4:
@srfreita A query. Are you working on edits in the background that use submodules? If so, we can close this PR and wait for a new GF update with submodules instead.

srfreita:
Not yet. I am waiting for the approve of the new GF version. Then I will incorporate your changes.
Thanks,

PR #278, opened 2020-05-13
Fixes #277. Fix array bounds in RRTMG LW

### LINKED ISSUE #277: Bug in RRTMG LW: Array bounds mismatch
A nice silent bug found by @WilliamJamieson in his NUOPC work. It turns out that this simple line:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/41ee6a92a3973395470eedbc7a788da20dcb9197/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSradiation_GridComp/GEOSirrad_GridComp/RRTMG/rrtmg_lw/gcm_model/src/gpu_rrtmg_lw_rtrnmc.F90#L736
was a troublemaker. Why you ask? Well, you'd think all would be good. Let's look at `pzd`:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/41ee6a92a3973395470eedbc7a788da20dcb9197/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSradiation_GridComp/GEOSirrad_GridComp/RRTMG/rrtmg_lw/gcm_model/src/gpu_rrtmg_lw_rtrnmc.F90#L753
Good bounds for an edge variable. What about `pz`:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/41ee6a92a3973395470eedbc7a788da20dcb9197/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSradiation_GridComp/GEOSirrad_GridComp/RRTMG/rrtmg_lw/gcm_model/src/rrtmg_lw_rad.F90#L589
Oh dear. A 74 level variable.
The issue is that we compile GEOS with a flag that will *reallocate* arrays if you do:
```fortran
B=A
```
if the right hand side has a different shape than the left. So, we were "silently" reallocating pzd!

I'm making a PR for a simple fix and I'll ping @dr0cloud as a second pair of eyes (though the RRTMG LW code is such a mess...)

mathomp4:
For some reason, a couple variables (and some commented ones) were declared as 0:LM+1 which we do not do in GEOS.

I ran a test of GEOS with this fix and it seemed zero-diff.


mathomp4:
I have added a blocker label to this for two reasons:
1. I'd like @dr0cloud to take a look and make sure it seems sensible to him
2. I'd like @WilliamJamieson to try this fix in his NUOPC work to make sure it solves his issue.

dr0cloud:
Matt, I am actively looking at this, almost finished. I'll send something after the STM 130-230.

dr0cloud:
Matt, it may be tomorrow before I respond ... just noticed all those totuflux type variables that look like they are commented out but aren't really. P.

dr0cloud:
OK, let me at least comment for the parts without the outputs (totuflux, etc.) ... 
1. I agree with the pz and tz changes [pt]z(...,0:G+1)->(...,0:G), where G=GPU_MAXLEVS.
2. I need to study the totuflux, ... outputs tomorrow.
3. The following variables are declared (..,G+1) and I believe they can all be made (..,G) with the notes below:
a. pavel, tavel ... change to (...,G) with no problems.
b. coldry and wbrodl ... change to (...,G) with no problems.
c. wkl ... doesn't seem to be used, right?
d. wx ... Probably can be changed to (...,G) but GPU gymnastics in gpu_rrtmg_lw_taumol.F90 is beyond me, so pls check.
e. cldfracq, c[il]wp: these can be changed to (...,G) SO LONG AS the corresponding allocations of cldfracd, c[il]wpd in gmcica_subcol_gen_lw.F90 are likewise changed from (...,nlay+1) to (...,nlay).

Matt, can you verify these additional changes while I look at totuflux, etc. It just irritates me to see all these G+1 in the code (whether 0: or 1:) if they are not needed. Any idea why G+1 was used rather than just ?. Its too strange if there was no reason! Don't pull anything till I look at totuflux, etc. 
Peter.

mathomp4:
@dr0cloud I have no idea why they were that way. Maybe some padding or something for the GPUs? Or just a plain ol mistake. 

I still don't want them that way, though. Array bounds should be what the array bounds are.  If you pad, you wouldn't pad by one anyway.

I'll try and look at the other ones as well. I was just going for a "minimal change" PR.

dr0cloud:
Matt, as for the output variables (totuflux ...) changes, I agree with your G+1->G changes.
Once you've reviewed my other suggestions 3a-e above, and esp check on wx, I'm good to go with these changes. We made the code less confusing :)

mathomp4:
@dr0cloud I'll take a look now/soon. Once I confirm everything, I'll ping @WilliamJamieson so he can try our "omnibus" RRTMGLW fix in his much more sensitive NUOPC test.

mathomp4:
@dr0cloud I'm not sure `wx` is used for...anything. Like `wkl`. And together those two are equivalent to 42 3d variables. Just allocating and deallocating every time through the code.

dr0cloud:
Matt @mathomp4 I was just worried whether the wx1-4 in gpu_rrtmg_lw_taumol.F90 are related to wx in a way that I don't understand ... there are these sreg and dreg calls in the code that I don't understand.

mathomp4:
> Matt @mathomp4 I was just worried whether the wx1-4 in gpu_rrtmg_lw_taumol.F90 are related to wx in a way that I don't understand ... there are these sreg and dreg calls in the code that I don't understand.

Yeah. I think the sreg and dreg are just allocate/deallocate. And it looks like they never care about wx! I think some of these arrays did once and then the dev found a better way to get the data, etc.

dr0cloud:
@mathomp4 Ok, sounds good Matt. If you can run the fixed code zero-diff and it works for William then I think we're done. I'm going to sign off on the review now.

mathomp4:
@dr0cloud I just pushed one more commit to grab code from @WilliamJamieson 's branch #279 . FIgure this way we can do both in one shot.

mathomp4:
I think at this point, we are probably safe. We have a bunch of `B = A` in GEOS that could cause these.

But my latest look through RRTMG LW makes me think we are safe with it at least. All the bounds I see look good.

PR #279, opened 2020-05-13
Fixed bug discovered by the NUOPC online-ctm due to out of bounds arrays

### LINKED ISSUE #277: Bug in RRTMG LW: Array bounds mismatch

WilliamJamieson:
This is similar the Fixes #277. "Fix array bounds in RRTMG LW #278" @mathomp4  pull request. Here we add an additional fix eliminate the possibility of a silent reallocation of an array of a different size during a variable assignment.


mathomp4:
@WilliamJamieson I'm putting a blocking label on this. @dr0cloud is looking at my other PR #278 and might have other changes possible.

sdrabenh:
These changes have been included in #278 so I am closing this PR

PR #293, opened 2020-06-05
Feature/yury/s2s3 merge

yvikhlya:
This PR has a number of zero diff updates from S2S3 tag in CVS. It depends on https://github.com/GEOS-ESM/MAPL/issues/405 and https://github.com/GEOS-ESM/MOM5/pull/14. @sanAkel please review GEOS_SurfGridComp.F90 and GEOS_OpenWaterGridComp.F90. Updates in GEOS_GcmGridComp.F90, GEOS_OgcmGridComp.F90 and GuestOcean_GridComp.F90 affect dual ocean mode. Updates to MOM_GEOS5PlugMod.F90 fix export of mixed layer depth from MOM5 and record restart in dual ocean mode.


yvikhlya:
@mathomp4 By the way, I just noticed that Externals.cfg has geos/v1.0.1 tag for MOM5, which does not have the required export, and this will not compile. Should I change it to geos5 branch?

mathomp4:
> @mathomp4 By the way, I just noticed that Externals.cfg has geos/v1.0.1 tag for MOM5, which does not have the required export, and this will not compile. Should I change it to geos5 branch?

No. The correct thing would be to make a new release of MOM5 (geos/v1.0.2) and then @sdrabenh can update the Externals/components.yaml to that. 

There are some changes in the geos5 branch that haven't been released yet:

https://github.com/GEOS-ESM/MOM5/compare/geos/v1.0.1...geos5

mathomp4:
I could make a new release (or you could) and then you could edit the Externals.cfg at the same time. (Note, the `components.yaml` in GEOSgcm would need to change at the same time)

atrayano:
Yury,

Matt (and Ben and I) are in a meeting at the moment. We are trying to move to use mepo (and the related config file “components.yaml”. The syntax there is self explanatory)

Just run at the top level (i.e. GEOSgcm)
mepo init
mepo clone (of course after you make a change to components.yaml)

Atanas

yvikhlya:
> I could make a new release (or you could) and then you could edit the Externals.cfg at the same time. (Note, the `components.yaml` in GEOSgcm would need to change at the same time)

Can I ask you to handle the release and dependency fix? You know the rules better than me.

atrayano:
Yury,

I’ve never done a release. We can ask Matt, could you cc him?

yvikhlya:
I actually meant @mathomp4, could you handle the release and dependency updates in config files?

mathomp4:
I have released a new version of MOM5 with all the bug fixes since the last. I've updated the `Externals.cfg` here. 

I also made https://github.com/GEOS-ESM/GEOSgcm/pull/153 which updates `components.yaml` to `geos/v1.0.2`

yvikhlya:
@mathomp4 Thanks Matt. It still needs the dependency on https://github.com/GEOS-ESM/MAPL/issues/405 to be satisfied before it can build.

yvikhlya:
@sanAkel 
1.  Should be addressed to @zhaobin74.
2.1 SRESTORE is an optional feature, which is off by default. It is different from MOM6 srestore which you mentioned. Its purpose is to prevent model crashes in long 1/4 degree runs due to low salinity. We don't know if it will be a problem in MOM6 or not and this makes it difficult to decide how to implement this feature properly.  May be the best idea would be to leave the code where it is, but disable it for now. I'll think it up and push the update. 

Question about GEOS_GcmGridComp.F90 should be addressed to @atrayano

zhaobin74:
@sanAkel, as to 1, I just followed suite with a 'd' suffix for FRACICEd. The SEAICEd was added by @atrayano. FRACICEd is self-explanatory once SAEICEd 's meaning is clear.

I'll address 2.2 with a comment

yvikhlya:
@sanAkel Added comment on freeze salinity and removed SRESTORE feature from Guest.

mathomp4:
@yvikhlya Once you can confirm things work with MAPL `master` and the edit @atrayano has suggested, I can release MAPL 2.1.5

yvikhlya:
@mathomp4 @atrayano `MODE=MAPL_Write2Disk` works and I pushed a commit.

atrayano:
@yvikhlya Thanks for testing it! The real question is if the version of MAPL on the master branch works too, so @mathomp4 can proceed with the release of v2.1.5

yvikhlya:
> @yvikhlya Thanks for testing it! The real question is if the version of MAPL on the master branch works too, so @mathomp4 can proceed with the release of v2.1.5

I've tested it with MAPL master.

mathomp4:
MAPL 2.1.5 has been released.

yvikhlya:
> MAPL 2.1.5 has been released.

Will you update components.yaml?

mathomp4:
> > MAPL 2.1.5 has been released.
> 
> Will you update components.yaml?

Done! I've updated the branch on GEOSgcm to have MAPL 2.1.5 in all the files.

mathomp4:
Note: CircleCI will fail because the CI isn't as smart as it should be. The reason is that it clones GEOSgcm first. But it doesn't know it needs a different branch of the *fixture* to get the CI to pass. 

If the GEOSgcm was updated to have MAPL 2.1.5 and mom 1.0.2, it would build and pass.

sdrabenh:
@mathomp4 correct, but for testing, the branch should be to be up-to-date with develop

mathomp4:
> @mathomp4 correct, but for testing, the branch should be to be up-to-date with develop

Yes. For testing, you'll want to first do:
```
git checkout feature/yury/s2s3-merge
```
after cloning the fixture. *Then* you can do a `mepo init && mepo clone`.

sdrabenh:
@yvikhlya my tests do not show this PR is 0-diff. In both amip and replay mode, openwater_internal_checkpoint has different values for the SSKINW field ...

`               Date     Time   Level Gridsize    Miss    Diff : S Z  Max_Absdiff Max_Reldiff : Parameter name
     7 : 2016-12-04 21:00:00       0    99389       0   99389 : F F       25.000     0.83333 : SSKINW     
  1 of 12 records differ
cdo    diffn: Processed 24 variables over 2 timesteps [0.26s 19MB].
`

yvikhlya:
@sdrabenh Just a single restart? Let me check this again.

yvikhlya:
@sdrabenh I amended changes to GEOS_OpenWaterGridComp.F90 which caused non-zero diff under condition when use_skin_layer=1. We will deal with SSKINW computation later in separate pull request.

PR #306, opened 2020-06-24
Updates for JEDI Integration

mathomp4:


mathomp4:
Note, however, this update also requires an update to the GEOSgcm fixture as well (https://github.com/GEOS-ESM/GEOSgcm/pull/167) to even build. It needs a newer version of FMS with CMake changes that allow the changes to MOM5 to build.

mathomp4:
After some thought, I think the answer is https://github.com/GEOS-ESM/GEOSgcm/pull/167 needs to be accepted first and *then* this should be able to build. I did move mom5 to 1.0.3 in https://github.com/GEOS-ESM/GEOSgcm/pull/167, but that could be pulled out and just be a move to FMS 2019.01.02 first.

PR #312, opened 2020-07-07
Feature/mathomp4/add ncar gwd

sdrabenh:


mathomp4:
A question for @JulioTBacmeister  and @wmputman. My code changes were a bit "not perfect" in that I have this:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/12009ae51e5646875c54c91876858254a7590988/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSgwd_GridComp/GEOS_GwdGridComp.F90#L816-L818

I did this for ease of testing. The "correct" thing to do is to actually put this file in $SHARE so that it can be accessed through `scratch/ExtData/` via `linkbcs`. @JulioTBacmeister: will this file change in the future? Or is it stable enough that I can work with the @mssolomo to move it up to the `$SHARE/gmao_ops` area?

PR #315, opened 2020-07-09
Feature/san akel/final stress rebase with develop 2020 jun16

sanAkel:
To be able to use MOM6


GEOSogcm_GridComp/GEOS_OgcmGridComp.F90
```diff
@@ -915,32 +917,36 @@ subroutine SetServices ( GC, RC )
...
   if(DO_OBIO/=0) then
 
-     ! Ocean to OceanBio
-     call MAPL_AddConnectivity ( GC,   &
-          SHORT_NAME  = (/'DH', 'T ', 'S '/),     &
-          DST_ID = OBIO,               &
-          SRC_ID = OCEAN,              &
-          RC=STATUS  )
-     VERIFY_(STATUS)
+!   if (trim(OCEAN_NAME) == "MOM") then  ! MOM5 only
```
yvikhlya, line 941: @sanAkel Why this `if` is commented?

GEOSogcm_GridComp/GEOSocean_GridComp/GuestOcean_GridComp/MOM6_GEOSPlug/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+  )
+
+esma_add_subdirectories(
+  mom6_cmake
+  )
+
+esma_add_library (${this}
+  SRCS ${SRCS}
+  DEPENDENCIES MAPL fms_r8 mom6
+# TYPE SHARED
```
yvikhlya, line 14: It is better to remove this comment. In MOM5 plug CMakeLists too.

GEOSogcm_GridComp/GEOS_OgcmGridComp.F90
sanAkel: @yvikhlya as you know, these variables cannot be exchanged between new MOM6 ocean model and ORAD/OBIO since MOM6 ocean does now allow us to export `DH, T, S`, similarly for `SWHEAT` since it takes net surface heating from UV/IR shortwave radiation bands. Therefore it only works with MOM5 ocean model, which by its design has its very well known limitations, thus superseded by MOM6 (vertical coordinate choices). So the natural thing was to put in the `if` logic to make this exchange only with MOM5. However, the reason for commenting it is for the sake of next project: Data Atmosphere and support OBIO/ORAD. That is why I felt it best to leave these connections in their present form: without if logic (as it is the present develop). However didn't get rid of the `if (if (trim(OCEAN_NAME) == "MOM") then` to remind ourselves of the follow-up project of refactoring/cleaning OGCM GC.

GEOSogcm_GridComp/GEOSocean_GridComp/GuestOcean_GridComp/MOM6_GEOSPlug/CMakeLists.txt
sanAkel: It took us a lot of trial and error to get these shared (ocean models) and static (plugs) configuration. 

With that comment, we are reminded of that. I'm sure sometime down the road we may forget what it took to get here, or if we don't forget, when it is time redo something, e.g., add MITgcm ocean or get rid of MOM5 ocean, this will serve us as a reminded.

tclune:
Yes - I assumed the comment is because we are not 100% certain that STATIC works yet.

GEOSogcm_GridComp/GEOS_OgcmGridComp.F90
yvikhlya: Obio will not run with MOM6 anyway, so I would be even better to put some assert to make sure that if we run obio, we run MOM5. We can figure this out later.

sanAkel: Indeed later is better for now. At that time we can consider some new ideas... e.g., a whole sale data ocean with 3D T, S, DH, etc for OBIO to run no matter what... when we get there...

yvikhlya:
@sanAkel 

https://github.com/GEOS-ESM/GEOSgcm_GridComp/tree/feature/yury/mom6/GEOSogcm_GridComp/GEOSocean_GridComp/GuestOcean_GridComp/MOM6_GEOSPlug/mom6_app

Does this look good to you? Can it go into this PR?

sanAkel:
> @sanAkel
> 
> https://github.com/GEOS-ESM/GEOSgcm_GridComp/tree/feature/yury/mom6/GEOSogcm_GridComp/GEOSocean_GridComp/GuestOcean_GridComp/MOM6_GEOSPlug/mom6_app
> 
> Does this look good to you? Can it go into this PR?

@yvikhlya this is good. Would you please commit to this PR? Thank you!

yvikhlya:
Pushed one more update which copies MOM6 configs into install/etc/MOM6.

PR #321, opened 2020-07-28
Updates to allow GEOS to build without MKL

mathomp4:
This set of PRs is to allow GEOS to build without MKL. Note that it currently does this by making RRTMGP unusable if MKL is not found. This is not ideal, but it's a first attempt.

Should be taken in concert with:
~https://github.com/GEOS-ESM/ESMA_cmake/pull/115~
~https://github.com/GEOS-ESM/NCEP_Shared/pull/8~

ETA: Turns out thanks to CMake knowledge of @tclune this is no longer true! Since each component specifies what they need, they are now independent!


tclune (changes requested):
Please add `find_package(MKL)` in CMakeLists.txt files for layers that need MKL.   This won't help much in the short term, but eventually will help disentangle dependencies for things like MAPL.

tclune (approved):

tclune (approved):
It is unfortunate that the CMake kludge for MKL Fortran is propagating - esp. since the right place to handle this should be inside the MKL package.

The loop (lines 3495-3510) wrapped by `#ifdef` seems suspicious.    Ideally we just have a different RNG and use it the same way as with MKL. ?

mathomp4:
@tclune Yeah. I'm not saying this is the *best* fix, just a workaround. If you, me, or @dr0cloud could come up with an alternative RNG with the same interface, that might be best of all. Plus, it can make @dr0cloud 's life easier with merges with the parent RRTMGP repo if we can limit the ifdefs.

sdrabenh (approved):

PR #333, opened 2020-08-17
Fixes #332. Bug fix for GCC 10 and DRHOS0

### LINKED ISSUE #332: GCC 10 has issues with DRHOS0 variable
This is an odd one. For unknown reasons, my GCC 10 tests at night are showing a non-zero-diff comparison of the `geosgcm_snowlayer` collection compared to its own baseline. This is the **only** collection where this happens and only this variable. 

When you investigate further, it is `DRHOS0` that is causing the issue:
```
               Date     Time   Level Gridsize    Miss    Diff : S Z  Max_Absdiff Max_Reldiff : Parameter name
     1 : 2000-04-15 09:00:00       1     4704    3357      10 : F T     0.051749     0.75000 : DRHOS0
     2 : 2000-04-15 09:00:00       2     4704    3357       4 : F T   9.9492e-43    0.052359 : DRHOS0
     3 : 2000-04-15 09:00:00       3     4704    3357      11 : T T       9.6729     0.26099 : DRHOS0
     4 : 2000-04-15 09:00:00       4     4704    3357       9 : T T       9.6737     0.25013 : DRHOS0
     5 : 2000-04-15 09:00:00       5     4704    3359       7 : T T       9.6741      1.0000 : DRHOS0
     6 : 2000-04-15 09:00:00       6     4704    3359       7 : T T       9.6742      1.0000 : DRHOS0
     7 : 2000-04-15 09:00:00       7     4704    3359       7 : T T       9.6743      1.0000 : DRHOS0
     8 : 2000-04-15 09:00:00       8     4704    3357       7 : T T       9.6744      0.0000 : DRHOS0
     9 : 2000-04-15 09:00:00       9     4704    3357       7 : T T       9.6746      0.0000 : DRHOS0
    10 : 2000-04-15 09:00:00      10     4704    3357       7 : T T       9.6748     0.88443 : DRHOS0
    11 : 2000-04-15 09:00:00      11     4704    3357       5 : F T       0.0000      1.0000 : DRHOS0
    12 : 2000-04-15 09:00:00      12     4704    3357       7 : T T       9.6750      1.0000 : DRHOS0
    13 : 2000-04-15 09:00:00      13     4704    3357       7 : T T   4.3057e+08      1.0000 : DRHOS0
    14 : 2000-04-15 09:00:00      14     4704    3357       7 : T T   2.6394e+07      1.0000 : DRHOS0
    15 : 2000-04-15 09:00:00      15     4704    3359       7 : T T       5.9890      1.0000 : DRHOS0
  15 of 150 records differ
```

The thing is, the differences seem to be different on different days. I'll try and figure out what might be happening if I can (though it might be a compiler bug?). I'm fairly certain it only happens with `Release` flags and not `Debug`.

If I get stuck, I might ping @smahanam or another Land scientist.

mathomp4:
This PR fixes a bug (#332) with GCC 10 and `DRHOS0`. 

The problem was that in StieglitzSnow.F90, `drho0` is initialized too late. In the code, at the top is an if-test:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/2590c4fdaece1901b6bef6132177eb67126f408b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L394
with a `return` built-in:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/2590c4fdaece1901b6bef6132177eb67126f408b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L441-L443

But, `drho0` was initialized too far down:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/2590c4fdaece1901b6bef6132177eb67126f408b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L830

Thus, the code could return without `drho0` being set to anything and apparently with GCC 10, that could be a denormal number (or even NaN as seen on macOS).

This PR initializes *ALL* the `intent(out)` members of the routine that are not in the initial set of "set to 0" at the top:
```fortran
       rmelt = 0.0
       mltwtr = 0.0
       drho0 = 0.0
       tksno = 0.0
```
as a safety.


mathomp4:
Currently going to do two tests:

1. Fully rebuild GCC 10 and make sure it still is consistent
2. Run this code within Intel and make sure it's zero-diff (🤞 )

If these pass, I'll ping @zhaobin74 to try it out in his code base(s) and make sure it works for him.

mathomp4:
I suppose I might have to ping @gmao-rreichle as well. Make sure this doesn't do anything untoward to GEOSldas...if they use this code?

mathomp4:
Okay. I am going to pull this out of draft. It seems zero-diff with Intel (as if Intel was flushing-to-zero denormals, it would be). 

But, I'm not sure how to label this. The *state* of the model doesn't care about `DRHOS0` as it's just diagnostic, but there is a chance on Intel and a guarantee on GCC 10 that the `geosgcm_snowlayer` collection (or any other collection that has it) could change a few values at the odd points where this was triggered.

I also don't know if `DRHOS0` is a state variable for GEOSldas. I'll take a look, but I might need @gmao-rreichle to confirm/deny.

mathomp4:
> I also don't know if `DRHOS0` is a state variable for GEOSldas. I'll take a look, but I might need @gmao-rreichle to confirm/deny.

I just took a look through GEOSldas and apparently it doesn't care about `DRHOS0` (no reference to it) or even the `drho0` from StieglitzSnow. It can be garbage and the GEOSldas will not mind one way or the other from my reading. `drho0` is just an array that comes and goes. 😄

mathomp4:
@zhaobin74 @gmao-rreichle  I guess I didn't do that because there are other places things happen in twos a la:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/2590c4fdaece1901b6bef6132177eb67126f408b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L373

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/2590c4fdaece1901b6bef6132177eb67126f408b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L827

I was going for the smallest change. But I can remove the other init-to-zero for `drho0` if you like

gmao-rreichle:
Thanks for the clarification, @mathomp4 .  I see where you're coming from...
For reference, here are lines ~825-830 in the bugfix branch:
```
!**** Update snow density by compaction (Pitman et al. 1991)

       excs = 0.
       mass = 0.
       w    = 0.
       drho0 = 0.
```
A quick check suggests that the init of "excs" is the same as that of "drho0", ie., the 2nd init could/should be removed.
The init of "mass" should probably also be moved up, similar to that of "drho0".
But good luck with searching for "w"...
In a nutshell, the code is messy, and the question is simply how much cleanup we want to do at this time.
Perhaps the best course of action would be to clean up the init of "drho0", "excs" and "mass" because that seems to be low-hanging fruit, and add a comment about this cleanup into the code where "w" is initialized (for future reference).
Alternatively, the "w" init could also be moved up along with the others and then subjected to a full 0-diff testing.   Weiyuan could test GEOSldas if that's what you decide to do.

cc: @zhaobin74

zhaobin74:
> Thanks for the clarification, @mathomp4 . I see where you're coming from...
> For reference, here are lines ~825-830 in the bugfix branch:
> 
> ```
> !**** Update snow density by compaction (Pitman et al. 1991)
> 
>        excs = 0.
>        mass = 0.
>        w    = 0.
>        drho0 = 0.
> ```
> 
> A quick check suggests that the init of "excs" is the same as that of "drho0", ie., the 2nd init could/should be removed.
> The init of "mass" should probably also be moved up, similar to that of "drho0".
> But good luck with searching for "w"...
> In a nutshell, the code is messy, and the question is simply how much cleanup we want to do at this time.
> Perhaps the best course of action would be to clean up the init of "drho0", "excs" and "mass" because that seems to be low-hanging fruit, and add a comment about this cleanup into the code where "w" is initialized (for future reference).
> Alternatively, the "w" init could also be moved up along with the others and then subjected to a full 0-diff testing. Weiyuan could test GEOSldas if that's what you decide to do.
> 
> cc: @zhaobin74

I agree, now is the right time to make this right. `excs, mass and drho0 ` should be fixed.

mathomp4:
I pushed a commit that removes the redundant `excs` and `drho0` set-to-zeros. The `mass` and `w` look fine to me as they are local to the routine and only ever referenced inside the if-block that uses them. `excs` and `drho0` are actual outputs of `snowrt` and so should be initialized as soon as possible.

PR #334, opened 2020-08-21
Feature/esherman/add ncl scripts

gmao-esherman:
Two scripts have been added. The getforcing.scr script generates MERRA2 forcing files for the Single Column Model. The gettiledata.ncl script creates a tilefile with multiple tiles closest to the specified latitude and longitude.

These scripts are not executed at compile or run time. They operate independently of the model, and are executed by the user only when they are generating a new MERRA2 experiment for the single column model. These scripts would normally float around @gmao-esherman or @rcullath directories on NCCS, but now they need a more permanent location.

### DIFF
--- a/GEOSagcm_GridComp/GEOSsuperdyn_GridComp/GEOSdatmodyn_GridComp/MERRA2Forcing/getforcing.scr
+++ b/GEOSagcm_GridComp/GEOSsuperdyn_GridComp/GEOSdatmodyn_GridComp/MERRA2Forcing/getforcing.scr
@@ -1,0 +1,612 @@
+#!/bin/csh -f
+
+# This script will generate MERRA2 forcing files for the Singe Column Model.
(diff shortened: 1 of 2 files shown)


mathomp4:
@gmao-esherman I suppose it might be good to put some of the scripts used in the current SCM in Git as well...

gmao-esherman:
> @gmao-esherman I suppose it might be good to put some of the scripts used in the current SCM in Git as well...

@mathomp4 Sure. What scripts do you have in mind?

mathomp4:
> > @gmao-esherman I suppose it might be good to put some of the scripts used in the current SCM in Git as well...
> 
> @mathomp4 Sure. What scripts do you have in mind?

I am not sure. I suppose as you learn more and more about the SCM we can figure out which should go in. Maybe some of the  IDL scripts (from @amolod ?) would be good?

PR #335, opened 2020-08-21
Feature/sm precipfrac pr2

smahanam:
This is a reissue of the earlier PR #256, available here:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/tree/feature/SM_precipfrac_PR

This reissue addresses all recommendations listed in PR 256 and is related to issue [150](https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/150)


gmao-rreichle:
@rdkoster , @smahanam : I edited the explanatory text in the Surface GridComp resource file:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/90d54be48c7be65330f4a9b1609143707b2866d8

I tried to clarify what FWETC and FWETL are, and how they are set, including providing template entries for these rc parameters.   My strong preference is to have template entries for all model parameters that can be set in the rc file, so it's very clear that they can be set.  I had missed this fact in the original prose.  Maybe check with Scott R about GEOS-ESM guidance (or lack thereof).

Probably more importantly, I think we need to make sure that the limitations of the "calibrated" values are expressed clearly in the rc file.  Please correct any mistakes in my understanding. 

cc: @biljanaorescanin

rdkoster:
I made a tiny additional wording change in the SurfaceGridComp resource file comment lines.  Aside from that, I am happy with Rolf's additions.

gmao-rreichle:
Thanks, @rdkoster , for the fix to the comment.  I hadn't noticed that this had been missing in the original revisions. 

I went ahead and merged develop into the branch (per the "Update Branch" button on this PR).

As far as I am concerned, once it has been 0-diff tested for the GCM and for GEOSldas by the developer ( @smahanam ), it's ready to be handed off to the gatekeepers ( @biljanaorescanin , @sdrabenh ) for their review, verification (re-test) and merge.  To hand it off, remove the "Contingent - DNA" label and press the "Ready for Review" button. 

PS: Separate 0-diff testing in the GCM and GEOSldas is necessary because the Surface GridComp resource file has changed.  I do not anticipate that any of the computations fail in GEOSldas if they work in the GCM, but whenever an interface changes it is important to verify that both the GCM and GEOSldas can handle the change properly.

gmao-rreichle:
Something is wrong with the logic applied to the relevant resource parameters in the Catch GridComp:

The most recent commit [aa2e174](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/335/commits/aa2e174bd72ead3f8ef1b0e30719154cd772b7ac) violates the logic of the GEOS_SurfaceGridComp.rc file, which is processed by ldas_setup.
All resource parameters specified here should be prefaced by "# GEOSagcm=>" and "# GEOSldas=>".  See the file header for details.

In creating LDAS.rc, ldas_setup removes "# GEOSldas=>" and consequently inserts the values specified in GEOS_SurfaceGridComp.rc into LDAS.rc as follows:
USE_FWET_FOR_RUNOFF:.FALSE. 
FWETC: 0.005
FWETL: 0.025.
If the above does not yield zero-diff, then the logic of how these resource parameters are used in the new branch of the CatchGridComp is flawed.  
Per the agreed-upon documentation in GEOS_SurfaceGridComp.rc, any FWETC and FWETL settings should be ignored when USE_FWET_FOR_RUNOFF:.FALSE. 
This does not seem to be the case, or the 0-diff test would have worked.

Revisions are needed.

cc: @smahanam @biljanaorescanin @rdkoster @weiyuan-jiang

gmao-rreichle:
I added myself as "assignee" to this pull request because I do not get email notifications for all GCM issues and pull requests, and I hadn't seen the comment by @smahanam until it was pointed out to me by @biljanaorescanin .

The issue is thorny in that we need to satisfy two things at once: (i) the integrity of the Surface GridComp rc file (which needs to work for both the GCM and GEOSldas) and (ii) the appropriate default values for 0-diff testing.

The simplest and most elegant way to achieve this is to be consistent in the default inputs for GEOSldas in the Surface GridComp file.  My most recent commit should fix the issue.

@rdkoster : Please review my changes here: [29c2513](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/335/commits/29c25136a85f7271ed2bff1150c4c8f2e7f2e3c4)

@weiyuan-jiang

gmao-rreichle:
> risky - now default are read from RC file while they are embedded in the code in develop

@smahanam : If you need the flexibility to change the parameter values in the resource file, then maybe the old parameter values shouldn't be hard-coded in the first place... Feel free to remove the hard-coded values from the src code if you prefer. 

I absolutely insist on having  the Surface GridComp file be uniform in design and with clear and consistent documentation.  We are not going back to the mess with which we were presented back in March.

cc: @rdkoster @biljanaorescanin @weiyuan-jiang

gmao-rreichle:
PS: There's nothing wrong with adding an ASSERT statement into the src code, something like (pseudocode):
if USE_FWET_FOR_RUNOFF==false.
   if FWETC /= 0.02 .or. FWETL /= 0.02
       then stop with a clear message   
  end
end
The message should instruct the user that the default values that go with USE_FWET_FOR_RUNOFF==false. are FWETC=FWETL=0.02.  If the user wants to really run with USE_FWET_FOR_RUNOFF==false. but have different values for  FWETC and FWETL, they can disable this stop in their src code.

cc: @smahanam @biljanaorescanin @rdkoster @weiyuan-jiang

biljanaorescanin:
@gmao-rreichle I talked with Sarith and since both him and Randy confirmed these changes are OK I will proceed with testing 
cc: @rdkoster,@smahanam,@weiyuan-jiang

biljanaorescanin:
This branch passed 0diff testing for both GEOSldas and GCM.

gmao-rreichle (approved):

Issue #337, opened 2020-08-27
Why is GuestOcean checking if it is replay


bena-nasa:
I was doing some development that involved running GEOS through the Jedi framework and working on the capability to run multiple model iterations from one execution by rewinding GEOS and restoring the previously saved state.

As I was doing this I was finding I could not rewind, integrate forward in time again and get back to the same state until Atanas reminded me that we have this private clock in the GuestOcean component. At some point, and I can not remember why, I put a check in that says, if you are doing replay, check if the private clock is ahead of the passed in clock and rewind the private clock back to the current time of the passed in clock. This is needed during replay as we have the extra predictor step. This takes place in the run method of GuestOcean around line 1018.
```
    call MAPL_GetResource(state,ReplayMode,  'REPLAY_MODE:',  default="NoReplay", RC=STATUS )
    if (trim(replayMode)=="Regular" ) then
       if (myTime > EndTime) then
          call ESMF_ClockSet(PrivateState%Clock,direction=ESMF_DIRECTION_REVERSE,rc=status)
          VERIFY_(status)
          do
            call ESMF_ClockAdvance(PrivateState%Clock,rc=status)
            VERIFY_(status)
            call ESMF_ClockGet(PrivateState%Clock,currTime=ct,rc=status)
            VERIFY_(status)
            if (ct==endTime) exit
          enddo
          call ESMF_ClockSet(PrivateState%Clock,direction=ESMF_DIRECTION_FORWARD,rc=status)
          VERIFY_(status)
          call ESMF_ClockGet( PrivateState%CLOCK, currTime=myTime, RC=STATUS)
          VERIFY_(status)
       end if
```

To get the Jedi test to pass I had to remove the restriction to only do this during replay. But as I was looking at this, I can't see why we need to check if it is replay at all. The only way the private clock could be ahead of the passed in clock is replay (and now Jedi). I can't think if why you would NOT want to rewind the private clock if it finds itself head of the passed in clock.

I thought Atanas said the only reason Guest even has a private clock is that you might want to run the ocean at a different timestep than the atmosphere so it might have to catch up. Indeed right after checking if we need to rewind the clock, the logic is, run the underlying ocean and advance my private clock to the time of the passed in clock. So there is no way the clock in Guest could ever find itself ahead of the clock passed in unless something like replay or Jedi has manipulated it so why have the logic to do the rewind only in replay.

[...]

Issue #322, opened 2020-07-28
Relaxed ocean and Atmospheric Replay


sanAkel:
- [ ] capability needs to be added 
- [x] MOM6 plug: `subroutine Record` needs to be checked. Needed only if we want to a write restart while the run is still proceeding

yvikhlya:
We need to have another tile file with lat-lon and MOM6 tripolar grids to interpolate SST and FRACI data for dual ocean set up.

sanAkel:
> We need to have another tile file with lat-lon and MOM6 tripolar grids to interpolate SST and FRACI data for dual ocean set up.

Requesting: @atrayano 

@yvikhlya please add a reference to the `assignee` so he/she gets a notification. Thanks!

sanAkel:
- [x] MOM6 plug: subroutine Record needs to be checked. Needed only if we want to a write restart while the run is still proceeding is **done**

sanAkel:
The atmospheric **reply** will/should work as it does now with Dual Ocean.

Things to do for implementing dual ocean with MOM6:

- Add RUN2 in MOM6 plug that takes DELTA_SST as import from Guest Ocean GC; DELTA_SST is the difference between MOM's SST and SST(from _data_)

- We need to have a mechanism for relaxing MOM's SST, which could be accomplished in one of the following ways:

1. Using DELTA_SST (import from Guest Ocn), then compute an equivalent heat flux due to restoration and add that heat flux to **one** of the shortwave radiation bands (only one of either: vis dir/dif, nir dir/dif).
The way to compute _this_ heat flux would follow `coupled_driver/MOM_surface_forcing_gfdl.F90`, look for: `CS%restore_temp`

2. Compute the heat flux due to restoration as in above, but add it MOM's forcing (type), look for: `CS%restore_temp` in `coupled_driver/MOM_surface_forcing_gfdl.F90` this will require following the logicals in MOM and being able to access public (hopefully *not* private!) types.

yvikhlya:
@sanAkel We already have a relaxation of MOM SST to data SST in guest (see DEL_TEMP field). Why do we need to do it different way in MOM6?

sanAkel:
[...]

PR #338, opened 2020-08-27
Remove clock rewind only in replay logic in Guest

bena-nasa:
A proposed small change in GuestOcean, detailed description in issue #337 ,there is logic that I just can't see any justifiable reason for having unless I'm missing something.
To support future work when running GEOS in Jedi this either needs to be removed or we have to somehow have a 2nd conditional in addition to if running replay. I can't see why we need any condition at all around the rewind.

In testing both replay and AMIP it is zero diff.

### DIFF
--- a/GEOSogcm_GridComp/GEOSocean_GridComp/GuestOcean_GridComp/GuestOcean_GridComp.F90
+++ b/GEOSogcm_GridComp/GEOSocean_GridComp/GuestOcean_GridComp/GuestOcean_GridComp.F90
@@ -1015,23 +1015,20 @@ contains
     call ESMF_ClockGet( PrivateState%CLOCK, currTime=myTime, RC=STATUS)
     VERIFY_(status)
 
-    call MAPL_GetResource(state,ReplayMode,  'REPLAY_MODE:',  default="NoReplay", RC=STATUS )
-    if (trim(replayMode)=="Regular" ) then
-       if (myTime > EndTime) then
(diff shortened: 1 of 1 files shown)


yvikhlya:
We also need to test dual ocean.

sanAkel:
For what it is worth, I would suggest following tests:

- no `dual` ocean:
  + [ ] MOM5
  + [ ] MOM6
- with `dual` ocean:
  + [ ] MOM5
  + MOM6 `dual` ocean is WIP https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/322

bena-nasa:
I did confirm that coupled/MOM5 is zero-diff with this change. How do you run the dual ocean with MOM5? I can try MOM6 but I just have a hard time conceiving how this can be non-zero diff for that given the logic in Guest.

atrayano:
Yesterday I took a day off and missed the entire discussion. Anyway, Ben and I had a discussion today: the rewinding of the clock is not related in any way to dual ocean. It was leftover code (by me) to restore the local clock if somehow the clock was in left in the "future" state (the only possibility for this to happen until now was doing replay. And JEDI). I am totally fine with the proposed change. There is a remote possibility that the model clock and the local clock can be configured (by mistake) so that the local clock "overshoots", and then this in turn will trigger the new logic, which would lead to very undesirable consequences. At the moment we are coupling the atmosphere and the ocean on the heartbeat so this is not a concern. Once we decide to couple less frequently, we need to make sure the substepping interval is multiple of coupling interval

PR #340, opened 2020-08-31
Feature/pnorris/#336 rrtmgp to master aug2020

dr0cloud:
Minor changes to bring RRTMGP up to date with Rob Pincus Master of Aug2020.
Largely adding system to allow turning off of bounds and value checking; by default these are still on.
Verified zero-diff both when RRTMGP is run or not. RRTMGP still off by default.


GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSradiation_GridComp/GEOS_RadiationShared/RRTMGP/rte/mo_rte_sw.F90
```diff
@@ -84,32 +85,41 @@ function rte_sw(atmos, top_at_1,                 &
...
+      if(any_vals_less_than(inc_flux, 0._wp)) &
+        error_msg = "rte_sw: one or more inc_flux < 0"
+      if(any_vals_outside(sfc_alb_dir,  0._wp, 1._wp)) &
+        error_msg = "rte_sw: sfc_alb_dir out of bounds [0,1]"
+      if(any_vals_outside(sfc_alb_dif,  0._wp, 1._wp)) &
+        error_msg = "rte_sw: sfc_alb_dif out of bounds [0,1]"
+      if(present(inc_flux_dif)) then
+        if(any_vals_less_than(inc_flux_dif, 0._wp)) &
+          error_msg = "rte_sw: one or more inc_flux_dif < 0"
+      end if
```
tclune, line 120: I was really just reviewing the cmake files, but sections like this caught my eye.

Is it the intent that each check may overwrite the `error_msg` of the previous check?   If you are certain that only one can hold then `else if` would make the logic clearer to others.   Also, I'm assuming that somewhere above `error_msg` is set to the empty string prior to these checks?

dr0cloud: Hi Tom, its Rob Pincus' code that I'm just importing. I guess he's reasoning that one might as well deal first with the last error in a group rather than the first (which is all that we usually get if we die at the first error). Eventually, one gets them all as they are successively fixed.). It could be called "the first shall be last" error doctrine!
Re the error_msg, yes it starts off "" and whenever if is set to something else, an error is eventually triggered. So it remains "" at the beginning of each new group of tests.

sdrabenh (approved):

dr0cloud:
Thanks Scott! Peter.

PR #341, opened 2020-09-04
Genericize NCAR GWD Beres file location

mathomp4:
Removes hardcoded BERES_FILE_NAME and makes it an ExtData file from AGCM.rc

Should be taken in concert with https://github.com/GEOS-ESM/GEOSgcm_App/pull/163


mathomp4:
Shown zero-diff both with `USE_NCAR_GWD: .FALSE.` and `USE_NCAR_GWD: .TRUE.`

sdrabenh (approved):

Issue #311, opened 2020-07-07
MKIAU_JCAP broken


wmputman:
When we do replays I am no longer able to use the MKIAU_JCAP option.

GEOSmkiau_GridComp/GEOS_mkiauGridComp.F90

 Integer*4 Resource Parameter: MKIAU_JCAP:60
 Current nymd:     20200120   nhms:            0   FAC:  1.00000
 Creating GRIDana...
 Integer*4 Resource Parameter: BKG2ANACNSRV:0
 Integer*4 Resource Parameter: ANA2BKGCNSRV:0
 CFIO: Reading
 /discover/nobackup/projects/gmao/g6dev/sdrabenh/valdat/era5/ana_eta_daily/Y2020
 /M01/era5_ana.eta_L137.20200120_00z.nc4 at     20200120             0

 REPLAY File Dimensions:         1440         721         137

 REPLAY File Variables, NQ:            7
 --------------------------
           1 )  dp
           2 )  phis
           3 )  ps
           4 )  qv
           5 )  tv
           6 )  u
           7 )  v

 REPLAY Options:
 ---------------
 REPLAY_TS ....... NO
 REPLAY_P ........ YES
 REPLAY_U ........ YES
 REPLAY_V ........ YES
 REPLAY_T ........ YES
 REPLAY_QV ....... YES
 REPLAY_O3 ....... NO

 Vertical Remapping ANA Data to BKG Topography and Levels ...

 Blending Between    1.000000      &    50.00000      mb
 No blending of QV based on TROPP

 Applying Mass Divergence Fix ...

[borgf058:162020:0:162020] Caught signal 11 (Segmentation fault: address not mapped to object at address 0x2aabc9579000)
==== backtrace ====
[borgf059:247038:0:247038] Caught signal 11 (Segmentation fault: address not mapped to object at address 0x10)
==== backtrace ====
[borgf143:47492:0:47492] Caught signal 11 (Segmentation fault: address not mapped to object at address 0x10)
[borgf050:166372:0:166372] Caught signal 11 (Segmentation fault: address not mapped to object at address 0x10)
==== backtrace ====

bena-nasa:
Breaks at c24 luckily when I enabled JCAP in a standard replay from MERRA2

mathomp4:
@bena-nasa I think we've been here before. From #134 (this [comment](https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/134#issuecomment-551985778):

[...]

PR #349, opened 2020-09-18
Updates to CatchCN GridComp so it can reproduce Fanwei's science-tested LDASsa version of CatchCN

gmao-jkolassa:
This branch has the objectives of (1) being able to reproduce Fanwei Zeng's published LDASsa Catchment-CN simulations and (2) introducing two minor code modifications.

The changes to be able to reproduce Fanwei's Catchment-CN simulations affect the following files:

**SurfParams.F90:**
•	Introduced separate parameter collections for different land model choices that are being cross-checked against the land model choice
•	Catchment-CN collections include additional LAND_FIX_CN parameter that adjusts part of catchmentCN.F90 and GEOScatchCN_GridComp.F90 to reflect what was used in Fanwei Zeng’s experiments (to reproduce Fanwei's experiments, LAND_FIX_CN needs to be set to .FALSE.)

**catchmentCN.F90**
•	FLWALPHA and RSWILT are being read from SurfParams instead of being hardcoded
•	LAND_FIX_CN toggle is introduced as equivalent of LAND_FIX in catchment.F90 as well as to allow for changing code to what was used in Fanwei Zeng’s experiments. 

**GEOScatchCN_GridComp.F90**
•	Added LAND_FIX_CN read from SurfParams to toggle RDC formulation in order to match Fanwei Zeng’s experiments

**GEOS_SurfaceGridComp.F90**
•	Added LSM_CHOICE to SurfParams_init call

**dbg_cnlsm_offline.F90**
•	Added LSM_CHOICE to SurfParam_init call (hardcoded as Catchment_CN, since this is a CN specific routine)

The minor code modifications affect the following routines:

**GEOS_SurfaceGridComp.rc**
•	Added default value of -9999 for C02_YEAR

**mk_Restarts**
•	Removed contstraint to Haswell nodes

cc: @gmao-rreichle , @smahanam , @rdkoster , @biljanaorescanin , @gmao-jkolassa


gmao-jkolassa:
This pull request is linked to GEOSldas pull request #311 [https://github.com/GEOS-ESM/GEOSldas/pull/311](url).

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
```diff
@@ -3379,7 +3380,8 @@ subroutine Initialize ( GC, IMPORT, EXPORT, CLOCK, RC )
 
 ! Init land and snow constants, currently different in Icarus and GEOSldas
 !-----------------------------------------------------------------------
-
+    call MAPL_GetResource ( MAPL, LSM_CHOICE, Label="LSM_CHOICE:", DEFAULT=1, RC=STATUS)
```
gmao-rreichle, line 3383: I might be missing something, but why do we need to find out LSM_CHOICE if it is not used later in the subroutine?
My best guess is that the call to  SurfParams_init() two line down needs to be changed to read:
`call SurfParams_init( LAND_PARAMS, LSM_CHOICE)`
If that is indeed the case, then I don't understand how this branch could have compiled properly (because LSM_CHOICE is not an optional argument for SurfParams_init(), which suggests that I'm missing something.  But still, something doesn't look quite right here.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/SurfParams.F90
```diff
@@ -45,7 +49,7 @@ subroutine SurfParams_init(LAND_PARAMS)
        return
     end if
       
-
+    if (LSM_CHOICE==1) then
```
gmao-rreichle, line 52: Please fix the indentation in the "if" block.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/SurfParams.F90
```diff
@@ -81,10 +85,70 @@ subroutine SurfParams_init(LAND_PARAMS)
...
-    end select
+      case DEFAULT                                                                            
+         write (*,*) "LAND_PARAMS not valid or incompatible with LSM_CHOICE ",LAND_PARAMS     
+         return                                                                               
+      end select                                                                                
+                                                                                              
+    else if (LSM_CHOICE==2) then       
+
+       select case (LAND_PARAMS)                                                                
+       case ("e0004s_transientCO2_05")  ! parameters to reproduce Fanwei Zeng's Catchment-CN.4.0 runs (e0004s_transientCO2_05) done with build /gpfsm/dnb31/fzeng/LDASsa_m3-16_0_p2_CatchCatchCN_for_MERRA3
```
gmao-rreichle, line 96: Can we come up with a more intuitive name for this (only) science-tested version of CatchCN?
"e0004s_transientCO2_05" sounds just too awkward for something that would be the model version of reference going forward.
Maybe we can refer to the peer-reviewed paper, as in `case ("Lee_et_al_BG_2018")`??
(assuming this is the model version used in [doi:10.5194/bg-15-5635-2018](http://doi.org/10.5194/bg-15-5635-2018)
I'm open to other suggestions...

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOS_CatchCNGridComp.F90
```diff
@@ -5820,7 +5821,9 @@ subroutine Driver ( RC )
         allocate(RSL2     (NTILES)) 
         allocate(SQSCAT   (NTILES))
         allocate(RDC      (NTILES))  
-	allocate(UUU      (NTILES))
+	allocate(RDC_TMP_1(NTILES))
```
gmao-rreichle, line 5824: fix indentation (throughout)

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/catchmentCN.F90
```diff
@@ -549,13 +549,14 @@ SUBROUTINE CATCHCN (                                           &
         tc4_orig(n)=tc4(n)
 
         ! Andrea Molod (Oct 21, 2016):
-
+        if (LAND_FIX_CN) then
```
gmao-rreichle, line 552: While Fanwei's code didn't have these protections in this "if" block, I doubt that using the protections makes much of science difference. 
The same protections are used in the latest version of Catchment (FP 5.25, SMAP L4).
For consistency between Catchment and CatchmentCN, I suggest changing the "if" statement to:
   `if (LAND_FIX) then`

removing the "if" statement again and always execute the statements in the "if"

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
gmao-jkolassa: Yes, the call to SurfParams_init needs to read call SurfParams_init( LAND_PARAMS, LSM_CHOICE). I am not sure why it does not. I'll commit and push that change again.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/catchmentCN.F90
```diff
@@ -1399,7 +1400,7 @@ SUBROUTINE CATCHCN (                                           &
           ELSE
             TC4(N)=TC4(N)+SIGN(10.,TC4X-TC4(N))
           ENDIF
-
+       ENDIF ! LAND_FIX_CN and ASNOW=0
```
gmao-rreichle, line 1403: change comment to match "if" statement above (currently inconsistent)

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/SurfParams.F90
gmao-jkolassa: How about "CN_CLM4", since the experiments that we are trying to match are really the only science-tested published version of CatchCN with CLM4 at this point. Otherwise, "Lee_et_al_BG_2018" is also fine by me.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/catchmentCN.F90
```diff
@@ -1363,7 +1364,7 @@ SUBROUTINE CATCHCN (                                           &
                            DHSDTC4(N)
 
 !       Ensure that modifications made to QA and TC are not too large:
-
+        IF ( (.NOT. LAND_FIX_CN) .OR. (ASNOW0(N) .EQ. 0. ) ) THEN
```
gmao-rreichle, line 1367: This seems wrong to me.  In subroutine catchment(), we have the following:
```
! JP added patch to exlude correction under snow:
        IF ( (.NOT. LAND_FIX) .OR. (ASNOW0(N) .EQ. 0. ) ) THEN          
           IF(ABS(QA1X-QA1(N)) .LE. 0.5*QA1(N)) THEN
              QA1(N)=QA1X
           ELSE
              QA1(N)=QA1(N)+SIGN(0.5*QA1(N),QA1X-QA1(N))
           ENDIF
```
This patch disables the calculations under snow in the recent model version.  (LAND_FIX is true for the recent model versions, then the if statement comes down to ASNOW==0)
The "if" statement here would always apply the calculations, regardless of snow, which persists the old soil-temperature-under-snow bug.
The correct "if" statement should be the same as in catchment(), I think.
While this may be a departure from Fanwei's model version, it is a much-needed bug fix.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/dbg_cnlsm_offline.F90
```diff
@@ -237,7 +237,7 @@ PROGRAM dbg_cnlsm_offline
 
 is_OFFLINE = OFFLINE_MODE /= 0
 
-call SurfParams_init("GCM") 
+call SurfParams_init("GCM",2)
```
gmao-rreichle, line 240: For clarity, I suggest adding a comment that LSM_CHOICE=2 implies CatchCN
(or just define LSM_CHOICE as a local integer parameter.
Also, I doubt we would want to hardwire "GCM" as the value for LAND_PARAMS going forward...
This should likewise be identified as LAND_PARAMS.

gmao-rreichle (changes requested):
Some elements need closer inspection.  I added in-line comments.
More broadly, there probably should be more comments that explain what the preferred land parameter settings are for CatchCN going forward, and that (I think) none of the older settings were ever used for science.
This makes me wonder if we shouldn't just start over and remove the old choices, which would make this a non-zero-diff change in the current GEOSldas test suite, but the whole point of the exercise was to come up with a science-tested version of CatchCN that can serve as meaningful test case.
For now let's address the individual comments first, then take it from there.

gmao-jkolassa:
Made changes following Rolf's suggestions. These include:

**SurfParams.F90:**
- fixed indentation
- renamed parameter collection to reproduce Fanwei Zeng's experiments to "CN_CLM4"

**catchmentCN.F90**
- removed QA range control toggle with LAND_FIX_CN; range control is now always executed
- changed under snow calculations to be controlled by LAND_FIX rather than LAND_FIX_CN and to be consistent with catchment.F90

**GEOScatchCN_GridComp.F90**
- fixed indentation

**GEOS_SurfaceGridComp.F90**
- added missing LSM_CHOICE argument to SurfParams_init call

**dbg_cnlsm_offline.F90**
- removed hardcoding of LAND_PARAMS choice 
- defined LSM_CHOICE parameter (still hardcoded as Catchment-CN)

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOS_CatchCNGridComp.F90
```diff
@@ -6203,7 +6203,7 @@ subroutine Driver ( RC )
         ! --------------------------------------------------------------------------
         ! LAI and type dependent parameters; RDC formulation now uses veg fractions gkw: 2013-11-25, see note from Randy
         ! --------------------------------------------------------------------------
-        IF (LAND_FIX_CN) THEN
+        IF (RDC_FORMULATION==0) THEN
```
gmao-rreichle, line 6206: Sorry, I forget why we wanted to keep the option of using RDC_formulation==0.
Assuming we do have a use for it,  I think we should make RDC_formulation==1 explicit, that is, allow only 0 or 1, and put a stop with an exit message if RDC_formulation is anything other than 0 or 1.
As it stands, setting RDC_formulation=2 would results in the same outcome as RDC_formulation=1, which will be confusing if we ever want to change it again and need to distinguish between RDC_formulation=1 and 2 for backward compatibility.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/GEOS_SurfaceGridComp.rc
```diff
@@ -50,7 +50,8 @@
 #
 #    Icarus  : Icarus/f522_fp and earlier (scientifically close to MERRA-2)
 #    V24_C05 : GEOSldas_m4-17_0
-#    NRv7.2  : SMAP L4_SM Version 4, f525_fp
+#    NRv7.2  : SMAP L4_SM Version 4 and 5, f525_fp
+#    CN_CLM4 : parameters to reproduce Fanwei Zeng's science-validated and published Catchment-CN.4.0 runs (e0004s_transientCO2_05)
```
gmao-rreichle, line 54: I think we need to make it clear here, that "Icarus", "V24_C05" and "NRv7.2" are for Catchment only and "CN_CLM4" is for CatchmentCN only.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/SurfParams.F90
```diff
@@ -94,53 +93,17 @@ subroutine SurfParams_init(LAND_PARAMS,LSM_CHOICE)
...
-         	WEMIN    = 13.                                                                       
-         	AICEV    = 0.149
-         	AICEN    = 19.851                                                                    
-         	FLWALPHA = 0.005  ! set to previously hardcoded CatchCN value                           
-         	ASTRFR   = 0.333  ! reverted  
-         	STEXP    = 1.     ! reverted     
-         	RSWILT   = 2000.  ! set to previously hardcoded CatchCN value                        
-                                                                                              
+         	LAND_FIX        = .TRUE.                                                                    
+         	RDC_FORMULATION = 1 ! 1: Fanwei's Catchment-CN version; 0: Catchment version
```
gmao-rreichle, line 97: Is the parameter "RDC_formulation" used at all in Catchment?
Maybe drop "0: Catchment version" here because it's confusing.
If there isn't a "LAND_PARAMS" option that uses RDC_formulation=0, do we even still need it?

gmao-jkolassa: Rolf,

thanks for your comments. I will address them this afternoon.

For this particular comment here, I believe we have already talked about this in one of our past Friday meetings and decided to keep the option to run the two different RDC_formulations for possible future experiments (with new parameter collections). The reasons was to keep a version that corresponds to what is used in the current GEOSldas Catchment-CN.

Did you want to reconsider the decision we made then?

gmao-jkolassa:
I have made the following changes:

**GEOS_SurfaceGridComp.rc:**
- Updated the land parameter documentation to identify which collections are to be used with Catchment and Catchment-CN, respectively.

**SurfParams.F90:**
- Changed the description of the RDC_FORMULATION options to reflect that they correspond to the formulation in Fanwei's LDASsa version and the previous GEOSldas Catchment-CN, respectively.

**GEOScatchCN_GridComp.F90**
- Modified the if-statement to check explicitly for RDC_FORMULATION options 0 and 1 and stop otherwise.

I have not yet addressed Rolf's suggested removal of the RDC_FORMULATION toggle. The main reason is that we had previously talked about this in a meeting and decided to keep both options. So I just wanted to verify first that this decision has changed before implementing the change.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/SurfParams.F90
```diff
@@ -15,26 +17,27 @@ module SurfParams
...
+  REAL, PUBLIC, SAVE :: CSOIL_2      		 ! J/K - heat capacity associated w/ tsurf  
+  REAL, PUBLIC, SAVE :: WEMIN        		 ! kg/m^2  minimum SWE in areal fraction
+  REAL, PUBLIC, SAVE :: AICEV, AICEN 		 ! Used in StieglitzSnow
+  REAL, PUBLIC, SAVE :: FLWALPHA     		 ! SRFLW multiplier with SRFLW < 0
+  REAL, PUBLIC, SAVE :: ASTRFR, STEXP		 ! stress parameters in energy2
+  REAL, PUBLIC, SAVE :: RSWILT       		 ! parameters in rsurfp
+
+  LOGICAL, PUBLIC, SAVE :: LAND_FIX  		 ! Used for fixes and init changes that
+  	   	   	   	     		 ! are still default in Icarus GCM
+  INTEGER, PUBLIC, SAVE :: RDC_FORMULATION	 ! controls RDC computation in Catchment-CN
```
gmao-rreichle, line 29: I think this line is now obsolete.

gmao-jkolassa:
Removed the option to toggle between different calculations of RDC value. This change affects two files:

GEOScatchCN_GridComp.F90: where the RDC value is calculated
SurfParams.F90: where the parameter to toggle (RDC_FORMULATION) was defined

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/SurfParams.F90
```diff
@@ -45,49 +47,72 @@ subroutine SurfParams_init(LAND_PARAMS)
...
+         	AICEV    = 0.149 
+         	AICEN    = 19.851
+         	FLWALPHA = 0.04
+         	ASTRFR   = 0.333  ! reverted 
+        	STEXP    = 1.     ! reverted
+         	RSWILT   = 500.   ! reverted
+
+      	case DEFAULT                                                                            
+         	write (*,*) "LAND_PARAMS not valid or incompatible with LSM_CHOICE ",LAND_PARAMS     
+         	return
```
gmao-rreichle, line 88: This comment is not about changes made by @gmao-jkolassa in connection with the main task of the pull request, and I might have missed something in my quick review.
In any case, it looks like SurfParams_init() simply returns to the calling subroutine without (re)setting any of the relevant parameters if the LAND_PARAMS string is unknown.
The relevant model parameters may or may not have been set before the call to SurfParams_init(), as default values or otherwise, and it's possible that the code will keep running using whatever values or defaults are set outside of SurfParams_init().  However, this is almost certainly a model configuration that isn't scientifically viable.
I don't think this is the desired behavior.  
If an unknown value for LAND_PARAMS is selected, shouldn't we exit?  That is, shouldn't we replace the "return" statement with an "ASSERT(.false.)" or "stop" statement?
@gmao-jkolassa , can you please look into this? 

cc: @weiyuan-jiang @biljanaorescanin

gmao-jkolassa: Rolf,
you are correct that if LAND_PARAMS and LSM_CHOICE are not compatible or the LAND_PARAMS collection is unknown, the model keeps running in the way it is currently coded. It does fail later, since no parameters have been defined. If you agree, I will replace the return statement with an ASSERT_(.FALSE.).

A similar issue currently exists if the LSM_CHOICE is different from the current two options, i.e. in the last part of the LSM_CHOICE if-statement. I can replace the 'return' used there with an ASSERT_(.FALSE.).

gmao-rreichle: Thanks, @gmao-jkolassa.   Unless @weiyuan-jiang objects, I think we should use ASSERT(.false.) in all three places, that is, for the two "default" statements for "case LAND_PARAMS" and for the "else" statement in the "if LSM_CHOICE" block.

weiyuan-jiang:
Since all checks failed, I will take a look at it.  I will replace the "assert" with new version that takes the extra message as argument. @gmao-rreichle @gmao-jkolassa

gmao-jkolassa:
Hi @weiyuan-jiang ,

I am testing a version with the new 'assert' right now. The old 'assert' caused some issues with CMake despite having MAPL in the CMakeLists.txt.

gmao-jkolassa:
Tested both the old 'ASSERT_' and the newer '_ASSERT' and both fail when compiling, not recognizing the '.FALSE.' as a logical input. I suspect that the dependencies are not built correctly (although MAPL is included in the CMakeLists.txt file). @weiyuan-jiang , would you mind taking a look at this after all?

weiyuan-jiang:
It should be able to compile now. @gmao-jkolassa

gmao-jkolassa:
The latest version of Catchment-CN in this branch has once again been science-verified and was found to reproduce Fanwei Zeng's validated and published LDASsa Catchment-CN simulations to within roundoff errors.

biljanaorescanin:
As we expected 2 tests fail 0diff comparison for GEOSldas related to Catchment-CN; globalcscn and gnuglobalcscn. These tests will be updated to include new information with inclusion of this branch. 
AGCM test passes with 0diff because it doesn't test Catchment-CN. 
cc: @gmao-rreichle , @rdkoster , @weiyuan-jiang  , @gmao-jkolassa

lcandre2 (dismissed):

weiyuan-jiang:
I suggest add "_RETURN(_SUCCESS)"  to the end of the subroutine "SurfParams_init" . And add "rc=status" when it is called. @gmao-jkolassa

gmao-jkolassa:
Implemented suggestions made by @weiyuan-jiang and updated branch.

biljanaorescanin:
Branch re-tested. 
Same result as before minor fix Weiyuan suggested, 2 tests fail 0diff comparison for GEOSldas related to Catchment-CN; tests globalcscn and gnuglobalcscn. These tests will be updated to include new information with inclusion of this branch.
AGCM test passes with 0diff because it doesn't test Catchment-CN.
cc: @gmao-rreichle , @weiyuan-jiang , @gmao-jkolassa

gmao-rreichle (approved):

sdrabenh (approved):

PR #357, opened 2020-10-01
Include new diagnostics for the constraints as well as a new option t…

sdrabenh:
…o just run the dry-mass constraint


mathomp4 (approved):
As usual, I trust @lltakacs and @sdrabenh. I mean it built, so I'm happy!

Issue #41, opened 2019-08-16
multiple versions of CatchCN desired


JustinPerket:
@smahanam and @fwzeng want to develop CatchCN with both CLM.4.0 and CLM4.5 BGC. They currently have functioning code for both. Unfortionally there's heavy duplications of file and subroutine names between their versions.

The quickest route to begin science development would be for them to create in their branch, a couple of the GEOSCatchCN_GridComp directory for 4.5. However, this would be undesirable to push to ldas's branch, or develop.

GEOSCatchCN_GridComp  could have subdirectories clm4.0 and clm4.5. Would it be possible to have either a build option or run-time resource option to select?

smahanam:
Maybe a build option is the way.

CATCHCN uses dynamic vegetation, and Carbon and Nitrogen parameterization schemes from CLM.  CATCHCN's sub-surface hydrology, snow and turbulence schemes are as same as those of  CATCH.  It has its own surface energy scheme. CATCHCN has been using CLM4's DYNVEG, Carbon and Nitrogen parameterization since its implementation. 

We are merely upgrading CLM4 routines to CLM4.5's - the rest remains the same. Once updated, CLM4 routines will be obsolete at some point. We will just need to keep an option for those who want to use "old" CLM4 codes mainly during the transition period. 
In old GNUmakefiles  we could use a directive like $(D)USE_CLM4 and build CLM4 codes if the argument is passed when the model is being compiled.

Creating GEOScatchCN4.5_GridComp is undesirable because I thought GCs are for distinctly different models. We are merely upgrading few parameterizations that we copy from CLM. For e.g. we updated our snow scheme to incorporate impact of dust and black carbon on snow (that came with new prognostic variables as well), but we did not create new GCs for that.
If we add NoahMP, CLM itself, VIC or some other new land surface model, we sure can open a new GC. 
My 2cnts.

tclune:
I'm sure I'm only half-understanding the issues here.  But my 2 cents:

Ideally we could instead just consider the new version the development branch and the old version the master branch.   But this thread would not exist if that option was workable.

If we want to be able to use both versions without complicating the externals configuration (any further than it already is), then some sort of CMake switch to toggle between the two versions seems to be preferable.      This could involve 2 directories or just slightly differently named files or even really large source files with a big `#ifdef` in them.   I defer to the people that have to live with the decision as to which is the most palatable.    I only ask that we aggressively clean up everything once the need to support both versions goes away.

[...]

Issue #232, opened 2020-03-09
Move misplaced mk_restarts directory; clean up duplication of ReadCNTilFile(), ReadTileFile_RealLatLon() etc,  


gmao-rreichle:
Issue #225 and pull request #224 addressed some cleanup of routines in the "mk_restarts" directory, including a renaming of subroutine ReadCNTilFile() to ReadTileFile_RealLatLon() for clarity.
This issue assumes that pull request #224 has been merged.
There are two additional concerns:

1) A few subroutines, incl. to_radian(), haversine() and ReadCNTilFile() a.k.a. ReadTileFile_RealLatLon() 
appear to be duplicated in mk_restarts/getids.F90 and GEOSsurface_GridComp/Shared/Raster/comp_CATCHCN_AlbScale_parameters.F90

2) Subroutine "to_radian()" should probably be eliminated in favor of the corresponding MAPL routine.

3) It is not clear why "mk_restarts" lives in GEOScatch_GridCom/ when it addresses re-tiling for other (non-catch) GridComps.  My feeling is that some reorganization would be needed.

@sdrabenh suggest that @tclune comment on this on behalf of the SI team to see what should be done.

gmao-rreichle:
More duplication:
Subroutine calc_soil_moist() is defined in mk_restarts/Scale_Catch.F90 and also in mk_restarts/Scale_CatchCN.F90, both of which duplicate the functionality of subroutine catch_calc_soilmoist() in GEOSland_GridComp/Shared/lsm_routines.F90
(Note also subroutines partition() and rzequil() in Scale_Catch[CN].F90.)

Re. item 3 above (code reorganization):
E.g., subroutine Scale_CatchCN.F90 is in the GEOS**catch**_GridComp/ subdirectory but should probably be in the Catch**CN** Grid Comp dir.

@gmao-jkolassa @biljanaorescanin @rdkoster @weiyuan-jiang

gmao-jkolassa:
I am happy to do clean up Scale_Catch.F90 and Scale_CatchCN.F90, but I think it makes sense to first move those routines to their final locations, just so that dependencies can be established correctly. 

I had been discussing with Rolf via email about possibly creating a mk_restarts directory in GEOScatchCN_GridComp and I have included his response below:

[...]

PR #368, opened 2020-10-28
Updates to support two versions of Catchment-CN: Catchment-CN.4.0 and Catchment-CN.4.5

gmao-jkolassa:
This PR is in response to issue #41 . The objective is to support two Catchment-CN versions, Catchment-CN.4.0 and Catchment-CN.4.5, and to allow the user to toggle between them.

To support both versions, the GEOScatchCN_GridComp directory has two subdirectories 'CLM40' and 'CLM45'. To toggle between both CatchCN versions it is currently necessary manually change the **CMakeLists.txt** files in the GEOScatchCN_GridComp and GEOSland_GridComp directories to point at the desired CLM subdirectory. If possible, it may make sense to implement this toggle at a higher level and in one place in the future. 

Because CLM4.0 and CLM4.5 are different model versions, there are extensive differences in the CN routines (in the respective CLM subdirectory), which I will not list here.

Restart files differ between both model versions (affects NUM_PFT, NUM_COL, CNPFT, CNCOL, some fixed parameters, and running means).

CLM4.5 has additional diagnostics. Modifications have been made to **GEOS_SurfaceGridComp.F90** and **GEOS_LandGridComp.F90** to support these. 

Additional files affected by this:

**Scale_CatchCN.F90:**
- added ability to handle Catchment-CN.4.5 files
- fixed bug that got in during MAPL2 upgrade and that resulted in array CNPFT being read with incorrect dimensions, leading to multiple carbon variables being set to 0 (this bug affected both CNCLM40 and CNCLM45)

**lsm_routines.F90:**
- added optional surface wetness outputs (swsrf1out, swsrf2out, swsrf4out) to catch_calc_soil_moist subroutine that are used in CLM4.5 simulations

**SurfParams.F90:**
- added new collection 'CN_CLM45' to be used with Catchment-CN.4.5; this parameter set reproduces LDASsa Catchment-CN.4.5 simulations generated by E.Lee for M. Folette-Cook's MAP project.

**GEOS_SurfaceGridComp.rc:**
- added documentation for 'CN_CLM45' parameter set

**NOTE:** The implementation of both Catchment-CN versions will require updates to GEOSldas and GEOSgcm_App repositories and will be linked to pull requests in both. 
GEOSldas PR [326](https://github.com/GEOS-ESM/GEOSldas/pull/326)
GEOSgcm_App PR [177](https://github.com/GEOS-ESM/GEOSgcm_App/pull/177)

**CAVEAT:** In its current version, this branch runs fine in offline mode and has been subject to some science-validation. Additional modifications are required to be able to run Catchment-CN.4.5 with the AGCM. These are currently implemented by @smahanam .

cc: @gmao-rreichle , @biljanaorescanin , @rdkoster , @smahanam , @weiyuan-jiang


mathomp4:
I am going to ping @tclune on this PR as the CMake here is sounds interesting/challenging. 

For your current development, a possible temporary solution is to use CMake `option()`. CMake has the ability to do build-time options, so you could have:
```
cmake .. -DBASEDIR=$BASEDIR/Linux -DCMAKE_INSTALL_PREFIX=../install -DCATCHCN_VERSION=4.0 ...
```
Or whatever you like. That `CATCHCN_VERSION` would then trigger which one you want to build. This could at least make your testing/development right now a bit easier.

However, if you ever anticipate wanting to have both codes built and and available to users, we might have to look at what Tom did for MOM5/MOM6 which was to use shared object libraries that are loaded at runtime. Then you could choose when you run the experiment do you load CN 4.0 or CN 4.5. This requires some careful CMake and maybe some renames here and there.

tclune:
I concur with what @mathomp4 wrote.       This might be better than the MOM5+MOM6 situation where many intermediate interfaces remained the same which trips up the loader.  (execution follows the correct component but then calls procedures in the other version.)

I encourage pursuing the "build both" option.

gmao-rreichle:
I agree with @tclune that "build both" would be best.  Failing that, we would need to implement the suggestion by @mathomp4 that lets the user choose the model version to be built with a simple compile flag.  Manual editing of CMake files by users is not workable.  (How would we implement this in the nightly tests?)
Keep in mind that CatchCN-CLM4.5 is only a stepping stone towards yet another version, CatchCN-CLM5.  Chances are that at any given time more than one versions will be used for science.
PS: I fixed the broken URL links in the comments in this PR and the related GEOSgcm_app and GEOSldas PRs.

gmao-jkolassa:
Thank you all for your feedback on this! I will start implementing a version that builds both models. Given my relative inexperience with CMake, I may need some help with this step.

Another question that I have (probably for Rolf), at what point would we want the user to chose which model they run? Should that be controlled through ldas_setup (and the AGCM equivalent) or is there a better place for this?

smahanam:
I am not sure how build both works. It sounds like 2 separate libraries, then we probably need a separate set of connectivity. etc

From my experience on CATCHCN applications and interactions with people that run CATCHCN, @mathomp4 's suggestion is nice and neat. Ideally, if we could have a default version with a build-time option to override the default version with any other desired version would be great.

Matt, please we would count you to make this change to CMakeLists.txt files. It was you who provided me this version many months ago for us to implement multiple versions of the CN model. Please check in your changes, I will test it at my end.

gmao-rreichle:
> at what point would we want the user to chose which model they run? Should that be controlled through ldas_setup (and the AGCM equivalent)
Can this be done in the rc file?  I suppose LSM_CHOICE and/or LSM_PARAMS could be used to select the specific version of CatchCN.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/SurfParams.F90
```diff
@@ -45,49 +51,82 @@ subroutine SurfParams_init(LAND_PARAMS)
...
+         	RSWILT          = 1500. 
+  
+	case ("CN_CLM45")  ! parameters to reproduce Eunjee Lee's Catchment-CN4.5 fire carbon emission simulations
+                LAND_FIX 	= .TRUE.
+                CSOIL_2  	= 70000. ! Post H5_0  
+                WEMIN    	= 13.
+                AICEV    	= 0.107
+                AICEN    	= 19.893
+                FLWALPHA 	= 0.005
+                FLWALPHA 	= 0.005
```
gmao-rreichle, line 115: remove duplicate entry for FLWALPHA

mathomp4:
All, as @tclune said, that "build both at the same time" for MOM5 and MOM6 had issues because both models defined, say, `ocean_init()` and so run problems were encountered: "Which ocean_init() is called?" or worse which "do_ocean_calc()" is caled... 

Now, one way to fix it would have been to rename all the common bits in one model as `ocean_init_mom5()` or the like. Tedious, but doable. But Tom figured out a clever way to use DSOs and LD_PRELOAD to make it work (though, [as I've learned recently, maybe not quite completely yet](https://github.com/GEOS-ESM/GEOSgcm_App/issues/179)). This was a good choice here because both mom5 and mom6 are external codes that we really wanted to avoid having custom versions of.

But, for Catchment? It's possible that with some changes to gridcomps, etc. this can be done much easier. We have (or had) multiple dynamics, different radiations, etc. All are controlled through AGCM.rc with things like `USE_RRTMG_IRRAD: 1.0` (though that's a bad switch. I was...not good at GEOS in those days).

Now the first suggestion (an `option()` at build time), is pretty simple. The build both takes effort, but it's worth it in the end!

A question for @gmao-jkolassa: To build this new code, is it self-contained in this branch? Or does it need changes in GMAO_Shared, etc.

gmao-jkolassa:
> To build this new code, is it self-contained in this branch? Or does it need changes in GMAO_Shared, etc.

@mathomp4: For building, this code is self-contained in this branch, i.e., there are no changes to GMAO_Shared. All the files that do need to be changed are the ones listed in bold in my original comment. Now, to run an actual experiment, additional changes have to be made to GEOSldas and GEOSgcm_App.

weiyuan-jiang:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/4c39ea2d8094d572b16a9905701f23a2e8c1d787/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/CMakeLists.txt#L5-L8

Are you going to switch clm4.0 and 4.5 by editing the Cmakefile? @gmao-jkolassa . Why not make GEOScatchCN a component and make clm4.0 and  clm5.0 exclusive children?

gmao-jkolassa:
> Are you going to switch clm4.0 and 4.5 by editing the Cmakefile?

@weiyuan-jiang, that is how it is implemented right now, but I think the plan is to move away from having to manually edit the CMakeLists.txt. 

> Why not make GEOScatchCN a component and make clm4.0 and clm5.0 exclusive children?

Could you show me how to do that?

weiyuan-jiang:
@gmao-jkolassa OK, let me think about it before final pull request

mathomp4:
Well, I have a CMake edit that would allow a build-time switch: `-DCATCHCN_VERSION=(4.0|4.5)` that defaults to 4.0. I could commit it (here or on a branch) or I can wait for @weiyuan-jiang to look at the (better) children-of-CN.

weiyuan-jiang:
@gmao-jkolassa @mathomp4 Look at this branch feature/jkolassa_cnclm45_jiang . I realize there are a lot files that have the same names which may create trouble if building simultaneously.  So we can have build-time switch. By default, it is clm40, later on you can change the default by switching

option(CLM4.0 "running clm4.0" OFF)
option(CLM4.5 "running clm4.5" ON)

mathomp4:
> @gmao-jkolassa @mathomp4 Look at this branch feature/jkolassa_cnclm45_jiang . I realize there are a lot files that have the same names which may create trouble if building simultaneously. So we can have build-time switch. By default, it is clm40, later on you can change the default by switching
> 
> option(CLM4.0 "running clm4.0" OFF)
> option(CLM4.5 "running clm4.5" ON)

@weiyuan-jiang I did it a bit differently as I forgot `option()` was only boolean. So I added a variable to be set at build-time:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/compare/feature/jkolassa_cnclm45...feature/mathomp4/selectcn

The `set()` is here:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/b8d662bde265c1e1811b6fcecd4fcd1eff3a4140/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/CMakeLists.txt#L14-L24

I then use that `${CATCHCN_COMPONENT}` in the two `CMakeLists.txt`

weiyuan-jiang:
I think we need to restructure a little bit. But your approach is neat. I will take advantage of that.

weiyuan-jiang:
I adopt your tricks. @mathomp4 .  @gmao-jkolassa , please take a look at my branch and feel free to make your changes.

gmao-jkolassa:
Thank you @weiyuan-jiang ! I will be taking a look at your branch today and test it for Catchment-CN.4.0 and Catchment-CN.4.5. If I understand it correctly, you implemented the build-time switch now. I think it would be good to nevertheless keep thinking about the run-time switch and what that would entail.

weiyuan-jiang:
I have merged your changes to my branch feature/jkolassa_cnclm45_jiang. We can have runtime switch for sure .

gmao-jkolassa:
Ok, thanks! The changes from this morning were just to bring my branch up to date with the latest develop.

gmao-jkolassa:
Successfully tested building both Catchment-CN.4.0 and Catchment-CN.4.5. Merged @weiyuan-jiang 's branch into this one.

gmao-rreichle:
@gmao-jkolassa : 
Issue #232 is the old CatchCN clean-up discussion that I had mentioned yesterday.
Contrary to what I had thought yesterday, #232 does not mention `math_routines.F90` explicitly.
But there are still a couple of suggestions in #232 that are worth looking into as part of this pull request.

@weiyuan-jiang @biljanaorescanin @smahanam

gmao-jkolassa:
@gmao-rreichle : 
Thank you, Rolf! I sent my assessment of the CN files to @weiyuan-jiang and he is currently working on cleaning up the code according to what we discussed yesterday. Maybe this is good time for him to include your suggestions from issue #232 as well.

gmao-rreichle:
@gmao-jkolassa @rdkoster : I just realized that the new directory names and the language on the slides for the code review with Bill do not quite match what I had expected.
Instead of
I thought we would have something like:
./GEOScatchCNCLM40_GridComp
./GEOScatchCNCLM45_GridComp 
This is a bit subtle, but the "40" and "45" numbering goes with "CLM", not "CN", unless we lock in the CLM versioning for CN. 
If we ever have an update to the hydrology without changing the underlying CLM version, we could then use something like
"CN2CLM4x".
Maybe rename the directories accordingly? 



cc: @weiyuan-jiang @biljanaorescanin

gmao-jkolassa:
Rolf,
that is a good point. I think your suggested naming scheme makes sense.

weiyuan-jiang:
Is the LSM_CHOICE redundant in the SurfParams.F90 ? @gmao-jkolassa

gmao-jkolassa:
> Is the LSM_CHOICE redundant in the SurfParams.F90 ? @gmao-jkolassa

It shouldn't be. We still need the LSM_CHOICE check since certain LAND_PARAMS collections should only be used with Catchment, while others should only be used with Catchment-CN. Also, the parameter sets to reproduce the science-validated runs of Catchment-CN4.0 and Catchment-CN4.5 are slightly different. So I would suggest to keep LSM_CHOICE is SurfParams.F90 for now. @weiyuan-jiang

weiyuan-jiang:
When I run clm45,  the debugging error message clearly shows that the compiler links to the wrong subroutine CN_DriverMod.F90 which exists in CNCLM40 and CNCLM45,  i.e.,  GEOS_CatchCNCLM45GridComp links to CLM40's CN_DriverMod.F90. So we need DSO strategy. @gmao-jkolassa @tclune @gmao-rreichle 

forrtl: severe (408): fort: (2): Subscript #3 of the array CNCOL has value 36 which is greater than the upper bound of 35

GEOSldas.x         0000000003F7976F  Unknown               Unknown  Unknown
GEOSldas.x         00000000011E3691  cn_drivermod_mp_c         823  CN_DriverMod.F90
GEOSldas.x         0000000001813770  geos_catchcnclm45        4169  GEOS_CatchCNCLM45GridComp.F90

weiyuan-jiang:
I eventually figure out why it crashes. Basically the linking is a mess and the final GEOSldas.x is not what we want.  If we build cnclm40 and cnclm45 as dynamic shared libraries, they are running fine.  But to do so, we need support from new MAPL

gmao-jkolassa:
This PR is contingent on [GEOSgcm PR 246](https://github.com/GEOS-ESM/GEOSgcm/pull/246), which updates MAPL to v2.5.0.

biljanaorescanin:
@sdrabenh this is ready for review. I did a simple test and it was zero diff for GCM. You will need MAPL 2.5 when running tests. 
+ https://github.com/GEOS-ESM/GEOSgcm_App/pull/177

smahanam:
Unless @weiyuan-jiang or someone has updated the below program (+ 2 perl scripts):
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/develop/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/mk_restarts/mk_CatchCNRestarts.F90

technically, this is not ready. Maybe, I am missing something.

gmao-rreichle:
I'm surprised by @smahanam 's comment that mk_CatchCNRestarts.F90 may not be ready.
Since @gmao-jkolassa  was able to run both CNCLM4.0 and CNCLM4.5, we should at least have some restarts that work.
If that's the case, then we should probably proceed with this set of PRs and sort out any remaining issues with mk_CatchCNRestarts.F90 later. 
Or are we missing something more fundamental?  
cc: @biljanaorescanin @weiyuan-jiang @rdkoster @sdrabenh

gmao-jkolassa:
I am surprised too. I double-checked the restart files that were written for the CNCLM4.5 simulations using this branch and they appear correct, i.e. arrays in the restart files have dimensions corresponding to CNCLM4.5 and not CNCLM4.0. 

I agree with @smahanam that this needs to be addressed, but possibly this could be done independently of this PR as suggested by @gmao-rreichle .

smahanam:
I also don't know, I wasn't following this since I handed over the branch to Jana few months ago. 

Jana, the program that I mentioned and the 2 perl scripts are mk_GEOSldasRestarts.F90/ldas_setup/preprocess_rst.csh equivalent for the AGCM. You and I communicate frequently, I am not sure you ever ran AGCM (Maybe, you did, then I withdraw my comment)

Unless, Weiyuan or someone has updated those AGCM side files (I did not do it when I first created this experimental branch for Melanie's project), this PR breaks AGCM (CATCHCN applications).

gmao-jkolassa:
@smahanam We have only tested this branch for offline simulations so far. Since the GCM currently does not use Catchment-CN, the changes subject to this PR should not have any impact (and I believe @biljanaorescanin has verified this). I believe that a few weeks back we agreed that the AGCM compatibility of CNCLM45 that will be required for Melanie's project would be implemented after we have merged this branch into 'develop'. Maybe I misunderstood?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM40_GridComp/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+esma_set_this ()
+
+string (REPLACE GEOScatchCNCLM40_GridComp_ "" is_openmp ${this})
```
tclune, line 3: This mechanism seems very odd.   All the arguments to the macro are fixed ("this" is just the directory name) so this will either always be true or always be false.     

Perhaps with a fuller explanation of how this is expected to be exercised, I can suggest a different approach.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM45_GridComp/CLM45/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
+esma_set_this ()
+string (REPLACE GEOScatchCN_GridComp_ "" is_openmp ${this})
```
tclune, line 2: E.g.,  we could add a CMake "option" for whether to build this layer with OpenMP.  It can then default to the desired value and be overridden on the CMake command line when desired.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/CMakeLists.txt
```diff
@@ -1,53 +1,15 @@
...
-  CNiniTimeVar.F90
-  CNBalanceCheckMod.F90
-  CN_DriverMod.F90
-  compute_rc.F90
-  catchmentCN.F90
-  catchcn_iau.F90
-  GEOS_CatchCNGridComp.F90 
-  )
+add_subdirectory (GEOScatchCNCLM40_GridComp)
+add_subdirectory (GEOScatchCNCLM45_GridComp)
```
tclune, line 4: I am confused that these subdirectories are created, but the corresponding libraries do not appear to be referenced in this top level CMakeLists.txt.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/CMakeLists.txt
```diff
@@ -1,53 +1,15 @@
...
-if (is_openmp)
-  target_compile_options(${this} PRIVATE ${OpenMP_Fortran_FLAGS})
+  SRCS GEOS_CatchCNGridComp.F90 
+  SUBDIRS Shared 
+  DEPENDENCIES GEOS_SurfaceShared GEOS_LandShared GEOS_CatchCNShared MAPL
+  INCLUDES ${INC_ESMF})
+
+# Special case for GEOScatch and GEOScatchCN components - build with / without openmp
+if (USE_OPENMP)
+  add_subdirectory (GEOScatchCN_GridComp GEOScatchCN_GridComp_openmp)
```
tclune, line 14: Yet another subdirectory with no corresponding link to the top library.

tclune (commented):
I have a few questions on the CMake changes that can probably be readily addressed.  See the 1-line comments interspersed.

gmao-rreichle:
Apologies for the obvious breakdown in communication.  Let me try to summarize today's discussion so far (based on the above and offline emails).  If I understand things correctly, merging the PR at this moment would: 
1) facilitate GEOSldas (land-only) simulations with CNCLM4.0 and CNCLM4.5;  
2) temporarily break the current capability to run the AGCM with CNCLM4.0 (Fanwei's science-tested version);
3) be only a stepping stone towards supporting Melanie's fire work (AGCM with CNCLM4.5).

To address 2. and 3., more work is needed per @smahanam 's comments (CatchCNRestarts.F90 and 2 perl scripts (preprocess_rst.csh and equivalent AGCM script).  Unless a higher priority is assigned to this, @smahanam would not immediately be available to help with this.

The question, mostly for @gmao-jkolassa, @rdkoster and @wmputman, is whether we should move ahead with the PR at this time.  The answer depends on whether there's a need for 1. and whether 2. and 3. are acceptable at this time. 

In any case, the CMake comments by @tclune should be addressed by @gmao-jkolassa and @weiyuan-jiang .

cc: @biljanaorescanin @sdrabenh

gmao-jkolassa:
@gmao-rreichle , to answer your questions: 

Currently, we are actively running offline science experiments using both CNCLM40 and CNCLM45. These are the main motivation for developing this branch that supports both versions of CNCLM. So I would argue that there is a need for 1.

As far as I know, there are no experiments planned in the near future that would use CNCLM40 coupled to the AGCM. But @rdkoster would know better which CNCLM experiments are planned.

For Melanie's project, we are planning to run CNCLM45 coupled to the AGCM. This capability currently does not exist in any branch and would have to be added. If @smahanam is not available to work on this in the near future, would it make sense for @weiyuan-jiang and myself to work on this?

Unless there are some strong arguments against this, my vote would be to go ahead with this PR to have the support for the offline simulations and then aim to implement the CNCLM-AGCM coupling in a separate PR within the next 1-2 months.

rdkoster:
We have no plans to use CNCLM40 coupled to the AGCM, at least not in the near future.

smahanam:
>>>For Melanie's project, we are planning to run CNCLM45 coupled to the AGCM. This capability currently does not exist in any branch and would have to be added. If @smahanam is not available to work on this in the near future, would it make sense for @weiyuan-jiang and myself to work on this?

If Melanie needs to run AGCM while I and Weiyuan are still buried in other stuff,  we sure can help set up CLM45 AGCM experiments for her project without the above updates.

smahanam:
>>>For Melanie's project, we are planning to run CNCLM45 coupled to the AGCM. This capability currently does not exist in any branch and would have to be added. If @smahanam is not available to work on this in the near future, would it make sense for @weiyuan-jiang and myself to work on this?

If Melanie needs to run AGCM while I and Weiyuan are still buried in other stuff,  we sure can help set up CLM45 AGCM experiments for her project without the above updates.

smahanam:
My humble opinion is that this PR should wait until this new feature is available for the main modeling system (AGCM) before we merge it to develop. This slight delay should not have any impact on science research projects because naturally they get the highest priority - as you  know this PR followed Eunjee/Melanie's offline research. We will not wait for this PR if Melanie wants to launch her AGCM experiments.

Either way, I am not going to consider we are setting a precedent for future PRs where the feature is not supported by the main modeling system (AGCM) and all subsystems (GEOSldas).

weiyuan-jiang:
@smahanam I think I have changed the file to make it consistent with the name changes but I will double check. However, I didn't know how to use perl script so I am not sure if they are compatible.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM40_GridComp/CMakeLists.txt
weiyuan-jiang: I actually don't know what happens here. I guess it was leftover when someone debugged and built it with openmp option. Maybe we can just removed it

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/CMakeLists.txt
weiyuan-jiang: We don't reference the libraries because it is DSO. The program will find it during run time

smahanam:
> @smahanam I think I have changed the file to make it consistent with the name changes but I will double check. However, I didn't know how to use perl script so I am not sure if they are compatible.

That wouldn't suffice Weiyuan. For e,g, CLM45 has number of new fields, number of column and pft dimensions are also different. Basically, mk_CatchCNRestarts.F90 does not know anything about CLM45 as it is.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/CMakeLists.txt
weiyuan-jiang: Same reason as above

weiyuan-jiang: but again we may remove the openmp option for this layer

gmao-jkolassa:
@weiyuan-jiang Yes, all the modifications to support offline simulations with CNCLM40 and CNCLM45 were made and are subject to the PR. The modifications to run CNCLM coupled to the AGCM have not yet been implemented. Earlier this month @smahanam and I agreed that we would go ahead with this PR and the @smahanam would implement to AGCM coupling capability working off of the new 'develop'. This will temporarily break the ability to run CNCLM coupled to the AGCM, a combination that no one is currently using. 

If it is possible to implement the AGCM coupling capability quickly, then I am happy to wait with the approval of this PR. However, it sounded like no one would be able to work on this anytime soon. So I am not sure about holding up the dual CNCLM support for the offline simulations until someone is available to work on the AGCM coupling. Just my 2c. If the preference is to wait with approving this PR, that's fine by me too.

smahanam:
>  Earlier this month @smahanam and I agreed that we would go ahead with this PR and 
I am really sorry Jana, this is a misunderstanding - or I actually forget whether we were even talking about a PR. 
If we strictly follow a protocol obviously this branch is not ready. If we want to operate as a one land team, that supports each other's projects, there is no issue at all,  We know we merged a bug, and we come back and fix it later.

> the @smahanam would implement to AGCM coupling capability working off of the new 'develop'. 
There is no issue here, I'll attend to this mid Feb.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM40_GridComp/CMakeLists.txt
weiyuan-jiang: I don't understand why we need this openmp option for this layer. Maybe @aoloso can tell us more on this? If it is useless, I am going to remove it.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/CMakeLists.txt
weiyuan-jiang: The openmp option is from CVS.   Do you know why we need this special openmp option to create different subdirectories? What is the benefit? @gmao-rreichle @smahanam

weiyuan-jiang: opps, should not ask question under this conversation. Let me move to the right place.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM40_GridComp/CMakeLists.txt
weiyuan-jiang: The openmp option is from CVS. Do you know why we need this special openmp option to create different subdirectories? What is the benefit? @gmao-rreichle @smahanam

gmao-rreichle: @weiyuan-jiang :  
At the outset of LDASsa in the early 2000s, the only parallelization option for LDASsa was OpenMP.  With the introduction of MPI, we no longer used OpenMP to run LDASsa.  But the OpenMP elements were left in place and eventually ended up in GEOSldas.  
I doubt OpenMP works for the main executable (GEOSldas.x), and it should be removed where in connection with GEOSldas.x.  
However, there's a chance that some of the boundary condition and/or restart processing still uses OpenMP.  
@smahanam should be able to tell us more.

smahanam: As far as I know, no module under GEOSland_GridComp uses openmp.

1) Restart processing
SUBROUTINE regrid_carbon_vars_omp at L2054 in
mk_CatchCNRestarts.F90 is an obsolete subroutine, The program uses MPI instead.

2) boundary condition
GEOSsurface_GridComp/Shared/Raster
YES two subroutines use and leave it.

Above both are independent executables, not part of GEOSgcm.x or GEOSldas.x

biljanaorescanin:
@gmao-jkolassa @smahanam @weiyuan-jiang @gmao-rreichle @rdkoster, just a quick update from model development meeting, @wmputman said we should keep all on the feature branch until we can actually run experiments with AGCM both CatchCN CLM 4.0 and CatchCN CLM 4.5.
There will be no merge at this time.

gmao-jkolassa:
Thanks for the update!

smahanam:
My apologies Jana again for raising the issue that until the  feature is available for the AGCM, this PR may not be ready.

1) I updated the mk_restart utility to create AGCM restarts for both CLM40 and CLM45. This includes:
a) updates in mk_restart directory of this branch
b) updates in GEOSldas branch
c) created a new branch in GMAO_Shared using the same name - feature/jkolassa_cnclm45

2) There were a few things that had been overlooked in feature/jkolassa_cnclm45 in GEOSgcm_App - fixed that
3)  merged the latest version of develop to this branch, also fixed a bug in GEOS_Surface

That all said, this branch is still not ready for sure. AGCM does not just run with both CLM40 and CLM45. There is a huge memory issue. NCCS kept revoking my node allocation each time I tried to run.

gmao-jkolassa:
Sarith,

Thank you for the update and the changes you have made so far! Please correct me if I misunderstood the earlier conversation in January, but I was under the impression that you were working on the coupling with the AGCM. As far as I know no one else has been assigned to implement this.

What is still needed to add this capability?

smahanam:
> Sarith,
> 
> Thank you for the update and the changes you have made so far! Please correct me if I misunderstood the earlier conversation in January, but I was under the impression that you were working on the coupling with the AGCM. As far as I know no one else has been assigned to implement this.
> 
> What is still needed to add this capability?

When I started working on this, I thought all I had to do is updating mk_restarts utility for the AGCM. After those updates, I started testing, fixed a minor bug and made some updates elsewhere too, and then I got stuck with this memory issue. I actually tried to investigate too (pretty much whole day yesterday), and realized this is beyond my limited skills.

OK then - I will work with Ben and fix this memory issue and what may come afterwards.

I am taking a day off today. That is also one of the reasons to throw the ball back to you. Anyway, this PR is not ready, I will work with Ben and make sure both CLM40 and CLM45 are available for the AGCM.

gmao-jkolassa:
Got it, thanks! Please let me know if there is anything you need me to do. Enjoy your day off!

smahanam:
Hi Jana,

I think this is ready now for you to do some tests yourself.
I'll continue to help, so let me know if you run into any problem.

smahanam:
I made some  modifications yesterday. Please redo whatever you are doing from scratch.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
```diff
@@ -224,15 +224,27 @@ subroutine SetServices ( GC, RC )
...
+       call MAPL_GetResource (SCF,  LAND_PARAMS,   label='LAND_PARAMS:',        DEFAULT="NRv7.2",   __RC__ )
+    elseif (LSM_CHOICE.eq.2) then                                                                                             
+       call MAPL_GetResource (MAPL, LAND_PARAMS,   label='LAND_PARAMS:',        DEFAULT="CN_CLM4",  __RC__ )           
+    elseif (LSM_CHOICE.eq.3) then                                                                                         
+       call MAPL_GetResource (MAPL, LAND_PARAMS,   label='LAND_PARAMS:',        DEFAULT="CN_CLM45", __RC__ )          
+    else
+       _ASSERT(.FALSE.,'unknown LSM_CHOICE')
+    end if
+    call MAPL_GetResource (SCF, CHOOSEMOSFC,   label='CHOOSEMOSFC:',            DEFAULT=1,          __RC__ )
     call ESMF_ConfigDestroy      (SCF, __RC__)
```
gmao-rreichle, line 246: @gmao-jkolassa, @mathomp4,
Why do some of the calls to `MAPL_GetResource()` use `MAPL` as the first argument while others use `SCF`? 
Is there a reason for the difference?  Or is this ripe for cleanup? 
(@gmao-jkolassa, I added an else statement to protect against unknown LSM_CHOICE and cleaned up the indentation and white space.)

smahanam: This is the design:

AGCM calls MAPL_GetResource with MAPL object to read parameters that are sitting in AGCM.rc because parameters in the main resource file have been read in to the MAPL object.
For e.g., LSM_CHOICE, 
The name of the Surface Resource file 'SURFRC:', default = 'GEOS_SurfaceGridComp.rc'
etc.

Then, to read parameters from GEOS_SurfaceGridComp.rc, the model first create the SCF structure and load parameters from GEOS_SurfaceGridComp.rc  to an ESMF object (SCF)
SCF = ESMF_ConfigCreate(rc=status) ; VERIFY_(STATUS)

Then, the model calls MAPL_GetResource  (passes SCF object)  to read parameter values from SCF.

In fact, when I first implemented this couple of years ago, I used call ESMF_GetAttribute to read parameters from SCF. Someone has changed it to call MAPL_GetResource.  After all under the hood, MAPL must be calling ESMF_GetAttribute to read SCF anyway (I am just guessing).

smahanam: I see a problem, 

should use SCF for all parameters that are in GEOS_SurfaceGridComp.rc not MAPL, 
For e.g. MAPL does not know LAND_PARAMS parameter in AGCM.

gmao-rreichle: Many thanks, @smahanam, very helpful!
@gmao-jkolassa:  I fixed the problem for LAND_PARAMS here, but you may want to go through all calls to MAPL_GetResource() in Surface and its children to see if a similar problem exists elsewhere.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
```diff
@@ -224,12 +224,22 @@ subroutine SetServices ( GC, RC )
...
+    ! and Runoff Routing Model (0: OFF, 1: ON)
+    ! -------------------------------------------------------
+    call MAPL_GetResource ( MAPL, LSM_CHOICE, Label="LSM_CHOICE:", DEFAULT=1, RC=STATUS); VERIFY_(STATUS)
     call MAPL_GetResource (MAPL, DO_OBIO, label="USE_OCEANOBIOGEOCHEM:",DEFAULT=0, RC=STATUS); VERIFY_(STATUS)
     call MAPL_GetResource (SCF, ATM_CO2, label='ATM_CO2:', DEFAULT=2, __RC__ ) 
 
     call MAPL_GetResource (SCF, catchswim,     label='N_CONST_LAND4SNWALB:',    DEFAULT=0,        __RC__ )
     call MAPL_GetResource (SCF, landicegoswim, label='N_CONST_LANDICE4SNWALB:', DEFAULT=0,        __RC__ )
-    call MAPL_GetResource (SCF, LAND_PARAMS,   label='LAND_PARAMS:',            DEFAULT="NRv7.2", __RC__ )
+    if (LSM_CHOICE.eq.1) then
```
mathomp4: Hmmm. There seems to be a disconnect here. One is looking in the `SCF` config (`GEOS_SurfaceGridComp.rc`) and two are looking in the `MAPL` config (`AGCM.rc`).

mathomp4: Well, it looks like GitHub showed me an old version this morning? Huh. Time to look at my settings...

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
gmao-jkolassa: Thanks for the explanation @smahanam . And thank you for the code changes @gmao-rreichle . I had not added an else-statement for the unknown LSM_CHOICE, since that is already handled by the setup. But I guess it is good to have multiple protections in case we change the setup.

gmao-jkolassa: And I will check the other MAPL_GetResource() calls in the surface GridComp to make sure they are correct elsewhere.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
gmao-rreichle: @mathomp4, @sdrabenh,
I made one more commit since posting the original comment.  Perhaps that's why @mathomp4 was seeing an older version?  
In any case, with Sarith's help I now understand that rc parameter values are stored in two separate places ("MAPL" and "SCF").  This was easy to fix here.  Upon further reflection, however, I'm wondering if this is really how we want to handle rc parameters.  Keep in mind that this was implemented well before we recently changed the land defaults and started using GEOS_SurfaceGridComp.rc for the AGCM (which had been ignored by the AGCM until then).  I don't recall off the top of my head where the rc parameters from the GEOS_SurfaceGridComp.rc file end up after these recent changes.  I don't think that the preprocessor appends them to AGCM.rc, but perhaps doing that would avoid having to look for rc values in different places?
Perhaps the biggest concern here is that using MAPL_GetResource() elsewhere in the code won't work for the rc parameters in GEOS_SurfaceGridComp.rc unless the whole ESMF_ConfigLoadFile() block is repeated there.
Maybe I'm missing something, but the fact that Jana's original implementation looked for Surface GC rc parameters values in "MAPL" and not in "SCF" suggests that there's ample room for error in future development efforts.
cc: @gmao-jkolassa @smahanam

gmao-jkolassa:
I ran a few more tests and when I use 

call MAPL_GetResource (**SCF**,  LAND_PARAMS,   label='LAND_PARAMS:',            DEFAULT="CN_CLM4",  __ RC __ )

LAND_PARAMS does not get changed to the new default. That is 'NRv7.2' is used even if LSM_CHOICE is 2 or 3. When I use

call MAPL_GetResource (**MAPL**,  LAND_PARAMS,   label='LAND_PARAMS:',            DEFAULT="CN_CLM4",  __ RC __ )

LAND_PARAMS gets correctly set to 'CN_CLM4' (for LSM_CHOICE=2) and 'CN_CLM45' (for LSM_CHOICE=3).

This seems to be counter to Sarith's earlier explanation on how the model should work, so I am not sure what the best way forward is.

gmao-rreichle:
Thanks for the tests, @gmao-jkolassa.   Note that the source code version of GEOS_SurfaceGridComp.rc has everything commented out.   In GEOSldas, a preprocessing script uncomments the lines that pertain to GEOSldas.  I don't remember how exactly @sdrabenh  addressed this for the AGCM when he recently addressed changing the defaults to the new land config.   Scott can probably help.

gmao-jkolassa:
Thanks @gmao-rreichle, that is what I suspected. MAPL_GetResource(SCF...) finds a LAND_PARAMS in GEOS_SurfaceGridComp.rc, so the 'DEFAULT' set in the MAPL_GetResource statement is not used. It may be necessary to have different defaults in the surface resource file for the different LSM choices, but I would probably need some help to implement this.

smahanam:
LSM_CHOICE is the only parameter that the MAPL object knows - so that is the only configurable parameter via AGCM.rc by  passing MAPL  as an argument to MAPL_GetResource ( as it is in the develop version of the code). I just checked the code and verified.

So the SCF (Surface Config File) structure contains everything else (all surface parameters). MAPL object can't help reading them via GEOS_SurfaceGridComp.rc

smahanam:
Sorry I wasn't following your changes:
elseif (LSM_CHOICE.eq.2) then                                                                                             
       call MAPL_GetResource (SCF, LAND_PARAMS,   label='LAND_PARAMS:',        DEFAULT="CN_CLM4",  __RC__ )           
    elseif (LSM_CHOICE.eq.3) then                                                                                         
       call MAPL_GetResource (SCF, LAND_PARAMS,   label='LAND_PARAMS:',        DEFAULT="CN_CLM45", __RC__ )          
    else
    
    is what you need. You used MAPL as an argument, MAPL did not know, the model used the default value.
    
    
    Maybe, I have missed something. It looks to me you are good.

gmao-rreichle:
@gmao-jkolassa, 
I wonder if it's really necessary to specify default values for non-default model choices.  As of now, the default model is Catch, and the default LAND_PARAMS is NRv7.2.  If someone wants to run CatchCN-CLM4 or CatchCN-CLM4.5, they need to change LSM_CHOICE from its default of 1 to either 2 or 3.  When they do that, they can also be expected to choose an appropriate LAND_PARAMS value.  Note that the code in SurfParams.F90 already protects against users forgetting this (the LAND_PARAMS default of NRv7.2 is not a permissible option with LSM_CHOICE=2 or 3.  So I think we're ok without further modifications. 

There are probably ways to implement the appropriate defaults for CatchCN, and I'm not opposed to a smart solution, but I can't think of a simple and effective way just now.  (It's also late on a Friday...)

Having said that, I'm still wondering about the wisdom of not being able to access the SurfaceGC rc parameters in the MAPL object.  But that's a different discussion and not specific to CatchCN-CLM4/4.5.

gmao-jkolassa:
@smahanam Just to clarify, using MAPL_GetResource(SCF...) will always set LAND_PARAMS to 'NRv7.2', which will cause problems in SurfParams_init if LSM_CHOICE is 2 or 3. So in its current for the code does not work for CatchCN.

@gmao-rreichle Currently, the user is not prompted to specify a LAND_PARAMS collection when setting up a GCM run. 'NRv7.2' is used by default, which causes the model to exit as per the check in SurfParams_init if LSM_CHOICE is 2 or 3. So, I think we have three options here: (1) add an LSM_CHOICE dependent default for LAND_PARAMS in the surface rc file, (2) modify gcm_setup to prompt the user for a LAND_PARAMS choice if LSM_CHOICE is 2 or 3, or (3) overwrite the specified LAND_PARAMS value with 'CN_CLM4' or 'CN_CLM45' inside the GEOS_SurfaceGridComp if LSM_CHOICE is 2 or 3. There may be another (better) option that I have not thought of.

smahanam:
I understand what you say Jana.

As of now
In Shared/SurfParams.F90
There is ONLY one parameter set for (LSM_CHOICE==2)  - just the default set of values.
There is ONLY one parameter set for (LSM_CHOICE==3)  - just the default set of values..
I am just wondering why do you need to worry today about an non-existing LAND_PARAM selection for when LSM_CHOICE > 1, in the first place. 

You all know future of CLM40 and CLM45 better than me. Why can't we just delete select case statements inside else if (LSM_CHOICE==2) and else if (LSM_CHOICE==3) in SurfParams.F90  for now,  and deal with this extra option whenever you ever need it in the future (if that ever happens).

Note also that other  than GEOS_SurfaceGridComp.rc, there are 283 rc files to configure various runtime parameters for the model. AGCM.rc file is for the AGCM driver. So I am not sure I understand why SCF params should be part of MAPL.
 
In the past, I have only configured non-land parameters (some chemistry, GOCART params) twice. Both times, I edited the relevant resource file after creating the experiment following Chem folks instructions. 
That is how others do it, that does not mean we should not explore other options. 
Usually, I trust the default settings set up by various groups (GOCART/chemistry folks in this example) when I run the model for our work. So the same way, the land group's job is to get the default setting for the AGCM correct to avoid confusing non-land users. People who want to configure SCF, they actually know what they are doing and if we want to follow others, we can edit GEOS_SurfaceGridComp.rc in EXPDIR just like others do.

This my candid opinion, just take or leave -- 

One example from CO2: My note in AGCM.rc (based on instructions given by CO2 group), see how they change USE_CNNEE default value to 1 for the CO2_GC. 

Again, take or leave... 

# Atmosphere-Land CO2 coupling
# ----------------------------
# USE_CNNEE: 0 #  options : 0 (default); 1 (Chem uses NNEE from CatchmentCN, NOTE you must add USE_CNNEE: 1 to CO2_GridComp.rc, CMIP/CO
2_GridComp.rc,
#                 MERRA2/CO2_GridComp.rc, MERRA2/19600101-20000331/CO2_GridComp.rc, MERRA2/20000401-present/CO2_GridComp.rc, and PIESA/
CO2_GridComp.rc 
#                           in /GEOSchem_GridComp/GOCART_GridComp/CO2_GridComp/ before you build the model.)
#

I am kind of signing off from this particular discussion. I will sure help you any other thing if need be, but these higher level discussions are not my thing.

gmao-rreichle:
@gmao-jkolassa:

> Currently, the user is not prompted to specify a LAND_PARAMS collection when setting up a GCM run. 'NRv7.2' is used by default, which causes the model to exit as per the check in SurfParams_init if LSM_CHOICE is 2 or 3. 

I see.  But as you can tell from my comments, I'm largely ignorant of gcm_setup.  That's really something that needs to be sorted out with @sdrabenh and whoever else is in charge of gcm_setup.

> [...] three options [...] (1) add an LSM_CHOICE dependent default for LAND_PARAMS in the surface rc file, (2) modify gcm_setup to prompt the user for a LAND_PARAMS choice if LSM_CHOICE is 2 or 3, or (3) overwrite the specified LAND_PARAMS value with 'CN_CLM4' or 'CN_CLM45' inside the GEOS_SurfaceGridComp if LSM_CHOICE is 2 or 3.

I'm not sure how (1) would work without some additional annoying keywords and scripting.

I'm not sure the AGCM group would want to prompt for LAND_PARAMS, but perhaps that would be an option.  This prompt could be done immediately after prompting for LSM_CHOICE (assuming there's a prompt for LSM_CHOICE...).
  
In any case, the Surface GridComp rc file needs to be edited by gcm_setup (via script) or by the user (manually).  The latter is what I suggested yesterday.   Let the user select LSM_CHOICE 2 or 3 and at the same time expect the user to know that they need to edit the Surface GridComp rc file accordingly.  This could be communicated to the user during gcm_setup, should they select CatchCN. 

Alternatively, gcm_setup could edit the LAND_PARAMS value in the Surface GridComp rc file.  For now, since there is only one possible value  of LAND_PARAMS for the two CatchCN flavors, gcm_setup wouldn't need to prompt for LAND_PARAMS and could just add the default LAND_PARAMS value into the Surface GridComp rc file.  Things get more interesting if there are ever two different versions of the same CatchCN flavor. 

Viewed more broadly, the problem here is that the current setup only allows for one default land model, which per se seems reasonable to me.   CatchCN is not the default model by a long shot, so I think it's fair to expect CatchCN users to "know what they are doing" and edit the Surface GridComp rc file.  Or, again, work with @sdrabenh to see if gcm_setup can be edited.

smahanam:
If you scan programs in the directory tree  from GOESphysics_GridComp for the word MAPL_GetResource, you will see 649 call MAPL_GetResource lines.

Now take a look at AGCM.rc, and check number of parameters whose default values are actually defined via AGCM.rc (for those that are read from MAPL instance for e.g. DO_OBIO) - you will see just a very few (almost none). 
So it seems, the AGCM's practice is setting default values  as a keyword inn call MAPL_GetResource line, while non-default values are need to know basis for those who have implemented. They change parameters in RC files after setting up the experiment. As I understand from my limited experience, that is how others configure parameters,

When we implemented the GEOS_SurfaceGridComp.rc, we understood ldas_setup works differently and land modelers actively work with different configuration etc., so we implemented different system for GEOSldas that you know how it works.
As part of design we follow below example using'=>" as a key for the script to turn on params.

#GEOSldas= >AEROSOL_DEPOSITION: 0

We also have lines like below
# GEOSagcm=>AEROSOL_DEPOSITION: 0 - in my view at that time, it was merely documenting available configurable params  via GEOS_SurfaceGridComp.rc. More importantly, I was not even thinking to propose gcm_setup to implement what we do in ldas_setup. For gcm_setup  our RC file is just one of 283 RC files, and gcm_setup has been working  its own way for more than a decade.

The only important thing is this: Since  LSM_CHOICE: 3 is configured via AGCM.rc, then the default value of LAND_PARMS for LSM_CHOICE : 3 should be read from the default keyword in call MAPL_GetResource for the AGCM. 
Just leave commented out paramteres in GEOS_SurfacceGridComp.rc as they are for the AGCM. I wouldn't propose to implement what ldas_setup is doing in gcm_setup.

My 2cnts...



In my mind,

tclune:
Note that sometime in the next year, we will start migrating GEOS components away from ESMF config files and over to YAML files.     the `<var>: <value>` syntax works in both, but YAML does not have anything like the `=>` syntax mentioned above.

In general the switch to standardized YAML is expected to make processing the resource files more straightforward as well as to allow other YAML-aware tools (e.g. Python) to work on the same resource files.   This is crucial for adopting workflows without dangerous duplication of configuration values across multiple files.

We expect the transition to be "gentle", allowing gridcomps to use either/both formats for some period of time before enforcing the abandonment of ESMF format.

gmao-jkolassa:
Thank you all for the additional information, this is very helpful! It seems that the consensus is that users who choose CatchmentCN in gcm_setup can be expected to know that a different LAND_PARAMS collection is needed and can be expected to manually edit the GEOS_SurfaceGridComp.rc. for now. I think it would be helpful to make a minor edit to gcm_setup to include a reminder to update LAND_PARAMS if a user chooses LSM_CHOICE 2 or 3. I will work with @sdrabenh to implement this.

smahanam:
Thank you Tom for the heads up !
Jana, there is branch in GEOSgcm_App repo under the same name.

If you want to remind the user, open
https://github.com/GEOS-ESM/GEOSgcm_App/blob/feature/jkolassa_cnclm45/gcm_setup
and just after line 1164
if( $LSM_CHOICE == 3 ) then
echo "IMPORTANT: please set LAND_PARAMS: to whatever in RC/GEOS_SurfaceGridComp.rc file in the experiment directory."
endif

just write your message within "" in the echo statement. And please verify yourself.

gmao-jkolassa:
Thanks, Sarith!

gmao-rreichle:
@tclune, just a quick note of explanation:

> migrating GEOS components away from ESMF config files and over to YAML files. the <var>: <value> syntax works in both, but YAML does not have anything like the => syntax mentioned above.

The => syntax is interpreted by ldas_setup and gcm_setup to pre-process the Surface GC rc source code file into a proper rc file in the run dir.  The resulting run dir *.rc file no longer contains "=>" and can be interpreted by both ESMF config and YAML. 

The approach was developed because we need different default values of the Surface GC rc variables for the GCM App (ie, FP) and the GEOSldas App (ie, SMAP L4).  We also want to have just one Surface GC rc source code file to avoid maintaining two separate files that are otherwise identical in terms of the help text provided with the rc variables.

Please suggest alternative solutions if YAML offers any.

tclune:
Processing such an isolated variant during setup rather than duplicating default resource files is sensible.     We can defer the exact mechanism by which this is managed until the setup scripts themselves are overhauled according to the "plan".   (Basically, we think we can write a generic setup script that applies across DAS, CTM, GCM, etc., but is itself configured by a resource file (will be YAML).   These resource files can indirectly embed some conditionals such as the one you are describing into the processing.    I'm wary of saying more until we have concrete examples and look to see what is the most intuitive/maintainable.

gmao-jkolassa:
I have completed the science tests for this branch. The following experiments were conducted:

GEOSldas_CN40 - offline simulation using Catchment-CN4.0 forced with MERRA-2 meteorology
GEOSldas_CN45 - offline simulation using Catchment-CN4.5 forced with MERRA-2 meteorology
AGCM_CN40 - coupled simulation using Catchment-CN4.0
AGCM_CN45 - coupled simulation using Catchment-CN4.5

Outputs from all four experiments compare well against existing science-validated simulations using Catchment-CN4.0 and Catchment-CN4.5. Differences are within roundoff for the two offline simulations, for which reference simulations exist. Land outputs from the two coupled simulations are scientifically close to existing (offline) simulations.

From my end this PR is good to go.

Please note that in total 4 repositories need to be changed to allow for the support of the two different Catchment-CN versions. This PR is thus contingent on the following PRs

[GEOSldas PR#326](https://github.com/GEOS-ESM/GEOSldas/pull/326)
[GEOSgcm_App PR#177](https://github.com/GEOS-ESM/GEOSgcm_App/pull/177)
[GMAO_Shared PR#203](https://github.com/GEOS-ESM/GMAO_Shared/pull/203)

smahanam:
Great to hear all went well Jana !

When I introduced GEOS_SurfaceGridComp.rc, I used call ESMF_ConfigGetAttribute to read parameters from SCF everywhere.

Now that @tclune mentioned that " abandonment of ESMF" is in the plans and someone has already changed call ESMF_ConfigGetAttribute lines in GEOS_SurfaceGridComp.F90 to call MAPL_GetResource as well , this is just land software team's consideration.

When the SI team start working on ESMF to YAML migration, they will have one less thing to worry if we (land) could change remaining call ESMF_ConfigGetAttribute lines to call MAPL_GetResource  in the codes that the land group manages (note, there are similar lines related to GEOS_SurfaceGridComp.rc in GEOS_LandIce too). 
Note : SCF = ESMF_ConfigCreate(rc=status) and call ESMF_ConfigDestroy   lines   (SCF, __RC__) should stay until the migration

I am not suggesting to include that change in this PR, though.  Maybe, it is also a good time to replace remaining old namelist files in GEOSldas repo.

tclune:
Please be careful with statements like "abandonment of ESMF" in a public setting!  :-)   Just the resource files, not the framework.

(And it is likely that ESMF will extend interfaces to support YAML, in which case it is more like upgrading to a new type of ESMF resource file.)

smahanam:
My apologies Tom ! I misquote your earlier post.

gmao-rreichle:
@gmao-jkolassa, @biljanaorescanin, @sdrabenh, 

I'm afraid we're not quite there yet.  There are a couple of minor issues and one bigger question:

1) `mk_GEOSldasRestarts.F90` was moved from the GEOSldas repo to here by Sarith, but we still have three "mk_restart" programs with overlapping/duplicate elements (`mk_GEOSldasRestarts.F90`, `mk_CatchRestarts.F90`, and `mk_CatchCNRestarts.F90`).  
My assumption was that someone would look at creating a unified "mk_restarts" preprocessor that works for both the AGCM and GEOSldas (where Catch and CatchCN could be in the same program or in different programs).
Unless this cleanup is executed, I don't see the wisdom of moving `mk_GEOSldasRestarts.F90` from one repo to another because this move breaks the change history.
If we merge the PR as is, we will have missed an opportunity to clean up mk_restarts, which would be a shame and makes future maintenance and updates more difficult and error-prone.  I'm just stating this for the record.  I have no energy left to push for this.   

2) `mk_GEOSldasRestarts.F90` is full of hardwired Discover paths in the wrong form.  The paths should point to the official symbolic names of the directories.  Physical hard-drives (as in /gpfsm/dnb42) or links (as in ./rreichle/l_data/) should not be used.   A proper path would look as follows /discover/nobackup/projects/gmao/ssd/...    I talked with @biljanaorescanin about this, who cleaned up the paths as I was writing this comment.

3) The mk_GEOSldasRestarts executable lost its ".x" extension (presumably for consistency), but the "Usage ..." help text and comments.

4) Before the GEOSgcm_GridComp PR can be merged, we need to update and test the GEOSldas PR.  The GEOSldas branch is now out of date with the latest develop.  This can be done by @biljanaorescanin.

cc: @rdkoster @weiyuan-jiang

smahanam:
1) GEOSgcm_GridComp is the birth place of mk_GEOSldasRestarts.F90. Rolf/Weiyuan moved it to GEOSldas repo last year for some reasons. That program is needed for me to prepare CLM45 restarts for the AGCM add other capabilities to regrid.pl. 
So it just came home.  

2) "My assumption was that someone would look at creating a unified "mk_restarts" preprocessor:"
That was my plan as well when I started looking in to it. It needed some thinking, and I put some thought in to this too. 
At the end, I decided to present something that works, supports both AGCM and GEOSLdas, more importantly good enough for the land software experts to understand the workflow of the regridding process that they had missed.  
Since it was I who objected to this PR back in January, I just wanted to show land software experts what they had missed, what needs to be done to make it work.

3) Both Jana and I myself tested GEOSldas repo after moving mk_GEOSldasRestarts here, so that part has been verified.

Unified F90 code: good idea indeed, now that I have presented a system that works the way it should, please go ahead and upgrade. ScakeCatch.F90 and ScaleCatchCN.F90 can also be unified to a single program easily.

Alternatively, if you want, merge  to develop this working version that supports 2 CLM versions, and is available for both AGCM/GEOSldas systems for now - since it has been tested. 
If you couldn't figure out yourself, I sure can do it later this year or in 2022. I many high priority projects on my plate now, sorry I am not available now.

weiyuan-jiang:
@smahanam , when you move mk_GEOSldasRestarts from GEOSldas to here, I think the restart names  "catchcnclm40_internal_restart" or "catchcnclm45_internal_restart" become "catchcn_internal_restart".  Why do you change it? If there is no specific reason, I want to change it back.  ( I will do it)

smahanam:
Glad you noticed @weiyuan-jiang 

When I started looking in to this, I realized the whole thing needed too many updates to bring it up to a working version that supports CLSM, CLM40 and CLM45 models in both GCM and GEOSldas. So I made that change to keep software/scripts simple and reduce the number of logics that we will need to add in order to accommodate 2 different file labels to get to a working version.

As it is, the model (I mean GEOSgcm.x or GEOSldas.x) uses 'catchcn' as the label for both catchcn40 and catchcn45 checkpoint/restart. By design, a model will always uses only one version for a simulation (CLM40, CLM45, CLM50) so using the same label should never confuse the model - i.e. no experiment is going to write both catchcn40_internal and catchcn45_internal during a single simulation. 

Those who run the model know very well how to distinguish the version by looking at the netcdf header and they also know they are not interchangeable. Also, our scripts have the necessary checks before reading/writing an _internal_rst.

Anyway, this PR presents a complete  package of updates. Depending on long term plans of CLM40 and CLM45, if you want, please go ahead and update. Perhaps, the practical thing at this point  is let this working version see develop (and main repos) first, and address this label and other issues in a future PR.

gmao-jkolassa:
Following the latest changes to this branch, I re-ran the science validation for this branch (as well as the associated branches in other repositories). I ran four science experiments in total.

**CN_40:** Offline simulation using Catchment-CN4.0 forced with MERRA-2 forcing data. Outputs from this run were within roundoff differences compared to the science-validated Catchment-CN4.0 simulation generated by Fanwei Zeng using LDASsa.

**CN_45:** Offline simulation using Catchment-CN4.5 forced with MERRA-2 forcing data. Outputs from this run were zero-diff compared to a Catchment-CN4.5 simulation previously generated using GEOSldas.

**AGCM_CN40:** Simulation using Catchment-CN4.0 coupled to the GCM. Outputs from this run were zero-diff compared to a simulation generated using the 'main' branch of GEOSgcm and selecting Catchment-CN4.0 as the land option. 

**AGCM_CN45:** Simulation using Catchment-CN4.5 coupled to the GCM. No true reference currently exists for this simulation. Instead I evaluated the surface states and fluxes produced in this simulation against those produced in an offline Catchment-CN4.5 simulation (forced with MERRA-2). Differences were small and on the order that can be expected when comparing an offline and coupled experiment. 

From my end, we are good to proceed with this PR.

gmao-rreichle:
Thanks, @gmao-jkolassa.  A quick question:

>CN_40: Offline simulation using Catchment-CN4.0 forced with MERRA-2 forcing data. Outputs from this run were within roundoff differences compared to the science-validated Catchment-CN4.0 simulation generated by Fanwei Zeng using LDASsa.

Just to confirm: Is this offline run zero-diff w.r.t. a simulation using the current GEOSldas `develop` branch?

gmao-jkolassa:
> Just to confirm: Is this offline run zero-diff w.r.t. a simulation using the current GEOSldas develop branch?

Yes, it is.

gmao-rreichle:
@gmao-jkolassa, regarding the most recent commit [85b3070](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/368/commits/85b3070b9acab9f7a3c2c8fcbcec1fca588e9fef):
I take this change to mean that PRESCRIBE_DVG will not work for CatchCNCLM45.
The resource variable is still around for CatchCNCLM40, although I'm not sure it ever worked.  
In any case, `GEOS_SurfaceGridComp.rc` probably needs some editing to point out the difference between CatchCNCLM40 and CatchCNCLM45 in this respect.
Are there other differences between CatchCNCLM40 and CatchCNCLM45 that need to be mentioned in `GEOS_SurfaceGridComp.rc`?
cc: @biljanaorescanin

gmao-jkolassa:
@gmao-rreichle Yes, I was changing the `ESMF_ConfigGetAttribute` statements to `MAPL_GetResource` statements yesterday (this update had not made it into the CN GridComps because of the changed file names) and I accidentally added in PRESCRIBE_DVG, which has not previously been part of the CN_CLM4.5 GridComp. The latest commit was to correct that. It is unclear to me whether the initial decision to not have the PRESCRIBE_DVG option for CN_CLM4.5 was intentional or a mistake. So I can either add it in properly for CN_CLM45 or keep it in CN_CLM40 only and update the Surface GridComp resource file as you suggested.

biljanaorescanin:
@sdrabenh this PR is ready for review. No more updates will be pushed to this branch (except in case any updates to develop happen prior to merging this PR). 

I've tested:
1) Develop when choosing catchment option vs this branch: it is zero diff
2) Develop when choosing catchment-CN option vs this branch option catchment-CN with CLM 4.0: It is zero diff. 
3) This branch has a new option where one can chose catchment-CN with CLM 4.5. We don't have a reference AGCM run to compare it to. @gmao-jkolassa performed comparison of the branch and offline experiments to confirm only small differences (expected to be seen between the offline and coupled runs). The results were approved by @gmao-rreichle and @rdkoster .

The implementation of both Catchment-CN versions requires approval of the following linked PRs: GMAO_Shared PR [203](https://github.com/GEOS-ESM/GMAO_Shared/pull/203), GEOSgcm_App PR [177](https://github.com/GEOS-ESM/GEOSgcm_App/pull/177). Once these are approved and merged, we will merge the updates in GEOSldas PR [326.](https://github.com/GEOS-ESM/GEOSldas/pull/326) .
I removed DNA label from all linked PRs.

sdrabenh (approved):

mathomp4 (approved):
Not the biggest fan of the CMake in here, but approving.

gmao-rreichle:
@mathomp4: If you have any specific concerns or improvements re. CMake, perhaps write up an issue so this can be addressed in a separate, follow-up PR.  Thanks for going with the flow and letting this PR advance.  It's been around for a while and is complex enough as it were.  Merging it now is the sensible thing to do.  It may not be perfect, but it does what it needs to do for science and has been tested successfully.

PR #373, opened 2020-11-03
runtime choice for clm4.0 and clm4.5

weiyuan-jiang:
I made changes so this branch can choose cn4.0 and cn4.5 during run time.  Here are the outstanding issues:

1) In LDAS.rc, we can add CATCHCN_VERSION:  CLM40  or CATCHCN_VERSION:  CLM45 to choose 4.0 or 4.5 before running GEOSldas.  However, we can overload LSM_CHOICE too. That probably can make setup neat. 

2) Now GEOS_CatchCNGridComp.F90 becomes a proxy of GEOS_CatchCN40GridComp.F90 ( CLM40)  and GEOS_CatchCN45GridComp.F90 (CLM45).  We need to verify that its import variables are the union of CLM40 and CLM45 , and its export variables are the intersection of CLM40 and CLM45. So far I am using CLM40's import and export.

@gmao-jkolassa


tclune (approved):
CMake changes ok.

gmao-jkolassa:
@weiyuan-jiang Thank you very much for preparing this branch! I took a look and had a few comments.

> We need to verify that its import variables are the union of CLM40 and CLM45

I looked through the MAPL_AddImportSpec statements and it seems that these already correspond to the union of CLM40 and CLM45. Both models differ in the variables read through MAPL_AddInternalSpec statements, but it seems that these are handled in the respective CN_GridComps anyway.

> and its export variables are the intersection of CLM40 and CLM45. So far I am using CLM40's import and export.

If I interpret this correctly that only variables that CLM40 and CLM45 have in common should be exported, then some changes were necessary. I have implemented and pushed these. Please let me know if I did not interpret this correctly.

I also tried to run a 1-day experiment with this branch, so far unsuccessfully. The build and ldas_setup run fine, but the run itself fails with the following error message 'ERROR: cannot find export TC', which appears to occur when the resource parameters are being read. This error message is the same irrespective of whether I specify a CATCHCN_VERSION (in LDAS.rc) or not. I also get the same error irrespective of whether I try to run this with GEOSldas branch 'develop' or 'feature/jkolassa_cnclm45' (the latter has changes needed to run CLM45). Or maybe my testing of this is premature and there were more changes you were going to make?

weiyuan-jiang:
The error means we need to add "TC" to the export in the file GEOScatchCN_GridComp.F90.  The variable "TC" is an internal variable in CLM40 and with "FRIENDLYTO" specification.  That means it is exportable. In the parent  GEOSland_GridComp, indeed it is exported. So  we need to add all those with "FRIENDLYTO" to the export list as long as they are in GEOSland_GridComp. @gmao-jkolassa

weiyuan-jiang:
Now I believe it can be run. After ldas_setup and before the run, you need to 
1) change LDAS.rc,  LSM_CHOICE: 2 is CN4.0, LSM_CHOICE: 3 is cn4.5
2) rename catchcn_internal_rst in input/restart to catchcn40_internal_rst or catchcn45_internal_rst
3) need to compare the results with the compile-time version. The system may mis-load the libraries. If that is the case, I will have to make other changes (ie. dynamic libraries).
@gmao-jkolassa

gmao-jkolassa:
Thank you @weiyuan-jiang ! I will test this and report back on how the compile-time and run-time versions compare.

gmao-jkolassa:
@weiyuan-jiang Just a quick update. I tried running a short experiment with the updated branch and there were a few issues. 

1. There was a  'MAPL_AddExportSpec' for CNFROOTC in the GEOS_LandGridComp.F90 and in GEOS_CatchCNGridComp.F90. Based on what you said earlier that the only the intersection of CLM40 and CLM45 variables should be exported I removed the AddExport statements, as this variables is unique to CLM45. Doing this allowed the run to get further.

2. There seems to be a problem with the call to MAPL_GenericInitialize in GEOS_LdasGridComp.F90. I _think_ that what is needed is a separate initialization for each child, but I am not sure. Would you mind taking a look? The GEOSldas branch to use is feature/jkolassa_cnclm45.

weiyuan-jiang:
1) that's great. I basically copied the land exports 
2) I relayed the land initialization to CN. Let me know if there are still problems. I am surprised that there is no Initialize method in catchCN  gridComp. Maybe we should consult this with @smahanam 
@gmao-jkolassa

PR #390, opened 2020-12-09
Changes needed to allow use of aggressive flags with GEOS

mathomp4:
Activate by using `-DCMAKE_BUILD_TYPE=Aggressive`


mathomp4:
I tried this without the changes to ESMA_cmake and ecbuild and `Release` was zero-diff.

tclune (approved):

PR #392, opened 2020-12-14
Add partition option for regrid

mathomp4:
This adds a `-partition` option to `mk_restarts` so that there can be a `-partition` option in `regrid.pl`.

Thus, https://github.com/GEOS-ESM/GMAO_Shared/pull/146 is dependent on this PR.


mathomp4:
Note: I'll test these together before making a PR. I might also have @smahanam stare at this one to make sure I didn't screw anything up. But if it works, I probably didn't.

smahanam:
Hi @mathomp4 ,

Your modifications look OK to me. As you know, @gmao-jstassi is the original author of those perl scripts.

mathomp4:
> Hi @mathomp4 ,
> 
> Your modifications look OK to me. As you know, @gmao-jstassi is the original author of those perl scripts.

@smahanam True. I just get wary touching things in surface. I still need to test it out a bit. Good thing is it's a hard fail if I get it wrong since SLURM will get all confused :)

mathomp4:
Okay I have tested this in concert with https://github.com/GEOS-ESM/GMAO_Shared/pull/146 and it seems to work and I can pass in a partition. I believe is this safe to take at any time since the users of `regrid.pl` can't even pass in a partition until this is in. But as `-partition` is not required, this works with older `regrid.pl`.

mathomp4:
Note for @smahanam and @gmao-jstassi, I added `-W` to the `sbatch` command in `mk_Restarts` because in testing I was getting race condition where it was trying to move files before they'd been generated. I think the `sbatch` was returning immediately when the rest of the script thought it was done.

gmao-rreichle (approved):
Looks ok to me, thanks.  Approving for the land group (code owners).

bena-nasa (approved):

sdrabenh (approved):

Issue #395, opened 2021-01-05
Array subscript out of bounds


yvikhlya:
Following #394. After fixing the bug with TA pointer, the same set up crashes in OGCM with the error:

```
forrtl: severe (408): fort: (3): Subscript #1 of the array GCS has value 0 which is less than the lower bound of 1

Image              PC                Routine            Line        Source
GEOSgcm.x          0000000012317E5F  Unknown               Unknown  Unknown
GEOSgcm.x          000000000ED93E5A  geos_ogcmgridcomp        2242  GEOS_OgcmGridComp.F90
```

Corresponding code:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/eb7e1994a5169a285b10a93df3cecaad632fa7c5/GEOSogcm_GridComp/GEOS_OgcmGridComp.F90#L2242

This is the issue only in dual ocean mode, crashes debug build.

atrayano:
@yvikhlya could you point me to your experiment setup. I need to check few values in the resource file

yvikhlya:
@atrayano 
```
~yvikhlia/geos5/v10170_do
~yvikhlia/aogcm/v10170_do
```

atrayano:
@yvikhlya thanks! I'll let you know what I find. From very quick code inspection, my top suspect is that the variable OBIO has not been changed and most likely still is zero.

yvikhlya:
@atrayano It looks like OBIO is assigned only if(DO_OBIO /= 0)

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/eb7e1994a5169a285b10a93df3cecaad632fa7c5/GEOSogcm_GridComp/GEOS_OgcmGridComp.F90#L233

In S2Sv3 tag else statement also has OBIO=MAPL_AddChild(...) Not sure where it was lost.

yvikhlya:
@atrayano Ah! ObioSimpleGridComp was deleted and is not added as child if(DO_OBIO = 0)

atrayano:
@yvikhlya I totally agree, but I do not think anything would trigger DO_OBIO to test "true". And the way I have coded the explicit calling of the children, OBIO is one of them. I can put put a simple protection in the loop around line 2242 to skip the iteration if a child Id is 0. This definitely will "fix"the error, but I want to think a little more if we are masking a bigger issue this way. I will get back to you by tomorrow

atrayano:
[...]

PR #396, opened 2021-01-06
Fixes #394 and #395 (ta and  obio dual ocean)

### LINKED ISSUE #394: Bug in GEOS_SurfaceGridComp.F90: TA used in Run2 without MAPL_GetPointer
It looks like @yvikhlya found a bug in GEOS_SurfaceGridComp.F90. Namely, there is this code in the Run2 procedure:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/eb7e1994a5169a285b10a93df3cecaad632fa7c5/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90#L5767-L5820

and specifically:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/eb7e1994a5169a285b10a93df3cecaad632fa7c5/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90#L5808-L5819

The fun part is, no where in Run2 is TA "gotten" as it were by `MAPL_GetPointer`. It's declared:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/eb7e1994a5169a285b10a93df3cecaad632fa7c5/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90#L4896

and then the next mention of it is line 5808.

The change came in with https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/4e3110e097500a64bac0069cc4ba82919c8fe04d with the commit message: "Merged with S2S-3 cvs tag". So I'm wondering if that CVS tag *does* have a `MAPL_GetPointer` that was never transferred to Git? Can @yvikhlya or @sanAkel take a look at that?

Also: I guess I'll ping/tag @gmao-rreichle as his name is around this code. And maybe @atrayano because, well, I get confused when we get near tiles.

atrayano:
This version fixes two issues: #394 and #395. The first of them was fixes by @yvikhlya in this branch:
 https://github.com/GEOS-ESM/GEOSgcm_GridComp/tree/bugfix/yury/dual_ocean
and my fix for #395 went on top


yvikhlya:
Let me test this real quick.

GEOSogcm_GridComp/GEOS_OgcmGridComp.F90
```diff
@@ -235,6 +235,7 @@ subroutine SetServices ( GC, RC )
        ORAD = MAPL_AddChild(GC, NAME='ORAD', SS=OradBioSetServices, RC=STATUS)
        VERIFY_(STATUS)
     else
+       OBIO = 0
```
sanAkel, line 238: @atrayano a comment would help, which would make it clear that (if I understand correctly!) there is no OBIO child

atrayano:
This would be great! Thank you, Yuri

My “testing” was to make sure the code compiled 😊

GEOSogcm_GridComp/GEOS_OgcmGridComp.F90
```diff
@@ -2234,6 +2236,7 @@ subroutine RUN ( GC, IMPORT, EXPORT, CLOCK, RC )
           CHLD = (/OBIO,ORAD,SEAICEd,OCEAN/)
           DO N=1, size(CHLD)
              ID = CHLD(N)
+             if (ID <= 0) cycle
```
sanAkel, line 2239: Same comment as above, please add a single description. Thank you!

atrayano: @sanAkel I agree, but my main motivation was to give @yvikhlya something quick to test.

yvikhlya:
Both these fixes seem to work, but debug build still crashes with floating point overflow error and no detailed traceback for some other reason. Regular build runs, but according to print out, memory usage grows all the time.

sanAkel:
> Both these fixes seem to work, but debug build still crashes with floating point overflow error and no detailed traceback for some other reason. Regular build runs, but according to print out, memory usage grows all the time.

@yvikhlya valgrid?

sanAkel:
> Memory leak issue is solved on a different branch. Debug build crash should probably be another issue too. I think we should merge these updates into develop, because these are necessary bug fixes.

I am opposed to merging code which is known to be buggy and crashes. Particularly if the reason(s) for crash are not know. cc: @tclune, @mathomp4, @sdrabenh, @wmputman

yvikhlya:
Memory leak issue is solved on a different branch. Debug build crash should probably be another issue too. I think we should merge these updates into develop, because these are necessary bug fixes.


> > Memory leak issue is solved on a different branch. Debug build crash should probably be another issue too. I think we should merge these updates into develop, because these are necessary bug fixes.
> 
> I am opposed to merging code which is known to be buggy and crashes. Particularly if the reason(s) for crash are not know. cc: @tclune, @mathomp4, @sdrabenh, @wmputman

Proposed fixes are correct and are **necessary** to run the model in dual ocean mode. The crash which we experience in debug build is a different issue and I would suggest to resolve it in different PR.

sanAkel:
> > > Memory leak issue is solved on a different branch. Debug build crash should probably be another issue too. I think we should merge these updates into develop, because these are necessary bug fixes.
> > 
> > 
> > I am opposed to merging code which is known to be buggy and crashes. Particularly if the reason(s) for crash are not know. cc: @tclune, @mathomp4, @sdrabenh, @wmputman
> 
> Proposed fixes are correct and are **necessary** to run the model in dual ocean mode. The crash which we experience in debug build is a different issue and I would suggest to resolve it in different PR.

Is that
>The crash which we experience in debug build is a different issue

**different issue** noted some where, number?

Issue #400, opened 2021-01-19
NOBM connectivity check


sanAkel:
# Check 
If following variables are being passed over to the children (OBIO and ORAD):

FROCEAN
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/e0cf9fabbe700c3e23571466f444973dd6b20e4e/GEOSogcm_GridComp/GEOS_OceanBioGeoChemGridComp/GEOS_OceanbiogeochemGridComp.F90#L317-L325

MASKO
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/e0cf9fabbe700c3e23571466f444973dd6b20e4e/GEOSogcm_GridComp/GEOS_OceanBioGeoChemGridComp/GEOS_OceanbiogeochemGridComp.F90#L327-L335

remove USE_ODAS
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/e0cf9fabbe700c3e23571466f444973dd6b20e4e/GEOSogcm_GridComp/GEOS_OceanBioGeoChemGridComp/GEOS_OceanbiogeochemGridComp.F90#L445-L473

# need exports from OBIO and ORAD, preferably on lat-lon grid:
- PCO2
- FCO2
- BIO
- PPDIATOM
- ...
- PPPHAEO
- DISCHARGE is already there
**Is everything ready for them to be exported??**

# Following imports and Exports to support GEOS_OradBioGridComp.F90 are needed:

COSZ
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/e0cf9fabbe700c3e23571466f444973dd6b20e4e/GEOSogcm_GridComp/GEOS_OradBioGridComp/GEOS_OradBioGridComp.F90#L134-L141

FROCEAN
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/e0cf9fabbe700c3e23571466f444973dd6b20e4e/GEOSogcm_GridComp/GEOS_OradBioGridComp/GEOS_OradBioGridComp.F90#L179-L186

# need exports, preferably on lat-lon grid:

- SWHEAT
- KPAR
- TIRRQ
- CDOMABSQ

Variables such as: 
MODIS Aerosols (Atmospheric Optics) that _Data Atmosphere_ cannot get from _real_ atmosphere should be **available** from data file(s) and Data Atmosphere has to provide them.

sanAkel:
Following seems to be an **error**!

[...]

PR #382, opened 2020-11-24
Change Intel Debug flags for MOM6

mathomp4:
This PR changes the Intel Debug flags for MOM6 (and FMS). The issue seems to be that `-check all,noarg_temp_created` is "too much" for MOM6 (and FMS). So, this change "degrades" the Intel flags to be `-check bounds,uninit` which matches the Intel Debug flags GEOS used to use before [ESMA_cmake v.3.14](https://github.com/GEOS-ESM/ESMA_cmake/tree/v3.1.4).

Note this has a companion PR in FMS as both need to degrade: https://github.com/GEOS-ESM/FMS/pull/16

PR #401, opened 2021-01-20
Revive OceanBioGeoChem

gmao-esherman:
If "USE_OCEANOBIOGEOCHEM:1", then the SEASALT gridded component will fail looking for exports from OBIO, which do not exist. A condition has been put in place that checks the name of SEASALT's children for "OBIO" and only retrieves the exports/pointers if the child is not OBIO.


yvikhlya:
I do not understand how this is all related to obio? Is this a problem because obio is meant to run with data atmosphere instead of agcm?

gmao-esherman:
> I do not understand how this is all related to obio? Is this a problem because obio is meant to run with data atmosphere instead of agcm?

No, I don't think this is an issue with Obio running with data atmosphere.

There is a gridded component called Obio (GEOS_ObioGridComp.F90) that is a child of Seawater. When "USE_OCEANBIOGOECHEM:1" it enables this Obio component and it is "born". It is confusing because there is another Obio that is a child of OGCM. The Obio that is the child of the OGCM is from GEOS_OceanbiogeochemGridComp.F90.

GEOS_SaltWaterGridComp.F90 loops over size(GCS) (where GCS are the children's gridded components) and retrieves pointers. If Obio is "born", Saltwater tries to retrieve exports/pointers that Obio does not have and the model crashes.

sanAkel:
> > I do not understand how this is all related to obio? Is this a problem because obio is meant to run with data atmosphere instead of agcm?
> 
> No, I don't think this is an issue with Obio running with data atmosphere.
> 
> There is a gridded component called Obio (GEOS_ObioGridComp.F90) that is a child of Seawater. When "USE_OCEANBIOGOECHEM:1" it enables this Obio component and it is "born". It is confusing because there is another Obio that is a child of OGCM. The Obio that is the child of the OGCM is from GEOS_OceanbiogeochemGridComp.F90.
> 
> GEOS_SaltWaterGridComp.F90 loops over size(GCS) (where GCS are the children's gridded components) and retrieves pointers. If Obio is "born", Saltwater tries to retrieve exports/pointers that Obio does not have and the model crashes.

@gmao-esherman rather than _born_ may be we should say, _instantiated_?

sanAkel:
@gmao-esherman as you said okay to it, I will push a few changes to the surface GC. Once all together, we just have to do a few quick checks: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/382#issuecomment-733187697 Since you already have the `default` run with `USE_OCEANOBIOGEOCHEM:0` I suggest please keep it _alive_! Thanks! It will come handy for zero-diff testing. We can skip testing with one of the ocean models. From your exp dir: `/gpfsm/dnb32/esherman/coupled_obio` I see you tested with `MOM5` ocean model, so in the check list we can skip `MOM6`.

gmao-esherman:
> > > I do not understand how this is all related to obio? Is this a problem because obio is meant to run with data atmosphere instead of agcm?
> > 
> > 
> > No, I don't think this is an issue with Obio running with data atmosphere.
> > There is a gridded component called Obio (GEOS_ObioGridComp.F90) that is a child of Seawater. When "USE_OCEANBIOGOECHEM:1" it enables this Obio component and it is "born". It is confusing because there is another Obio that is a child of OGCM. The Obio that is the child of the OGCM is from GEOS_OceanbiogeochemGridComp.F90.
> > GEOS_SaltWaterGridComp.F90 loops over size(GCS) (where GCS are the children's gridded components) and retrieves pointers. If Obio is "born", Saltwater tries to retrieve exports/pointers that Obio does not have and the model crashes.
> 
> @gmao-esherman rather than _born_ may be we should say, _instantiated_?

Yes! Instantiate is a much clearer word to use.

sanAkel:
@gmao-esherman since this feature of `USE_OCEANOBIOGEOCHEM:1` does not **yet** work with the code changes (as of 01/21) would you please draft this PR, so @sdrabenh  gets notified. Thanks!

gmao-esherman:
> @gmao-esherman since this feature of `USE_OCEANOBIOGEOCHEM:1` does not **yet** work with the code changes (as of 01/21) would you please draft this PR, so @sdrabenh gets notified. Thanks!

Agreed, done!

yvikhlya:
Normally, we submit a PR when a code is actually working, passes all tests and ready to be merged with develop. We don't want to have a disfunctional code in develop, right? Until then work can be done on a long live feature branch. I would suggest to make a branch something like feature/obio and accumulate all bug fixes there until this feature actually works, then make a PR.

gmao-esherman:
> Normally, we submit a PR when a code is actually working, passes all tests and ready to be merged with develop. We don't want to have a disfunctional code in develop, right? Until then work can be done on a long live feature branch. I would suggest to make a branch something like feature/obio and accumulate all bug fixes there until this feature actually works, then make a PR.

I agree. There was some initial confusion on my part of what this PR was supposed to be. I have since taken @sanAkel's suggestion and made this PR a draft.

GEOSogcm_GridComp/GEOS_OradBioGridComp/clrtrans.F90
```diff
@@ -199,13 +202,21 @@ subroutine sfcrfl(rad,lam,aw,bw,wfac,theta,wspd,rod,ros)
        ifst = 1
       endif
 
+if(wspd > 1000.0) then
+ wspd = 0.0
```
sanAkel, line 206: @gmao-esherman I know nothing about this code part of the model, but, `wspd` values of > 1000.? Something is wrong somewhere **else**! 

Simply resetting to `0.0` is bypassing the problem!

gmao-esherman: Yes I agree. This is NOT a permanent solution. The issue is with the 'UU' variable (where the pointer is WSD) in GEOS_OradBioGridComp. WSD has at least one value of 1.0E14 that causes the model to crash in the clrtrans subroutine. We can filter out the bad value(s) in the line of code below. It's just a question of how.

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/31d9619259ce92f0a6b72a6906e81e4a38e1351b/GEOSogcm_GridComp/GEOS_OradBioGridComp/GEOS_OradBioGridComp.F90#L939

Maybe something like...
`if (WSM(i,j) == 1.0E14) then cycle`
`else`
`wspd = WSM(i,j)`
`end if`

Whatever fix we apply, we should also apply to GEOS_OceanbiogeochemGridComp.F90. I am not sure why the kloop didn't crash with the same bad wind speed.
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/31d9619259ce92f0a6b72a6906e81e4a38e1351b/GEOSogcm_GridComp/GEOS_OceanBioGeoChemGridComp/GEOS_OceanbiogeochemGridComp.F90#L1324

yvikhlya:
> WSD has at least one value of 1.0E14 that causes the model to crash in the clrtrans subroutine.

This looks like a wrong land-sea mask.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
```diff
@@ -3957,6 +3957,12 @@ subroutine RUN1 ( GC, IMPORT, EXPORT, CLOCK, RC )
     call MAPL_GetPointer(IMPORT  , DZ    , 'DZ'    ,  RC=STATUS); VERIFY_(STATUS)
     call MAPL_GetPointer(IMPORT  , UU    , 'SPEED' ,  RC=STATUS); VERIFY_(STATUS)
 
+do i = 1, ubound(UU,1)
```
sanAkel, line 3960: these should come off. Good for debugging, but not leave stuff like this!

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
```diff
@@ -3957,6 +3957,12 @@ subroutine RUN1 ( GC, IMPORT, EXPORT, CLOCK, RC )
...
+  do k = 1, ubound(UU,2)
+    if (UU(i,k) > 1000.0) print*,'SURFACE UU = ',UU(i,k)
+  end do
+end do
```
sanAkel, line 3964: same as above, please take them off.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSsaltwater_GridComp/GEOS_SaltWaterGridComp.F90
```diff
@@ -1987,6 +1990,11 @@ subroutine SALTWATERCORE(NT,RC)
     if(associated(USTR3)) USTR3 = sqrt(sqrt(TXO*TXO+TYO*TYO)/water_RHO('salt_water'))**3
     if(associated(UUEX))  UUEX  = UU
 
+if(mapl_am_i_root()) print*,'SALTWATER size(UU) = ',size(UU)
```
sanAkel, line 1993: same as above

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSsaltwater_GridComp/GEOS_SaltWaterGridComp.F90
```diff
@@ -1987,6 +1990,11 @@ subroutine SALTWATERCORE(NT,RC)
...
+do i = 1, ubound(UU,1)
```
sanAkel, line 1994: same as above

GEOSogcm_GridComp/GEOS_OgcmGridComp.F90
```diff
@@ -2584,6 +2584,21 @@ subroutine OBIO_RunTransforms(DO_DATAATM, RC)
          VERIFY_(STATUS)
       endif
 
+do k = 1, ubound(UU,1)
```
sanAkel, line 2587: same as above

GEOSogcm_GridComp/GEOS_OradBioGridComp/GEOS_OradBioGridComp.F90
```diff
@@ -955,7 +955,8 @@ subroutine RUN ( GC, IMPORT, EXPORT, CLOCK, RC )
 !  atmosphere.  These are called "grottoes" because they are assumed
 !  to have ocean underneath with land overhead.  Set irradiance to 0
 !  to represent this condition.
-       if (slp < 0.0 .or. slp >1.0E10)then
+!       if (slp < 0.0 .or. slp >1.0E10)then ! original line, EMS
+       if ((slp < 0.0 .or. slp >1.0E10) .or. (wspd == MAPL_UNDEF)) then ! extended to include condition for wspd, EMS
```
sanAkel, line 959: I would say do it wholesale for all variables!

GEOSogcm_GridComp/GEOS_OradBioGridComp/clrtrans.F90
```diff
@@ -202,9 +202,9 @@ subroutine sfcrfl(rad,lam,aw,bw,wfac,theta,wspd,rod,ros)
...
       endif
 
-if(wspd > 1000.0) then
- wspd = 0.0
-end if
+!if(wspd > 1000.0) then
+! wspd = 0.0
+!end if
 !  Foam and diffuse reflectance
       if (wspd .gt. 4.0)then
```
sanAkel, line 209: things like `4.0` and `7.0` does not look good!

sanAkel (commented):
@gmao-esherman please comments. 

Please take print statements off. Thanks!

GEOSogcm_GridComp/GEOS_OradBioGridComp/clrtrans.F90
gmao-esherman: Agreed. This is part of the original code. I did not write this.

GEOSogcm_GridComp/GEOS_OradBioGridComp/GEOS_OradBioGridComp.F90
gmao-esherman: will do.

GEOSogcm_GridComp/GEOS_OgcmGridComp.F90
gmao-esherman: Yes, I will remove all of the debugging statements.

sanAkel:
@gmao-esherman Can https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/400 be _fixed_ ? by making whatever changes are needed?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSsaltwater_GridComp/GEOS_SaltWaterGridComp.F90
```diff
@@ -1149,112 +1265,113 @@ subroutine RUN1 ( GC, IMPORT, EXPORT, CLOCK, RC )
    VERIFY_(STATUS)
 
    do I = 1, size(GCS)
```
sanAkel, line 1267: @gmao-esherman please add comment before the **first**
```
do I = 1, size(GCS)
...
```

to imply the following: "get pointers to exports from all children, other than OBIO"

and before the **second** do loop, the comment could be: "get pointers to exports from OBIO child"

GEOSogcm_GridComp/GEOS_OgcmGridComp.F90
```diff
@@ -230,11 +230,12 @@ subroutine SetServices ( GC, RC )
 ! ----------------------------------------------------------------
 
     if (DO_OBIO/=0) then
-       OBIO = MAPL_AddChild(GC, NAME='OBIO', SS=ObioSetServices, RC=STATUS)
+       OBIO = MAPL_AddChild(GC, NAME='OBIOtest', SS=ObioSetServices, RC=STATUS)
```
sanAkel, line 233: @gmao-esherman `OBIOtest` is going to be changed to some other name of your choice, right?

sanAkel (changes requested):
This PR can be merged with suggested comments added. 

Thank you @sdrabenh

sanAkel:
this PR would also:
- fix https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/394
- fix https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/395

essentially it _seems_ to include all changes from https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/396 @gmao-esherman not sure how you pooled all those changes!

gmao-esherman:
> this PR would also:
> 
>     * fix #394
> 
>     * fix #395
> 
>     * fix #400
> 
> 
> essentially it _seems_ to include all changes from #396 @gmao-esherman not sure how you pooled all those changes!

I am pretty sure I merged in Atanas' bugfix branch into mine a little while back.

yvikhlya:
> > this PR would also:
> > ```
> > * fix #394
> > 
> > * fix #395
> > 
> > * fix #400
> > ```
> > 
> > 
> > essentially it _seems_ to include all changes from #396 @gmao-esherman not sure how you pooled all those changes!
> 
> I am pretty sure I merged in Atanas' bugfix branch into mine a little while back.

Yes, I see the related commits. This is good. #396 can be closed then.

GEOSogcm_GridComp/GEOS_OgcmGridComp.F90
gmao-esherman: Yes. I will change `OBIOtest` to `OBIO_OGCM`.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSsaltwater_GridComp/GEOS_SaltWaterGridComp.F90
gmao-esherman: What second do loop are you referring to?

sanAkel: My bad, there is only one!

gmao-esherman: I have added that comment in my latest commit.

GEOSogcm_GridComp/GEOS_OgcmGridComp.F90
gmao-esherman: The above mentioned change is included in my latest commit.

sanAkel (approved):

sdrabenh (approved):

PR #402, opened 2021-01-28
relocation of mk_restarts and Raster

weiyuan-jiang:
There is consensus among land group that mk_restarts should not be under GEOScatch_GridComp because many programs in it have nothing to do with cathment. Since those programs are all related to surface, a better choice is to relocate them under GEOSsurface_GridComp. The Shared directory is not a good place to host mk_restarts due to circular dependency. So a new directory GEOSsurface_GridComp/Utils is created to host mk_restarts. Also Raster is moved from Shared because it is more related to mk_restarts.  This PR addressed part of isse #232


smahanam:
Would you mind bringing mk_GEOSresatsrt.F90 here too? If you do so, this directory does not have to be a library anymore. Also, I can use that program for other planned development work on the AGCM side.
Raster - we can rename to mk_surface_bcs, or just surface_bcs or something.

Take or leave above comments.

tclune (approved):
cmake changes ok

weiyuan-jiang:
@smahanam Why this is important? My understanding is in the future we don't need mk_GEOSldasRestarts.F90.  We only need a script that calls mk_CatchRestarts.x or mk_CatchCNResarts.x built here.

smahanam:
No debate here :
mk_CatchRestarts.F90 is just fine, so are ScaleCatch.F90, and ScaleCatchCN.F90 as they are. mk_CatchCNRestarts.F90 needs some work to add CLM45 pre-processing for Jana's work.

Now your question about why we need it back. In the other thread, we expressed our desire to keep all codes together while GEOSldas had already left mk_restarts - keep all codes together is always my personal preference - no debate here.

I also had a plan, which was postponed or just forgotten since GEOSldas left.
mk_CatchCNRestarts.F90:
Right now, what it does is similar to the RESTART: 0 option in ldas_setup

Then,  we added RESTART: 2 option which is the culmination of few years of interactions with people that run CATCHCN -
I thought we could upgrade AGCM side and add RESTART: 2 equivalent option to the AGCM too - I was thinking to use mk_GEOSldasRestarts.F90 as part of the process. 

Again, take or leave - if Shantha sees that land has already broken the mk_restarts directory, I am signing off from that debate - I didn't break it, anyway.

gmao-rreichle (approved):
We should go ahead with this PR as is.
The comment by @smahanam about moving a GEOSldas-specific mkRestarts program should be addressed separately.  I created a GEOSldas issue [#360](https://github.com/GEOS-ESM/GEOSldas/issues/360) to that effect.

biljanaorescanin:
@sdrabenh I tested PR for AGCM one day run to confirm its zero diff, and it is for scratch and history.

sdrabenh (approved):

PR #405, opened 2021-02-02
Performance modification for surface rc file in run methods

bena-nasa:
This is a modification to help with performance at high scales and just general code cleanup.  Multiple gridded components were wanting to get a parameter from the surface resource file. Rather than loading it once in setservices and remembering, they were loading the config file every execution of run which is a file access. At NAS, at large core counts  this was causing performance issues, probably due to Lustre. In general there is just no reason load this every time. Just do it once and remember it.

I have tested that is is zero-diff and that the option is being picked up still with this change.


zhaobin74:
Was zero-diff test done with coupled model?

gmao-rreichle (approved):
Thanks, @bena-nasa , for cleaning up.  Always appreciated.

Did you look at whether there are similar issues in the Land GC?  (incl Catch, CatchCN, and vegdyn)

GEOSldas does compile the Lake GC but is not using it.  The saltwater GC is not  in GEOSldas (sparse checkout).  That is, the changes so far should be trivially 0-diff for GEOSldas.  To be safe, though, it's probably a good idea for @biljanaorescanin  to run the usual GEOSldas tests.

bena-nasa:
These were the only I found. I just wanted to get this in as this change made a world of difference for Bill Putman's run at NAS and is a small change so we should get it in going forward.

biljanaorescanin:
@gmao-rreichle its zero diff for GEOSldas.

sanAkel:
@yvikhlya when you have time, on behalf of @GEOS-ESM/ocean-team would you please check following. 

Check following coupled model runs:

- **debug** build, not checking for zero-diff. (When debugging, zero-diff is _not needed_.) Run for 1 hour, with 450 s time step. 
    - [x] MOM5 runs
    - [x] MOM6 runs

- **optim** build, Run for **6 hours**, with 450 s time step. Testing zero-diff to `develop` whose `SHA` is **??**
    -  MOM5: 
       +  [x] runs
       +  [x] zero-diff
       
    - MOM6:
       +  [x] runs
       +  [x] zero-diff

Please fill in:
- Logs are at **??**
- SHA code of develop used for checking zero-diff
Thanks!

bena-nasa:
This a very minor  code change, I confirmed that when I toggled CHOOSEMOSFC in the GEOS_Surface rc file that it was picking it up, i.e. 

I ran the stock code with both
CHOOSEMOSFC: 0 and CHOOSEMOSFC: 1 
and confirmed that this was non-zero diff so changing this changes the answer.

Then I rank 4 more experiments:

I ran with the stock code, as well as a build with this change and ran with
CHOOSEMOSFC: 0
Those two were zero diff.

Then I set
CHOOSEMOSFC: 1
ran both the stock and new code, those were also zero diff to each other

So it did appear to pickup  the CHOOSEMOSFC value so there is probably no need to run a huge amount of testing

sdrabenh (approved):

mathomp4 (approved):

PR #411, opened 2021-02-23
Fixed a bug in rotation angle for surface ocean stress.

yvikhlya:
Apparently, MOM6 uses a different definition of rotation angle on tripolar grid than CICE or MOM5 (clockwise positive vs counter clockwise positive). If sign is incorrect, this causes a buildup of sea ice along the tripolar zipper line. This PR fixes this issue. 

**This PR affects only the coupled model with MOM6.**


zhaobin74:
Looks good. Thanks,

zhaobin74 (approved):
Looks good. Thanks,

sanAkel (approved):

sanAkel:
@mathomp4 Please note: this is non-zero diff for MOM6 coupled model.

PR #412, opened 2021-02-24
Remove unneeded gfortran ifdefs

mathomp4:
The OGCM GC and the Route GC have a small chunk of code when using gfortran as a compiler. My guess is myself or @atrayano or someone added this because of issues with GNU and MAPL. 

This code needs to be removed so that the @GEOS-ESM/mapl-team can remove some no-longer-needed GNU workarounds currently in MAPL (see https://github.com/GEOS-ESM/MAPL/pull/716). But, if you remove the workarounds in MAPL, these deallocates will cause a crash!

Testing seems to indicate that this code isn't "necessary" at present, though it probably does cause a slight memory leak with GNU without this but with the workaround in MAPL. But GNU is not our operational compiler, so for testing at low-res, this isn't a problem.

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOSroute_GridComp/GEOS_RouteGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOSroute_GridComp/GEOS_RouteGridComp.F90
@@ -519,10 +519,6 @@ contains
     call ESMF_FieldDestroy(field0, rc=status)
     VERIFY_(STATUS)
 
-#ifdef __GFORTRAN__
-    deallocate(tile_area_src)
-#endif
-    
     ! redist pfaf (NOTE: me might need a second routehandle for integers)
 
     route%pfaf => arbSeq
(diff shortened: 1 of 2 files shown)


yvikhlya (approved):
I am all for removing unnecessary code.

gmao-rreichle:
@mathomp4 , @atrayano , @smahanam : 
Does this have anything to do with why the routing code has been under construction for the past couple of years?
At the GMAO Theme meeting in Nov 2018, I presented very encouraging science results from Sarith, but I remember vaguely that everything was put on hold until some software infrastructure problem could be resolved.  

cc: @rdkoster

mathomp4:
> @mathomp4 , @atrayano , @smahanam :
> Does this have anything to do with why the routing code has been under construction for the past couple of years?
> At the GMAO Theme meeting in Nov 2018, I presented very encouraging science results from Sarith, but I remember vaguely that everything was put on hold until some software infrastructure problem could be resolved.
> 
> cc: @rdkoster

@gmao-rreichle The code in question in this PR has been there since the Git repo began 2 years ago. My guess would be maybe @atrayano or myself did a Valgrind run in CVS days and found a memory leak occurring there? But I think @bena-nasa and/or @atrayano has been working with Sarith on getting Route up to date.

smahanam:
SI Teams aforementioned help came (Many thanks to Ben and Atanas), now I am in the process of coupling the river routing model to the GCM

sanAkel (approved):
Please get rid of them, Thank you @mathomp4

gmao-rreichle (approved):
Since for land this only touches the Routing GridComp, which is still under development and not yet tested routinely, I'm approving the PR so you can go ahead, assuming it's also ok with @smahanam

PR #415, opened 2021-03-15
Update to GFE Namespace

mathomp4:
GFE now uses CMake namespaces.


mathomp4:
No longer contingent as it uses a CMake option to control it and by default the namespace style is not enabled.

tclune (approved):

PR #428, opened 2021-05-10
Updates to use MAPL_GetResource in Surf

mathomp4:
NOTE: This PR requires MAPL 2.6.7 (soon to be released) because it needs https://github.com/GEOS-ESM/MAPL/pull/834 to enable the `MAPL_GetResource` call with an `ESMF_Config`

This PR is due to a request by @wmputman to have GEOSgcm output the configuration values that `GEOS_SurfaceGridComp.rc` sets a la how we see outputs in the log from `AGCM.rc`. To do this takes a few changes:

1. You need support in MAPL for calling `MAPL_GetResource` with an `ESMF_Config` (see https://github.com/GEOS-ESM/MAPL/pull/834)
2. You need to set `PRINTRC: 1` in `GEOS_SurfaceGridComp.rc` so that it will output even defaults (this echoes `AGCM.rc.tmpl`
3. Obviously, we'll also want the detemplating code in GEOSgcm_App that @sdrabenh is working on.

I also converted the calls to "match" how other `MAPL_GetResource` calls look like (config, val, label, default, rc).

This PR will be a Draft until MAPL 2.6.7 is released an in GEOSgcm and GEOSldas. Until and once MAPL 2.6.7 is in, I will work with @biljanaorescanin and @gmao-rreichle to get this "just right".


GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/GEOS_SurfaceGridComp.rc
```diff
@@ -234,3 +234,11 @@
 #
 # GEOSagcm=>SCALE_ALBFPAR: 0
 # GEOSldas=>SCALE_ALBFPAR: 0
+
+# ---- Print Resource Parameters
```
gmao-rreichle, line 238: May I suggest a slightly more verbose message here, something like:
```
# ---- Print Resource Parameter Values to Log File
#
#    0: Print only parameter values that differ from default settings
#    1: Print all parameter values
```
My aim here is to make it overly clear that this is only about printing rc values to a log file and not about setting rc values.  Seems obvious to us, but outside users may otherwise stumble here.  Just a suggestion.

sanAkel (dismissed):

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
```diff
@@ -225,12 +225,12 @@ subroutine SetServices ( GC, RC )
...
-    call ESMF_ConfigGetAttribute (SCF, label='ATM_CO2:', value=ATM_CO2, DEFAULT=0, __RC__ ) 
+    call MAPL_GetResource (SCF, ATM_CO2, label='ATM_CO2:', DEFAULT=0, __RC__ ) 
 
-    call ESMF_ConfigGetAttribute (SCF, label='N_CONST_LAND4SNWALB:', value=catchswim, DEFAULT=0, __RC__ ) 
-    call ESMF_ConfigGetAttribute (SCF, label='N_CONST_LANDICE4SNWALB:', value=landicegoswim, DEFAULT=0, __RC__ ) 
-    call ESMF_ConfigGetAttribute (SCF, label='LAND_PARAMS:', value=LAND_PARAMS, DEFAULT="Icarus", __RC__ )
-    call ESMF_ConfigGetAttribute (SCF, label='CHOOSEMOSFC:', value=CHOOSEMOSFC, DEFAULT=1, __RC__ ) 
+    call MAPL_GetResource (SCF, catchswim,     label='N_CONST_LAND4SNWALB:',    DEFAULT=0,        __RC__ )
+    call MAPL_GetResource (SCF, landicegoswim, label='N_CONST_LANDICE4SNWALB:', DEFAULT=0,        __RC__ )
+    call MAPL_GetResource (SCF, LAND_PARAMS,   label='LAND_PARAMS:',            DEFAULT="Icarus", __RC__ )
```
gmao-rreichle, line 232: I suppose this is where the current AGCM gets its default for LAND_PARAMS?  GEOSldas doesn't have GEOS_SurfaceGridComp.F90.  FP has been using "LAND_PARAMS: NRv7.2" since f525_fp.
I guess this is being changed in a separate PR?  I suppose there's a potential for confusion with the default set in GEOS_SurfaceGridComp.rc.  
Sorry if this goes beyond the scope of this PR, but I thought I should point it out while I'm looking at it. 
cc: @wmputman @sdrabenh @mathomp4

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOS_CatchCNGridComp.F90
```diff
@@ -250,61 +250,61 @@ subroutine SetServices ( GC, RC )
...
-    call ESMF_ConfigGetAttribute (SCF, label='DTCN:' , value=DTCN,  DEFAULT=5400. , __RC__ )
+    call MAPL_GetResource (SCF, DTCN, label='DTCN:', DEFAULT=5400. , __RC__ )
     ! ATM_CO2
     ! 0: uses a fix value defined by CO2
     ! 1: CT tracker monthly mean diurnal cycle
     ! 2: CT tracker monthly mean diurnal cycle scaled to match EEA global average CO2
     ! 3: spatially fixed interannually varyiing CMIP from getco2.F90 look up table (AGCM only)
     ! 4: import AGCM model CO2 (AGCM only)
-    call ESMF_ConfigGetAttribute (SCF, label='ATM_CO2:' , value=ATM_CO2,  DEFAULT=2  , __RC__ )
+    call MAPL_GetResource (SCF, ATM_CO2, label='ATM_CO2:', DEFAULT=2  , __RC__ )
```
gmao-rreichle, line 288: Default set for ATM_CO2 here is inconsistent with default set in Surface GridComp F90
cc: @gmao-jkolassa @rdkoster @sdrabenh @mathomp4

sanAkel:
@yvikhlya would you be able to check/verify which of the ⤴️ options could be used. 

It looks me, from one my _recent_ runs, as if all of `RC/GEOS_SurfaceGridComp.rc` are defaults, @wmputman: Should the ocean/coupled GCM _stay_ like that?

mathomp4:
@gmao-rreichle Since GEOSgcm has now moved to using Icarus-NLv3 by 'default' (i.e., it's what you get by pressing "Enter" for the Land BCs question in `gcm_setup`), it does make sense to change this. So, I've updated the defaults per your comments. 

Please review.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/dbg_cnlsm_offline.F90
```diff
@@ -242,7 +242,7 @@ PROGRAM dbg_cnlsm_offline
 
 is_OFFLINE = OFFLINE_MODE /= 0
 
-call ESMF_ConfigGetAttribute (SCF, label='LAND_PARAMS:', value=LAND_PARAMS, DEFAULT="Icarus", __RC__ )
+call ESMF_ConfigGetAttribute (SCF, label='LAND_PARAMS:', value=LAND_PARAMS, DEFAULT="NRv7.2", __RC__ )
```
gmao-rreichle, line 245: @mathomp4 
For CatchCN, the default should be "CN_CLM4", I think.  Both "Icarus" and "NRv7.2" should fail for CatchCN (ie, cleanly exit with a help message).  Since CatchCN isn't tested by the modeling group, you wouldn't know.
@gmao-jkolassa , can you please confirm the above?
@rdkoster

mathomp4: Yeah. I was just looking for "Icarus" and I found that one

sanAkel (commented):
All look okay to me, @yvikhlya probably would test?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/dbg_cnlsm_offline.F90
gmao-rreichle: I just realized this is only the "debug" helper program.  I went ahead and fixed the default here.
It'd be good if @gmao-jkolassa could verify the other CatchCN related defaults in the PR, including stuff like ATM_CO2 and the like.  Not sure what's best if someone is running the AGCM with CatchCN.
cc: @mathomp4 @rdkoster

sdrabenh (approved):

zhaobin74 (approved):
looks good.

PR #438, opened 2021-06-10
make_bcs is now open to public

smahanam:
As per modeling group's request, GEOS make_bcs utility is now open for everyone. 
The make_bcs script and fortran codes needed number of updates to accomplish this. 

The procedure is documented in the below user guide.
https://github.com/GEOS-ESM/GEOSgcm_GridComp/wiki/make_bcs-User-Guide


mathomp4:
@smahanam Query: In re a label for zero-diff or not, obviously it's zero-diff for the *model* since the model doesn't run it during a normal AMIP/EMIP run. But were we to run this new (?) `make_bcs`, would we get the same files we got before for our current BCs?

smahanam:
> @smahanam Query: In re a label for zero-diff or not, obviously it's zero-diff for the _model_ since the model doesn't run it during a normal AMIP/EMIP run. But were we to run this new (?) `make_bcs`, would we get the same files we got before for our current BCs?

This is zero-diff for the model runs. However, this can't reproduce some of the older BCs files (so in that sense, non-zerodiff BCs)

mathomp4:
> > @smahanam Query: In re a label for zero-diff or not, obviously it's zero-diff for the _model_ since the model doesn't run it during a normal AMIP/EMIP run. But were we to run this new (?) `make_bcs`, would we get the same files we got before for our current BCs?
> 
> This is zero-diff for model runs. However, this can't reproduce some of the older BCs files (so in that sense, non-zerodiff BCs). So please feel free to change the label.

What do you mean by "older"? Which older BCs might not be possible? I suppose that will help @sdrabenh determine how to number this.

smahanam:
> > > @smahanam Query: In re a label for zero-diff or not, obviously it's zero-diff for the _model_ since the model doesn't run it during a normal AMIP/EMIP run. But were we to run this new (?) `make_bcs`, would we get the same files we got before for our current BCs?
> > 
> > 
> > This is zero-diff for model runs. However, this can't reproduce some of the older BCs files (so in that sense, non-zerodiff BCs). So please feel free to change the label.
> 
> What do you mean by "older"? Which older BCs might not be possible? I suppose that will help @sdrabenh determine how to number this.

Old means archived F25, GM4, ICA, NL3 and NL4 BCs.

We can run this script create scientifically equivalent of old BCs, but they are not going to be zero-diff with OLD BCs.

**An excerpt from wiki page:**
(2) Select the Land BCs version to use:
Select

F25 for Fortuna-2_5
GM4 for discover/nobackup/ltakacs/bcs/Ganymed-4_0/
ICA for discover/nobackup/ltakacs/bcs/Icarus/
NL3 for discover/nobackup/ltakacs/bcs/Icarus-NLv3/
NL4 for discover/nobackup/projects/gmao/smap/bcs_temporary/NLv4/
equivalent parameter configuration and ancillary data.

(OR press ENTER if you want to use the latest develop version of land BCs.)

NOTE: Due to compiler differences, code improvements and bug fixes that have taken place since above archived BCs were created, some parameter files produced by current source codes may differ with those in archived BCs. Nevertheless, the impact of those differences on science is insignificant and the parameter files produced by current source codes can be considered as scientifically equivalence of achieved BCs.

smahanam:
> > > @smahanam Query: In re a label for zero-diff or not, obviously it's zero-diff for the _model_ since the model doesn't run it during a normal AMIP/EMIP run. But were we to run this new (?) `make_bcs`, would we get the same files we got before for our current BCs?
> > 
> > 
> > This is zero-diff for model runs. However, this can't reproduce some of the older BCs files (so in that sense, non-zerodiff BCs). So please feel free to change the label.
> 
> What do you mean by "older"? Which older BCs might not be possible? I suppose that will help @sdrabenh determine how to number this.

This is that old story about non-zerodiff issue. You may recall Matt (@mathomp4 ), you actually used Ganymed-4_0 tag to create Icarus BCs because you wanted Icarus BCs to be identical with MERRA-2 (Ganymed-4_0) BCs.

mathomp4:
@smahanam Okay. I think my answer is "Ask @smahanam when making BCs for older models, and trust in the script." 😄

smahanam:
> @smahanam Okay. I think my answer is "Ask @smahanam when making BCs for older models, and trust in the script." 😄

@mathomp4 - Please feel free to direct all questions to the land group. 

@rdkoster showed below slides at a modeling group meeting. The slides established that the impact of these non-zerodiff parameters on science is insignificant. 
https://gmao.gsfc.nasa.gov/gmaoftp/sarith/histo3.pdf 

The IDL script that produced above slides was added to the repository for future use if somebody says I can't blindly trust this script.

The differences are in ar.new, bf.dat and ts.dat files (and catch_param.nc4 - this nc4 file keep a copy of the same parameters in binary format which is what mk_restart reads to write the initial catch_internal_rst file). 

I must thank you for asking this question. When I started preparing my reply, I was surprised to find  visdf and nirdf files too differed because they were nicely zerodiff when we analyzed this issue for the modeling group. 
The problem was I had entered bug few weeks ago and spent more than a  week debugging and finally fixing it.

Below example is good for this review:
SLES12:
/discover/nobackup/smahanam/bcs/NLv3/SMAP_EASEv2_M36/

vs SLES11 in
/discover/nobackup/ltakacs/bcs/Icarus-NLv3/Icarus-NLv3_EASE/SMAP_EASEv2_M36/

Note that xxdiff will show some white space differences that is related to the bug I referred to in the previous paragraph.  make_bcs wiki page says the script supports 1km EASEv2 (which is a new addition). There are 129 million tiles - so 9 digit tile_id's could not fit in to the 'i8'  formatted fortran space and I increased it to 'i10'. SLESS11 used i8, so for this diff, I manually deleted the extra 2 white spaces that this PR had added.

biljanaorescanin:
I have two questions/comments:
1. If I run make_bcs with C90,NL3 and ocean resolution T4 and repeat that for different ocean C90, NL3 and T4MOM6 I get BCS output directory with the same name "**CF0090x6C_TM1440xTM1080**" . If this is to be created for shared drive I suppose you will need to change the name for one of the two directories. 
2. If I look at the land tile file *Pfafstetter.til in output I see 12 columns in the file but in README file that gets created with make_bcs it is listed that file has only 10 columns?  I got to this by trying to find list of what each column is. 
cc: @gmao-rreichle

yvikhlya:
>     1. If I run make_bcs with C90,NL3 and ocean resolution T4 and repeat that for different ocean C90, NL3 and T4MOM6 I get BCS output directory with the same name "**CF0090x6C_TM1440xTM1080**" . If this is to be created for shared drive I suppose you will need to change the name for one of the two directories.


T4 and T4MOM6 have the same resolution, but these are actually different grids. To avoid name conflict, I placed corresponding files into different subdirectories for now

```
/home/yvikhlia/aogcm/atmosphere_bcs/MOM5
/home/yvikhlia/aogcm/atmosphere_bcs/MOM6
```

Not sure if this is the best way to handle such situation though.

tclune:
Some convention is needed, and all choices are more-or-less equivalent.  Have to embed the MOM version somewhere.   But this does show me that there are even more axes for managing BCs than I had previously considered.

smahanam:
I am using NLv3 BCs directory as an example. Below  Icarus-NLv3_Tripolar could be two directories,  Icarus-NLv3_MOM5, and  Icarus-NLv3_MOM6


 Icarus-NLv3_EASE/
 Icarus-NLv3_MERRA-2/
 Icarus-NLv3_Ostia/
 Icarus-NLv3_Reynolds/
 Icarus-NLv3_Tripolar/
 LAND/

biljanaorescanin:
1. @gmao-rreichle In make_bcs we are missing link for 2 paths. So, GEOSldas fails when using T1 and T2MOM6 with cold start. 
We are missing clsm directory population - there is no link added. The link is not added because we are missing files here: /**discover/nobackup/projects/gmao/ssd/aogcm/ocean_bcs**
  
For T1:              360x210 to MOM5 directory, and 
For T2MOM6: 120x65    to MOM6 directory

Plus we need these two links added to make_bcs:
ln -s /discover/nobackup/projects/gmao/ssd/aogcm/ocean_bcs/MOM5/360x210 data/MOM5/360x210
ln -s /discover/nobackup/projects/gmao/ssd/aogcm/ocean_bcs/MOM6/120x65 data/MOM6/120x65

@yvikhlya are in charge of that directory? Could you add those files so I can fully test all BCS's with GEOSldas?

2.  For T4 and T4MOM6 we still have same problem during BCS creation. Still, the same name directory is created. If we run BCS creation for T4 and T4MOM6 from the same level, they will rewrite each other. I need fix for make_bcs to embed the MOM version to BCS output directory creation. We would have to separate all MOM5 and MOM6... add prefix for one or the other. @mathomp4 or @tclune if I add MOM version just to the output dir would that be ok, I know its not the prettiest of solutions because we are adding subdirectories but if we can't change one of the two runs name...? Do you have any other suggestion how to address this? @yvikhlya linking paths were not a problem, just output directory name.

yvikhlya:
> ln -s /discover/nobackup/projects/gmao/ssd/aogcm/ocean_bcs/MOM5/360x210 data/MOM5/360x210
> ln -s /discover/nobackup/projects/gmao/ssd/aogcm/ocean_bcs/MOM6/120x65 data/MOM6/120x65

We don't have such configurations. Where does this come from?

biljanaorescanin:
> > ln -s /discover/nobackup/projects/gmao/ssd/aogcm/ocean_bcs/MOM5/360x210 data/MOM5/360x210
> > ln -s /discover/nobackup/projects/gmao/ssd/aogcm/ocean_bcs/MOM6/120x65 data/MOM6/120x65
> 
> We don't have such configurations. Where does this come from?
> 
> UPDATE: we don't have T1 (MOM5 120x65) either. Is it ok to remove it?

I would remove T1 if we don't use it or don't have it. But it is on ocean group to decide. 
That linking needs to be added in make_bcs ... as I said in my comment, for CLSM directory to create **catchment.def** file which I need so I can run this resolution with GEOSldas. All other resolutions populate clsm except T1 and T2MOM6.

yvikhlya:
@biljanaorescanin @gmao-rreichle Removed T1 ocean configuration, because we don't have it. Commented out T2MOM6 ocean configuration, we don't currently have it, but @sanAkel is working on it and it is coming soon.

sanAkel:
@yvikhlya (or others), 
is there any specific reason why names such as `Tn, n=1, 2, ...` are preferred? Can a `file name` such as _whatever_ holds data for `120x65` be a user arg? I am NOT looking at this code or the set of modifications in this PR (yet). 

**Why am I asking this question**?
- Because if it was so, I would not see [above](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/438#issuecomment-878418520) references to `T1`.
- Also as far as I am concerned, names such as `T1` have no _scope_ outside this `bcs` application. For instance, when we set up a MOM`x=5,6` experiment, we don't ask it to be for some `Tn` resolution; `gcm_setup` encodes resolutions `120x65` as-is, so why bother with names such as these?

yvikhlya:
> @yvikhlya (or others),
> is there any specific reason why names such as `Tn, n=1, 2, ...` are preferred? Can a `file name` such as _whatever_ holds data for `120x65` be a user arg? I am NOT looking at this code or the set of modifications in this PR (yet).
> 
> **Why am I asking this question**?
> 
>     * Because if it was so, I would not see [above](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/438#issuecomment-878418520) references to `T1`.
> 
>     * Also as far as I am concerned, names such as `T1` have no _scope_ outside this `bcs` application. For instance, when we set up a MOM`x=5,6` experiment, we don't ask it to be for some `Tn` resolution; `gcm_setup` encodes resolutions `120x65` as-is, so why bother with names such as these?

I don't know who invented these code names. IMO it would be less confusing to point directly to MAPL_Tripolar.nc file when we generate a coupled set up.

biljanaorescanin:
@sanAkel These are only shortcuts for purpose of making this directory. They are not used in any other way in the code. It is only related to creation of BCS's. 
gcm_setup uses same shortcuts. For example, if you choose O2 in the gcm_setup, than the model will run MERRA2 in chosen resolution.  
Here is same thing in BCS:
   O2 --   Med-Resolution Reynolds 1/4 deg (Lon/Lat Data-Ocean: 1440x720) 
So, the setup here does match one used in gcm_setup. 
I've added a commit with some clarification of user's initial choice options.  
cc: @gmao-rreichle

sanAkel:
> @sanAkel These are only shortcuts for purpose of making this directory. They are not used in any other way in the code. It is only related to creation of BCS's.
> gcm_setup uses same shortcuts. For example, if you choose O2 in the gcm_setup, than the model will run MERRA2 in chosen resolution.
> Here is same thing in BCS:
>   O2 --   Med-Resolution Reynolds 1/4 deg (Lon/Lat Data-Ocean: 1440x720) 
> So, the setup here does match one used in gcm_setup.
> I've added a commit with some clarification of user's initial choice options.
> cc: @gmao-rreichle

Thanks @biljanaorescanin, I understand your reasoning here. However, can it not be made a user specified input arg to simply point to  `MAPL_Tripolar.nc` which is going to be used anyway? Again, I get it that you may want to keep those `Tn` names for whatever reason(s), but then you got to use the `MAPL_Tripolar.nc` anyway!

yvikhlya:
> > @sanAkel These are only shortcuts for purpose of making this directory. They are not used in any other way in the code. It is only related to creation of BCS's.
> > gcm_setup uses same shortcuts. For example, if you choose O2 in the gcm_setup, than the model will run MERRA2 in chosen resolution.
> > Here is same thing in BCS:
> >   O2 --   Med-Resolution Reynolds 1/4 deg (Lon/Lat Data-Ocean: 1440x720) 
> > So, the setup here does match one used in gcm_setup.
> > I've added a commit with some clarification of user's initial choice options.
> > cc: @gmao-rreichle
> 
> Thanks @biljanaorescanin, I understand your reasoning here. However, can it not be made a user specified input arg to simply point to `MAPL_Tripolar.nc` which is going to be used anyway? Again, I get it that you may want to keep those `Tn` names for whatever reason(s), but then you got to use the `MAPL_Tripolar.nc` anyway!

Santha, I would suggest to keep it the way it is for now, because we have bigger issues to resolve, namely river routing and restart regrid. We can deal with code names and other minor issues later.

gmao-rreichle:
> Thanks @biljanaorescanin, I understand your reasoning here. However, can it not be made a user specified input arg to simply point to `MAPL_Tripolar.nc` which is going to be used anyway? Again, I get it that you may want to keep those `Tn` names for whatever reason(s), but then you got to use the `MAPL_Tripolar.nc` anyway!

@sanAkel, during setup, users need to manually enter a string for the desired grid and resolution.  Using the "full" names in this process is likely to lead to errors when users have to type long alpha-numeric strings, which is avoided by the shorthand (e.g., "T4").  If you have a better suggestion, please advise.
Also, I had suggested to @biljanaorescanin that the accompanying text can be edited to make it easier for users to understand what is asked of them.  I think she made some changes but I haven't had a chance to take a look.

gmao-rreichle (approved):

mathomp4 (approved):

sdrabenh (approved):

PR #440, opened 2021-06-11
Bugfix/cakelle2/restore cnv mfc

christophkeller:
Simple bug fix that makes sure that diagnostics 'CNV_MFC' still works when the UW shallow convection code is disabled. Right now, the field `UMF_SC` in `GEOS_MoistGridComp.F90` is not being initialized if UW shallow convection is turned off, but that field is still being added to the convective mass flux diagnostics `CNV_MFC` afterwards. This bug fix resolves this issue by setting all values in `UMF_SC` to zero if UW shallow convection is turned off, as is done for the other fields specific to shallow convection.

This is a zero-diff change in the default setting with UW shallow convection turned on, and only affects diagnostics `CNV_MFC` if UW shallow convection is turned off. In the latter case, chemistry code that uses `CNV_MFC` as input goes crazy, which is how I found this bug.


narnold1 (approved):
Thanks for catching this.  This had been fixed on a feature branch last year, but looks like it never made it into develop.

mathomp4:
@narnold1 You might need to re-approve. I had to push in a CI change so that the build test could work. (@christophkeller made a PR at juuuuuuuuust the wrong time 😄 )

narnold1 (approved):

sdrabenh (approved):

PR #442, opened 2021-06-24
update ldas increment alarm setting in CatchGC 

saraqzhang:
[updated 1 July 2021 by @gmao-rreichle]
This PR makes minor changes to how the AGCM ingests increments from the land analysis in the coupled land-atmosphere DAS (LADAS).
Specifically, previously hard-wired values that control the ingestion of land increments are replaced with resource parameters "LANDASSIM_DT" and "LANDASSIM_T0" that are read with MAPL_GetResource.  The resource parameters will be added into AGCM.rc by fvsetup, see pending [ADAS PR#80](https://github.com/GEOS-ESM/GEOSadas/pull/80).
Branch was successfully 0-diff tested by @saraqzhang.

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
@@ -2855,14 +2855,14 @@ subroutine Initialize ( GC, IMPORT, EXPORT, CLOCK, RC )
        ! ADAS corrector interval/alarm
        ! Need to know length of the ADAS Corrector segment, because LDAS increments are added
        !    *only* during the Corrector segment and not during the subsequent Predictor segment.
-       call MAPL_GetResource ( MAPL, ADAS_INTERVAL, Label="ADAS_INTERVAL:", &
+       call MAPL_GetResource ( MAPL, ADAS_INTERVAL, Label="ASSIMILATION_CYCLE:", &
             DEFAULT=21600, RC=STATUS)
        VERIFY_(STATUS)
        
        call ESMF_TimeIntervalSet(Interval_c, s=ADAS_INTERVAL, rc = STATUS  )
        VERIFY_(STATUS)
 
-       ALARM_C = ESMF_AlarmCreate(name="CORRECTOR_ALARM",clock=CLOCK, &
+       ALARM_C = ESMF_AlarmCreate(name="SEGMENT_ALARM",clock=CLOCK, &
                  ringInterval=Interval_c,RC=STATUS)
        VERIFY_(STATUS)
 
@@ -4092,7 +4092,7 @@ subroutine RUN2 ( GC, IMPORT, EXPORT, CLOCK, RC )
         type(ESMF_Grid)               :: TILEGRID
         type(MAPL_LocStream)          :: LOCSTREAM
         integer, pointer              :: mask(:)
-        type(ESMF_ALARM)              :: LDAS_ALARM, CORRECTOR_ALARM
+        type(ESMF_ALARM)              :: LDAS_ALARM, SEGMENT_ALARM
 
         character(len=ESMF_MAXSTR)    :: LDASINC_FILE_TMPL
         character(len=ESMF_MAXSTR)    :: LDASINC_FILE
@@ -5172,24 +5172,24 @@ subroutine RUN2 ( GC, IMPORT, EXPORT, CLOCK, RC )
         if(LDAS_INCR >0 )  then
 
            ! LDAS increments are added *only* during the ADAS Corrector segment and not during the
-           ! subsequent Predictor segment.  This is controlled by CORRECTOR_ALARM.
+           ! subsequent Predictor segment.  This is controlled by SEGMENT_ALARM.
            
-           ! get ADAS CORRECTOR ALARM 
-           call MAPL_StateAlarmGet(MAPL,CORRECTOR_ALARM,"CORRECTOR_ALARM",RC=STATUS)
+           ! get ADAS SEGMENT ALARM 
+           call MAPL_StateAlarmGet(MAPL,SEGMENT_ALARM,"SEGMENT_ALARM",RC=STATUS)
            VERIFY_(STATUS)
            
-           if(ESMF_AlarmIsRinging(CORRECTOR_ALARM, RC=STATUS)) then
-              ! in ADAS corrector segment 
-              call ESMF_AlarmRingerOff(CORRECTOR_ALARM, RC=STATUS)
+           if(ESMF_AlarmIsRinging(SEGMENT_ALARM, RC=STATUS)) then
+              ! segment switching point  
+              call ESMF_AlarmRingerOff(SEGMENT_ALARM, RC=STATUS)
               VERIFY_(STATUS)
-              ! handle LDAS switch during ADAS corrector segment 
+              ! handle LDAS corrector switch  
               if (CATCH_INTERNAL_STATE%LDAS_CORRECTOR) then
                  CATCH_INTERNAL_STATE%LDAS_CORRECTOR = .FALSE.
               else
                  CATCH_INTERNAL_STATE%LDAS_CORRECTOR = .TRUE.
               endif
 
-           endif ! CORRECTOR_ALARM ring 
+           endif ! SEGMENT_ALARM ring 
 
            if (CATCH_INTERNAL_STATE%LDAS_CORRECTOR) then
               ! field list 
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
@@ -2856,7 +2856,7 @@ subroutine Initialize ( GC, IMPORT, EXPORT, CLOCK, RC )
        ! Need to know length of the ADAS Corrector segment, because LDAS increments are added
        !    *only* during the Corrector segment and not during the subsequent Predictor segment.
        call MAPL_GetResource ( MAPL, ADAS_INTERVAL, Label="ASSIMILATION_CYCLE:", &
-            DEFAULT=21600, RC=STATUS)
+             RC=STATUS)
        VERIFY_(STATUS)
        
        call ESMF_TimeIntervalSet(Interval_c, s=ADAS_INTERVAL, rc = STATUS  )
@@ -5185,8 +5185,6 @@ subroutine RUN2 ( GC, IMPORT, EXPORT, CLOCK, RC )
               ! handle LDAS corrector switch  
               if (CATCH_INTERNAL_STATE%LDAS_CORRECTOR) then
                  CATCH_INTERNAL_STATE%LDAS_CORRECTOR = .FALSE.
-              else
-                 CATCH_INTERNAL_STATE%LDAS_CORRECTOR = .TRUE.
               endif
 
            endif ! SEGMENT_ALARM ring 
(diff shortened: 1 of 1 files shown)


GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
```diff
@@ -5151,113 +5165,106 @@ subroutine Driver ( RC )
...
               ! in ADAS corrector segment 
               call ESMF_AlarmRingerOff(CORRECTOR_ALARM, RC=STATUS)
               VERIFY_(STATUS)
               ! handle LDAS switch during ADAS corrector segment 
-              if ( CATCH_INTERNAL_STATE%LDAS_CORRECTOR) then
+              if (CATCH_INTERNAL_STATE%LDAS_CORRECTOR) then
                  CATCH_INTERNAL_STATE%LDAS_CORRECTOR = .FALSE.
-                 !CALL WRITE_PARALLEL('C_ALARM ringing. Switching off LDAS corrector segment')
               else
                  CATCH_INTERNAL_STATE%LDAS_CORRECTOR = .TRUE.
```
gmao-rreichle, line 5189: Why do we turn **on** the "LDAS_CORRECTOR" switch at the **end** of the Corrector segment after it's been **off**?  Isn't the target use case to add the LDAS increments during the first 6 hours (ie, Corrector segment) and then turn off the LDAS increments at the end of the Corrector segment (ie, for the subsequent Predictor segment)?  I'm probably missing something.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
```diff
@@ -2838,75 +2834,97 @@ subroutine Initialize ( GC, IMPORT, EXPORT, CLOCK, RC )
...
-      call ESMF_TimeIntervalSet(Interval_c, s=ADAS_INTERVAL, rc = STATUS  )
+       
+       ! Set up LDAS_CORRECTOR switch and alarm
+       CATCH_INTERNAL_STATE%LDAS_CORRECTOR = .TRUE.
+       
+       ! ADAS corrector interval/alarm
+       ! Need to know length of the ADAS Corrector segment, because LDAS increments are added
+       !    *only* during the Corrector segment and not during the subsequent Predictor segment.
+       call MAPL_GetResource ( MAPL, ADAS_INTERVAL, Label="ADAS_INTERVAL:", &
+            DEFAULT=21600, RC=STATUS)
```
gmao-rreichle, line 2859: It looks like we still hardwire the 6-h time interval for the Corrector segment.  Is this also hardwired into the GCM for the atmospheric tendencies?  Or does the AGCM read the time interval from a resource file?  If the length of the Corrector segment is provided to the AGCM as a resource variable, we should probably read it here.

gmao-rreichle:
See inline comments/questions above. 
I also cleaned up the LDAS coupling code a bit more (comments, indents, etc).  
Here the commit (link ignores white-space diffs):  [1e3b4bb](https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/1e3b4bb0c1652e48181413f114dedbb04a2baaa8?w=1)
In the "Files Changed" tab, it's also useful to ignore white space differences:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/442/files?w=1

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
saraqzhang: @gmao-rreichle 
this is a toggle operation. take a bit of explanation.  I will send you email about it since it is not related to the PR.

saraqzhang:
Hi Rolf,
thanks for cleaning up the code. I should have removed those printouts.

(1) "LDAS_CORRECTOR" and C_ALARM"
The diagram here shows how the "C_ALARM" and "LDAS_CORRECTOR" operate in  DAS 12hr GCM run:
At restart time (21z) , C_ALARM = not ringing, and LDAS_CORRECTOR =true .  So LDAS_CORRECOTR is true till  C_ALARM =ringing  at 3z ( 6hr interval) , LDAS_CORRECTOR toggle to .false.  at 3z.
It was misleading that the name of C_ALARM and the printout message suggested it represents adas corrector segment.  It is just an alarm ringing at 6h interval.  it works based on the assumption that adas requires 6hr corrector first, then 6h predictor.  it should be referred as " predictor alarm " .

(2)  "ADAS_INTERVAL"
This parameter name is not in current resource files ( I think it used to be several years ago when the code was first implemented, not sure).   In current AGCM.rc.tmpl  it is called "ASSIMILATION_CYCLE" ,  we can change the name to that.  However, in current ADAS system setup ,  ASSIMILATION_CYCLE is kept commented off in AGCM.rc, so GCM uses the default value = 21600.


Sara


________________________________
From: Rolf Reichle ***@***.***>
Sent: Thursday, June 24, 2021 8:24 PM
Cc: Zhang, Sara (GSFC-610.1)[SCIENCE APPLICATION INTL CORP] ***@***.***>; Author ***@***.***>
Subject: [EXTERNAL] Re: [GEOS-ESM/GEOSgcm_GridComp] update ldas increment alarm setting in CatchGC (#442)


See inline comments/questions above.
I also cleaned up the LDAS coupling code a bit more (comments, indents, etc).
Here the commit (link ignores white-space diffs): 1e3b4bb<https://gcc02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FGEOS-ESM%2FGEOSgcm_GridComp%2Fcommit%2F1e3b4bb0c1652e48181413f114dedbb04a2baaa8%3Fw%3D1&data=04%7C01%7Csara.q.zhang%40nasa.gov%7Cf5e0dff0384e48f229a308d937567a52%7C7005d45845be48ae8140d43da96dd17b%7C0%7C0%7C637601667557141337%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=347cHOmDpSZcejvnjCvCTSBqj9TY55NS%2Bfjn2U1vSiQ%3D&reserved=0>
In the "Files Changed" tab, it's also useful to ignore white space differences:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/442/files?w=1<https://gcc02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FGEOS-ESM%2FGEOSgcm_GridComp%2Fpull%2F442%2Ffiles%3Fw%3D1&data=04%7C01%7Csara.q.zhang%40nasa.gov%7Cf5e0dff0384e48f229a308d937567a52%7C7005d45845be48ae8140d43da96dd17b%7C0%7C0%7C637601667557151296%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=KGFTUQmiA9Utpapuyqagl%2B8%2FXaceXokuT5cBnnG3KvU%3D&reserved=0>

saraqzhang:
@gmao-rreichle 
rename  CORRECTOR_ALARM to SEGMENT_ALARM, since it  rings every 6h at the segment changing point ( from corrector to predictor  if in adas mode of 12h  GCM run) .

gmao-rreichle:
@saraqzhang, thanks for the explanations.  Here are my 2c:

> ASSIMILATION_CYCLE is kept commented off in AGCM.rc, so GCM uses the default value = 21600.

It beats me why an important config parameter like this is "hardwired" in the ADAS (ie, using the default hidden away in the src code) and not listed explicitly in AGCM.rc.   If ASSIMILATION_CYCLE is commented out in AGCM.rc when the ADAS is running, "our" value will be whatever default we put into our MAPL_GetResource() call.  This also means that matching hard-coded defaults must be maintained in two different places...   My strong suggestion would this parameter be in AGCM.rc (i.e., **not** commented out).  This is something we should bring up with Ricardo.

> C_ALARM [...] is just an alarm ringing at 6h interval.  it works based on the assumption that adas requires 6hr corrector first, then 6h predictor.  it should be referred as " predictor alarm " .

The name SEGMENT_ALARM in the latest commit is fine.  But this still doesn't quite address my original comment.  If the AGCM run is ever extended beyond 12 hours, then the AGCM will start looking for land increments again after the first 12 hours, which doesn't make any sense.   Once LDAS_CORRECTOR has been changed to FALSE (at the end of the corrector segment), it should never be toggled back to true.  That is, lines 5188-5189 in GEOS_CatchGridComp.F90 (current  version of the branch) should be deleted.
Or am I missing something here?

saraqzhang:
they might trust the default more than a user's input on this number, which is  6h for decades and not likely  to change.  We can ask Ricardo about using fvsetup to set the parameter in AGCM.rc.tmpl 
I will delete the line 5188-5189 so ldas corrector will not be toggled to .true.  in case GCM goes beyond 12h.

saraqzhang:
After GCM gets the default number in ASSIMILATION_CYCLE,  it  sets the parameter in  ESMF Config Attributes. We can consider  the option to use MAPL_ConfigGetAtrributes to get the value ( one default by GCM top level if it is not in AGCM.rc),   therefore avoid using MAPL get resource in CatchGC  which would have our own default.

saraqzhang:
@gmao-rreichle 
the comiitted changes are
(1)  removed DEFAULT in GetResource for ASSIMILATION_CYCLE, so that we get the gcmGC config attribute value.  In current gcmGC,  this config attribute ASSIMILATION_CYCLE will be from AGCM.rc (if it is set there) , or set inline to a default (21600). 
(2) removed lines 5188-5189 in GEOS_CatchGridComp.F90, so that LDAS_CORRECTOR will not be toggled back to true.

gmao-rreichle:
Thanks, @saraqzhang.  Quick question:

> (1) removed DEFAULT in GetResource for ASSIMILATION_CYCLE, so that we get the gcmGC config attribute value. In current gcmGC, this config attribute ASSIMILATION_CYCLE will be from AGCM.rc (if it is set there) , or set inline to a default (21600).

I assume that with this change, VERIFY_(STATUS) will fail and trigger the program to exit if ASSIMILATION_CYCLE is not found in AGCM.rc (where it's not present according to your earlier comment) or has otherwise been added into the MAPL object.  But how does this relate to your earlier comment:

> We can consider the option to use MAPL_ConfigGetAtrributes to get the value ( one default by GCM top level if it is not in AGCM.rc), therefore avoid using MAPL get resource in CatchGC which would have our own default.

We're still using MAPL_GetResource and not MAPL_ConfigGetAttributes()

saraqzhang:
@gmao-rreichle 

>I assume that with this change, VERIFY_(STATUS) will fail and trigger the program to exit if ASSIMILATION_CYCLE is not >found in AGCM.rc (where it's not present according to your earlier comment) or has otherwise been added into the MAPL >object. 
yes  the program will quit  if ASSIMILATION_CYCLE is not in AGCM.rc and not found in MAPL object. 

> We're still using MAPL_GetResource and not MAPL_ConfigGetAttributes()
It turns out that MAPL_ConfigSetAttribute  ( in gcmGC that sets ASSIMILATION_CYCLE to default when it is not in AGCM.rc  )   puts the parameter value in  " config object "  that is like a resource but not a file. We can use MAPL_GetResource to get the value in catchGC, because GetResource is now designed  to look for ASSIMILATION_CYCLE in both rc file and in config object ( it calls MAPL_ConfigGetAttributes) .

gmao-rreichle (approved):

biljanaorescanin (approved):

biljanaorescanin:
I've run GEOSldas tests and all passed. Scott, I also did one day AGCM runs and its zero diff.

PR #444, opened 2021-07-07
change restart names for catchcn to make it consistent

weiyuan-jiang:
@smahanam , Is there any other codes that need to be changed to be compatible ? I will double check my ldas_setup program.


smahanam:
If you are merely changing variable names in mk_GEOSldasRestarts.F90, all changes are inside the code and your changes look OK to me.

You need run and verify, both regrid.pl and ldas_setup, that is the only way.

gmao-jkolassa:
Tested successfully.

Issue #447, opened 2021-07-29
Restructuring under GEOSocean_GridComp


yvikhlya:
This was planned since a while ago, but I can't find a related open issue. Currently, under `GEOSocean_GridComp` directory, we don't have `GEOS_OceanGridComp.F90` component. Instead, we have `GuestOcean_GridComp` and `GEOSdatasea_GridComp`, which is actually a child of `GuestOcean_GridComp`. This is inconsistent with GEOS conventions for nesting components. To resolve this inconsistency, the following is proposed:
- Create `GEOSocean_GridComp/GEOS_OceanGridComp.F90` component.
- Move functionality of `GuestOcean_GridComp/GuestOcean_GridComp.F90` into `GEOS_OceanGridComp.F90`. 
- Make `GEOSdatasea_GridComp`, `MOM_GEOS5PlugMod` and `MOM6_GEOSPlug` to be children of `GEOS_OceanGridComp.F90`.
- Delete `GuestOcean_GridComp`.
These are structural zero-diff changes. We may need help from SIteam to make `CMakeLists` correct.

mathomp4:
I know something like this has been on @sanAkel 's mind. Maybe he's done some work?

sanAkel:
> I know something like this has been on @sanAkel 's mind. Maybe he's done some work?

yes, indeed! 
@yvikhlya when I am back from break, will do a PR to address this issue. Thanks!

yvikhlya:
> > I know something like this has been on @sanAkel 's mind. Maybe he's done some work?
> 
> yes, indeed!
> @yvikhlya when I am back from break, will do a PR to address this issue. Thanks!

No worries. I already have it almost done. I'll push a branch shortly.

sanAkel:
@yvikhlya 
Could you also please move all things related to dual ocean into a subroutine within the GC? 

For example: 
- init_dualOcean()
_ run_dualOcean()
Etc

Like we have for OBIO related functionality. This will make it better readable. Thanks!

yvikhlya:
@mathomp4 I have now a new include directory GEOS_OceanGridComp, how do I tell cmake to use it instead of Guest_OceanGridComp?

sanAkel:
Given ⤴️ merged PRs, this issue can be closed.

PR #448, opened 2021-07-30
Restructuring under GEOSocean_GridComp

yvikhlya:
This PR addresses #447 and is contingent on https://github.com/GEOS-ESM/GEOSgcm/pull/317.


tclune (approved):
cmake changes ok

mathomp4:
I'm adding @GEOS-ESM/gcm-gatekeepers (aka @sdrabenh) as this PR, as @yvikhlya says, is one of those "must be done at the same time as a GEOSgcm update" things. Want to make sure he knows about this so we can coordinate it.

mathomp4:
Note: This will never pass CI here. 

It's a fun dependency issue, but the main thing is that by the time the CI system knows to checkout this branch, it's already done a `mepo clone` on the `main` of GEOSgcm. Thus, MOM5 and MOM6 are in the "wrong place" and so things go boom. You can see that the GEOSgcm CI in https://github.com/GEOS-ESM/GEOSgcm/pull/317 did build because it put MOM5 and MOM6 in the "right place". (Well, AWS failed but that's wonky at the best of times).

yvikhlya:
Also note that I renamed GEOSocean_GridComp to GEOS_OceanGridComp. This way there is no need in "OVERRIDE" statement in CmakeLists.

sdrabenh (approved):

mathomp4 (approved):
Approving as superuser.

PR #449, opened 2021-08-10
mk_GEOSldasRestart can take restart hour and restart name now

weiyuan-jiang:


gmao-jkolassa:
@weiyuan-jiang , I completed the coupled model simulation with this branch (and the associated branches) and the science results from them look good too. Note, that my experiments used restart files at the correct resolution. Please let me know if you would also like me to test the restart regridding in the context of GCM simulations.

PR #451, opened 2021-08-27
Moved lightning flash rate (LFR) calculation from MOIST to ChemEnv

mmanyin:
This pull request is one of three that must be done together (GEOSgcm_GridComp, GMAO_Shared and GEOSchem_GridComp). The result is zero-diff for PCHEM simulations. It is non-zero-diff for simulations that use Lightning Flash Rate, such as GMI.
This set of PR's is focused on the Lightning Module, which now computes both Flash Rate and NOx Emissions. The module has been moved from GMAO_Shared to Chem_Shared, since its output is used only within Chemistry code. The lightning module has also been significantly re-worked. Different Flash Rate algorithms are now available.


sdrabenh:
@mmanyin after testing these 3 branches/PRs, it appears to be non-zero-diff. In both AMIP and replay modes, the `CH3I` variable has slightly different values in `tr_internal_checkpoint` and the `FROCEAN` variable was dropped from `moist_import_checkpoint`. Obviously, the same differences occur with `CH3I` in the history `geosgcm_tracer`. Are these changes intended as this is listed as 0-diff?

mmanyin:
@sdrabenh indeed you are correct, this is an oversight of mine!  The changes came in the CHEM PR 137 (https://github.com/GEOS-ESM/GEOSchem_GridComp/pull/137) over 2 months ago, not the trio of PR's from last week.  And my recent zero-diff tests were against 'develop', not the previous release-- my fault.  So where does that leave us?  (1) a small number of passive tracers (TR) are non-zero-diff due to re-factoring (on the order of round-off), and one of those (CH3I) is run as part of the standard release; none of the TR tracers has any effect on any other part of the model, so this difference is extremely minor, but it does exist.  Issue (2): the FROCEAN field is no longer needed by MOIST-- admittedly I should have noted this difference, but does this "removal of deadwood" trigger a non-zero-diff flag?  Bottom line-  should we postpone the 3 PR's to make for a cleaner release today, and label them all "non-zero-diff" for the next round of updates?  Please advise.

sdrabenh:
@mmanyin thanks for the explanation on what is going on. These changes seem minor, but nevertheless, still non-zero-diff. I think the best course of action is to save these 3 PRs for the next round of non-zero-diff updates. I would be more tempted to take it now if it was only dropping FROCEAN. However, some tracer values are different. I plan to take these PRs soon since there are other non-zero-diff updates waiting in the wings regarding orbital changes and compiler upgrades. Does this sound reasonable?

mmanyin:
@sdrabenh  yes, your plan to include these in a non-zero-diff release sounds like the right way to go.  Sorry for the confusion.

sdrabenh (approved):

PR #459, opened 2021-09-23
Closes #458. Makes Satsim Subcolumns Resolution-Based

### LINKED ISSUE #458: Make Satsim Columns Resolution-Based
Per [a comment by @wmputman yesterday](https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/db8ce15c17385973c8d739aae3f55dfd1432b538#commitcomment-56896280):

> @sdrabenh and @dr0cloud we need to get this update for GEOS_SatsimGridComp.F90 onto develop. It is zero diff for resolutions c720 and coarser, but required for memory purposes as we go to higher resolutions.

mathomp4:
This PR adds code to Satsim to make it choose subcolumns based on horizontal resolution.

Per @wmputman: 

> It is zero diff for resolutions c720 and coarser, but required for memory purposes as we go to higher resolutions.

I'll do some tests to confirm.

NOTE: The 0-diff and non-zero-diff labels reflect this fact. For all normal users, it's zero-diff.  For @wmputman-level users, it's not, but, well, at C1000+ you are a pioneer user and shouldn't expect zero-diff. 😄


mathomp4:
Query for @wmputman, I think this code would fail at C10000+ due to `imchar` being 4 characters long. Should we future proof this? Or is C10000+ such a pipe dream at this point, it's not worth the effort (and I'm guessing other places in GEOS would fail then as well).

mathomp4:
Okay. I can confirm that at least at C90 the satsim output is identical with this change. I turned on all the satsim variants and enabled the:

* geosgcm_conv
* geosgcm_lidar
* geosgcm_parasol
* geosgcm_modis
* geosgcm_radar
* geosgcm_isccp
* geosgcm_misr

collections in `HISTORY.rc` and they were all identical.

mathomp4:
Blocking for a bit. I see a possible issue.

mathomp4:
Yeah. There is some incorrect Fortran here. The issue is:
```fortran
    Ncolumns = MIN(30,MAX(1,INT(4*5760*4/imsize)))
    call MAPL_GetResource(MAPL,Ncolumns,LABEL="SATSIM_NCOLUMNS:",default=Ncolumns,  RC=STATUS)
    VERIFY_(STATUS)
```
Note how `Ncolumns` is being used as both an output and an input. Turns out this is illegal Fortran. @tclune made a little tester:
```fortran
module foo_mod
   implicit none
contains
   subroutine foo(x,y)
      real, intent(inout) :: x
      real, intent(in) :: y
      x = y*2
   end subroutine foo
   subroutine bar()
      real :: x
      x = 5.
      call foo(x, x)
   end subroutine bar
end module foo_mod
program main
   use foo_mod
   call bar()
end program main
```
and NAG returns:
```console
$ nagfor -C=all t.F90 
NAG Fortran Compiler Release 7.0(Yurakucho) Build 7057
[NAG Fortran Compiler normal termination]
$ ./a.out
Runtime Error: t.F90, line 10: Assignment to X affects dummy argument Y
Program terminated by fatal error
Abort trap: 6
```

It turns out there are quite a few of these around GEOS. I might make a meta-issue on GEOSgcm to track the ones I find.

sdrabenh (approved):

Issue #464, opened 2021-10-06
MAPL_VarReads in landsurface regridding codes lack error checking!


bena-nasa:
We were trying to debug an issue with regrid.pl, that it was giving different answers relative to some previous release. Close inspection the log revealed a bug of error messages that the start+count during a netcdf read was exceeded but because the rc status of MAPL_VarRead is not checked in mk_CatchRestart.x code this just slipped through and produced a restart (albeit with 0 values for some variables).
If it had been being checked the code should have died and regrid.pl would have failed alerting us to a problem.
So mk_CatchRestart.x and any other code in the surface regridding directory that uses MAPL_VardRead or any other subroutine or function that has an rc code needs to be checked!

PR #467, opened 2021-10-07
add error checking in mk_restarts executables

weiyuan-jiang:
This PR addresses issue #464 . @gmao-jkolassa , Would you please help test this branch with regrid.pl. ( catch, catchcnclm40, catchcnclm45, maybe landice)


gmao-jkolassa:
Thanks @weiyuan-jiang! I'll go ahead and test this for all the cases you mentioned.

gmao-jkolassa:
@mathomp4 When running this for catch only there is still an error message that appears, which does not impact the results. 
The error message is:
`Use of uninitialized value $cnlist[0] in concatenation (.) or string at /discover/nobackup/jkolassa/regrid_test/GEOSgcm/install/bin//regrid.pl line 2766, <STDIN> line 6.`
@weiyuan-jiang has a fix for it by changing some lines in regrid.pl, but since this would involve another GMAO_Shared PR we were wondering whether to put that in now or wait until another GMAO_Shared PR comes along (since as I said this does not actually impact the results for catch or catchcn). 

What do you think?

sdrabenh (approved):

PR #479, opened 2021-11-02
Adding SHOC+EDMF updates

narnold1:
Major update to Turbulence, with changes to SHOC scheme and new EDMF updraft scheme.  Also includes new double-gaussian cloud PDF option in MoistGC, and small changes to DatmoDyn.


tclune (approved):
cmake changes ok

sdrabenh:
@narnold1 I'm re-labeling this PR to non-0-diff because there are different variables in moist and turbulence restarts, so the restarts are not the same. Is the expectation that the old/existing variables will be 0-diff?

narnold1:
The restart changes all involve variables that are unused with default settings, so the model output remains 0-diff.  The update removes a number of current turb_internal variables that were used by the old version of SHOC (no longer needed), and there are 2 new turb_internals and 1 new moist_internal, again only needed with non-default settings.

mathomp4 (approved):

mathomp4 (approved):

mathomp4:
@narnold1 Would like me (or others on @GEOS-ESM/cmake-team) to add compiler definition to the CMake? I see things like:
```fortran
!#define UWDIAG 1
!#define PDFDIAG 1
...
!#define EDMF_DIAG 1
!#define USE_SCM_SURF 1
```
And we could make those options that could be set at CMake time (`-DPDFDIAG=ON`) or the like if it's easier for you?

mathomp4:
@narnold1 Also: another question for me: will this make the SCM non-zero-diff? I see you updated Datmo so "probably", is my guess?

narnold1:
@mathomp4, Adding compiler definitions wouldn't be worth it.  These options won't be used very often.

I *think* the SCM should still be 0-diff, though I haven't verified that.  The changes in DatmoDyn are mostly adding options, not changing existing code.

mathomp4:
@narnold1 Okay. Just thought I'd offer it. If it does become something you'd like, just let me know. It's a simple couple lines in CMake to enable it.

sdrabenh (approved):

mathomp4 (approved):

PR #492, opened 2021-11-29
Feature/pnorris/#422 #443 rrtmg lw sw combo generator

dr0cloud:
This is a large non-zero-diff PR with fastLW+SW that tidies up RRTMG LW and SW and makes them both about 2/3 the runtime and makes them use a combined cloud generator. The changes have been discussed and vetted in modeling group over the past 2-3 months. Most changes are "statistically zero-diff" in the sense that they have minimal difference (~< 0.01 W/m2) in global TOA fluxes (in 3 month MERRA-2 replays) and minimal changes in zonal means. But any difference in the stochastic cloud generator (even changing the PRNG) will have large instantaneous changes in clouds (and therefore radiative fluxes) that will average to near-zero over space and time. The few changes that were statistically non-zero-diff have been understood as being better.


dr0cloud:
Scott, please also merge to ORBIT default on (AGCM.rc add two lines) at or near this PR (before or after). Thanks, Peter.

tclune (approved):
CMake changes

mathomp4:
Per @Gvilla1000-nasa, he says that this might be a GNU bug. He encountered something similar with an:
```fortran
  if (function()) then
```
where he instead had to do:
```fortran
  condition = function()
  if (condition) then
```

GNU had some issue with doing an if-test on a function itself. (But apparently Intel and NAG were okay with doing that.)

mathomp4:
> Scott, please also merge to ORBIT default on (AGCM.rc add two lines) at or near this PR (before or after). Thanks, Peter.

@sdrabenh Just making sure you see this. @dr0cloud can you put the two lines here? Or do you have a PR in GEOSgcm_App that does it?

sdrabenh:
> > Scott, please also merge to ORBIT default on (AGCM.rc add two lines) at or near this PR (before or after). Thanks, Peter.
> 
> @sdrabenh Just making sure you see this. @dr0cloud can you put the two lines here? Or do you have a PR in GEOSgcm_App that does it?

@mathomp4 and @dr0cloud I can add these changes to this PR

dr0cloud:
Thanks Scott. And Matt, do we need to do anything about this GCC bug? Or can we just leave it as is?

mathomp4:
> @sdrabenh @mathomp4 Thanks Scott. And Matt, do we need to do anything about this GCC bug? Or can we just leave it as is?

@dr0cloud I'm doing some testing now to try and figure out the "right fix" to suggest you try out.

dr0cloud:
@mathomp4 How about ...
#if (Compiler == GCC) go take a hike!
Or is that a no no?

mathomp4:
@dr0cloud Well, the solution was pretty simple. I think we just need to make:
```fortran
   function condensate_inhomogeneous result (inhomo)
```
be:
```fortran
   function condensate_inhomogeneous() result (inhomo)
```
For some reason, parallel make is "hiding" the real error which can be seen in serial build. When I make that change, it seems to build. I also then did a test run of GEOS (GNU Debug) and, well, nothing crashed. So that's nice.

dr0cloud:
@mathomp4 Matt fixed it .. need "function f() result (r)" ... not "function f result (r)" for GCC. Thanks Matt!

mathomp4 (approved):

PR #496, opened 2021-12-08
Make big 5-d buffer in Moist optional

mathomp4:
This adds a new `AGCM.rc` resource: `USE_MOIST_BUFFER` that allows one to avoid the big 5-d `buffer` array in Moist.

Tests at C24 L72 and C720 L181 show this is zero-diff if `USE_MOIST_BUFFER: .TRUE.` or `USE_MOIST_BUFFER: .FALSE.`

I'm going to keep this in Draft mode until @tclune and @wmputman have a chance to look at it. And maybe `USE_MOIST_BUFFER` is a bad name for this. :smile:


tclune:
Hard to see how any short name can capture the intent here.   This is a time vs memory tradeoff switch.   Not liking anything that immediately occurs to me in that sense though.

bena-nasa (approved):

sdrabenh (approved):

PR #500, opened 2021-12-15
WORK IN PROGRESS: Updates for Standards Conformance (NAG)

mathomp4:
This PR tracks changes needed to compile GEOSgcm_GridComp with the NAG Compiler. NAG is *very* strict and will not allow the use of Fortran extensions that are endemic to GEOS. 

---

The first category are procedures that are now part of the Fortran Standard that used to be extensions in some compilers:

- `iargc()` → `command_argument_count()`
- `getarg()` → `get_command_argument()`
- `getenv()` → `get_environment_variable()`
- `system()` → `execute_command_line()`
- `call exit(0)` → `stop`
- `call exit(N)` → `error stop N`

NAG also doesn't support (without flags that need to be used *everywhere* as @tclune can elucidate on) the use of `real*4` or `real(kind=4)`. The reason is that NAG does not use 4 for 32-bit real and 8 for 64-bit, instead they use 1 and 2. (The Standard does not specify what "kind" a 32-bit real must be, only that there must be one.) So, many changes are of the type:

- `real*4` → `real(REAL32)`
- `real(kind=8)` → `real(kind=REAL64)`

Finally, in this, some catch code uses `nan` and `inf` defined like:
```fortran
  real*8, parameter :: inf8 = O'0777600000000000000000'
  real*8, parameter :: nan8 = O'0777610000000000000000'
  real*4, parameter :: inf4 = O'17740000000'
  real*4, parameter :: nan4 = O'17760000000'
  real,   parameter :: inf = inf4
  real,   parameter :: nan = nan4
``` 
NAG does not allow this as it sees these as, well, not numbers at compile-time. So with @tclune we made functions to return these values via `ieee_arithmetic` a la:

```fortran
   function nan() result(nan_32)
      real(REAL32) :: nan_32

      nan_32 = ieee_value(nan_32,  ieee_quiet_nan)

   end function nan

   function inf() result(inf_32)
      real(REAL32) :: inf_32

      inf_32 = ieee_value(inf_32,  ieee_positive_inf)

   end function inf
```


sdrabenh:
@mathomp4 mind if I close this for now? It looks very old

mathomp4:
> @mathomp4 mind if I close this for now? It looks very old

@sdrabenh Actually, keep it open for now and let me see if I can merge in develop. If it doesn't work, I'll do it. This branch was too much work for me to let go of now!

mathomp4:
Okay. The merge worked, which is good. For now I'll close the PR and can always bring it back if needed.

PR #503, opened 2021-12-17
fix timers in catchcn

weiyuan-jiang:
The PR fixes mis-matches of timers and removes some tab keys in CatchCN gridcomp. It has been tested even though it won't affect GEOSgcm.
GEOSldas's PR https://github.com/GEOS-ESM/GEOSldas/pull/487  depends on this PR.  @gmao-rreichle @sdrabenh


gmao-rreichle (approved):

gmao-rreichle:
@sdrabenh, @mathomp4 : Unfortunately, there is possibly some urgency to merging this PR, and we would also need a GEOSgcm_GridComp release to go with it.  (No need for a GEOSgcm release.)  This is because we had hoped to create a GEOSldas release using the latest version of MAPL (v2.14.0), which contains a bug fix that is necessary for GEOSldas.  But MAPL v2.14.0 is also much less tolerant of our less-than-perfect calls to the MAPL_Timer routines.  
I recognize that the PR and request for a GEOSgcm_GridComp release is very short notice.  It **is** ok if it's not possible or desirable to merge this PR and create a GEOSgcm_GridComp release immediately.  We **can** issue a GEOSldas release using MAPL v2.10.1 and GEOSgcm_GridComp v1.13.0.  This new release just wouldn't fix the timer issues, but they can also be fixed later. 
**The important thing is to make a decision either way, so please let us know**, thanks, Rolf
cc: @weiyuan-jiang, @biljanaorescanin

sdrabenh (approved):

PR #507, opened 2021-12-20
Fixes #506. Initialize uninitialized part of selffac array

### LINKED ISSUE #506: Crash in rrtmg_lw_setcoef in develop Intel Debug
Nightly tests show a crash in `rrtmg_lw_setcoef.F90` running with Intel Debug:
```
forrtl: error (75): floating point exception
GEOSgcm.x          000000000F24BBBB  Unknown               Unknown  Unknown
libpthread-2.22.s  00002B9F9F156CE0  Unknown               Unknown  Unknown
GEOSgcm.x          0000000009ACA891  rrtmg_lw_setcoef_         563  rrtmg_lw_setcoef.F90
GEOSgcm.x          0000000009AF94B2  rrtmg_lw_rad_mp_r         577  rrtmg_lw_rad.F90
GEOSgcm.x          0000000009AE2E3E  rrtmg_lw_rad_mp_r         338  rrtmg_lw_rad.F90
GEOSgcm.x          000000000960D3CE  geos_irradgridcom        3327  GEOS_IrradGridComp.F90
GEOSgcm.x          00000000095514B6  geos_irradgridcom        1245  GEOS_IrradGridComp.F90
```

Pointing to:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/79613015551cf5d21f52f1fcad3167672b070c37/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSradiation_GridComp/GEOSirrad_GridComp/RRTMG/rrtmg_lw/gcm_model/src/rrtmg_lw_setcoef.F90#L563

Fix is being looked at now.

mathomp4:
Per #506, the `selffac` array is only partially-filled, when the code then tries to run over the entire array, a crash occurs in debugging mode.

In the old code (v10.13.1 and older), this was set by:
```fortran
!  Set up factors needed to separately include the water vapor
!  self-continuum in the calculation of absorption coefficient.
         selffac(lay) = water * forfac(lay)
```
in the upper atmosphere. However, code analysis by @dr0cloud found that `selffac` isn't actually ever *accessed* in RRTMG LW in the upper atmosphere. So in the PR we instead set `selffac` to zero in the upper atmosphere to save on unneeded operations.

Runs with both Release and Debug builds show that this is indeed true and is zero-diff. 

Closes #506

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSradiation_GridComp/GEOSirrad_GridComp/RRTMG/rrtmg_lw/gcm_model/src/rrtmg_lw_setcoef.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSradiation_GridComp/GEOSirrad_GridComp/RRTMG/rrtmg_lw/gcm_model/src/rrtmg_lw_setcoef.F90
@@ -507,6 +507,18 @@ contains
                indfor(lay,icol) = 3
                forfrac(lay,icol) = factor - 1.
 
+               !  Set up factors needed to separately include the water vapor
+               !  self-continuum in the calculation of absorption coefficient.
+               !  In the old code this was set as:
+               !
+               !     selffac(lay,icol) = water * forfac(lay,icol)
+               !
+               !  However, selffac is never used outside of this routine in 
+               !  the upper atmosphere. So to save operations we just set to zero.
+               !  Tests confirm that this is zero-diff to actually filling the
+               !  array with the product above.
+               selffac(lay,icol) = 0.0
+
                ! factors needed to separately include the minor gases
                ! in the calculation of absorption coefficient
                scaleminor(lay,icol) = pavel(lay,icol)/tavel(lay,icol)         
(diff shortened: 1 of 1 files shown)


dr0cloud:
Confirmed. selffac is only use in rrtmg_lw_taumol and only in the lower atmosphere.

sdrabenh (approved):

dr0cloud (approved):
Yes, this solves a floating-point exception problem when compiled debug mode. The bug had no impact on any earlier model results, but it now runs when compiled in debug mode.

dr0cloud:
Note that this bug had no effect on previous runs/results. It just prevents the model from crashing when compiled in debug mode (as in the nightly tests). The uninitialized variable which trigged the floating-point exception is never used in the upper atmosphere (except for being scaled by another variable colh2o, but that scaled value is never used PERIOD in the upper atmosphere). That’s why the value selffac is set to in the upper atmosphere branch is not important …setting it to 0 with a comment as Matt has done is the best solution for speed and readability.

PR #516, opened 2022-01-05
fix mk_GEOSldasRestart for catchcnclm45

weiyuan-jiang:
When using CatchCNCLM45 with RESTART =0 or 2, the variables VAR_PFT and VAR_COL are wrong and fixed in this PR.


gmao-rreichle:
Thanks, @weiyuan-jiang, for the prompt fix. 
Since the fix touches only CatchCNCLM45, it should be trivially 0-diff for all nightly tests.  I changed the label to "0-diff trivial".  No need to run tests. 
@biljanaorescanin, did you verify that the fix solves your problem?  If so, then please approve and merge.

biljanaorescanin:
I can confirm this does fix the CatchCNCLM45 problem.

biljanaorescanin (approved):

PR #517, opened 2022-01-10
make_bcs clarification of options

biljanaorescanin:
This PR needs to make clear separation of what user is choosing on screen for boundary conditions creation. We had recent issue where under development branch was used as suggested default. Hopefully this PR will help with that. 

I've added to PR two small fixes one is mask we use for EASE grid. To be aligned with what archived EASE grid boundary conditions used in past. Other is format when CatchCN_CLM4.5 boundary conditions are created. 

Since all of the changes are only triggered during run for boundary conditions creation package it is a zerodiff PR. 

@gmao-rreichle @sanAkel let me know if either of you wants me to add or remove something. 

-----------------------------------------------------------------------------------
UPDATED 9 Feb 2022 by @gmao-rreichle:

The PR now includes the following changes:
- `make_bcs` (primary user interface)
  - fixed bugs in functionality, incl. better protection against user input errors
  - improved documentation
  - cleanup
- `mkCatchParam.F90`
  - improved log statements
- `create_README.csh`
  - corrected documentation, incl. 'format' statements
  - cleanup
- `mk_GEOSldasRestarts.F90`
  - fixed bug in handling of CatchmentCNCLM45 restarts
  
The PR is trivially 0-diff for all current model and GEOSldas tests because it only touches code that is not exercised in any of the current tests.  The land group verified that bcs generated after and just before the above modifications are 0-diff.  
(Note that the archived bcs in /discover/nobackup/ltakacs/bcs were generated with a much older version and are not 0-diff against bcs generated with the current make_bcs package.  The land group verified that the archived bcs are scientifically equivalent to bcs  generated with the current make_bcs package.)

@sdrabenh @rdkoster @biljanaorescanin


gmao-rreichle:
@biljanaorescanin, 
Thanks for putting this together.   Re. your comment:

> I've added to PR two small fixes one is mask we use for EASE grid. To be aligned with what archived EASE grid boundary conditions used in past.

Are the bcs generated before and after this change 0-diff?  Even though there isn't a nightly test for make_bcs, we still need to make sure and test that the make_bcs package is 0-diff.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
```diff
@@ -138,25 +138,26 @@ endif
...
+echo "                    the parameter files produced by current source code is considered to be"
+echo "                    scientifically equivalent to the corresponding archived BCs"
 echo "                                      "
-echo "              (OR press ENTER If you want to use the develop version of land BCs.)"
+echo "              OR press ENTER to select NL3 (current default)."
 set dummy = `echo $<`
 set dummy = `echo $dummy | tr "[:lower:]" "[:upper:]"`
-set lbcsv =  DEF
-if( $dummy == 'F25'  | $dummy == 'GM4'  | $dummy == 'ICA'  | $dummy == 'NL3'  | $dummy == 'NL4'  | $dummy == 'NL4p') set lbcsv = $dummy
+set lbcsv =  NL3
```
gmao-rreichle, line 159: @biljanaorescanin : I don't think we need a separate "DEF" value for the default of "lbcsv".
I assume this line sets the default value for "lbcsv".  
It would be better to set the "lbcsv" default value before the "echo" commands and then print the value of "lbcsv" in line 156 (rather than hard-coding "NL3" there as well as here).   Please implement this if it sounds ok to you.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
```diff
@@ -263,7 +264,7 @@ else
 endif
 
 if($HRCODES == m1 | $HRCODES == m3 | $HRCODES == m9 | $HRCODES == m36 | $HRCODES == m25) then
-     set GLOBAL_CATCH_DATA = ${l_data}/GEOS5_10arcsec_mask.nc
+     set GLOBAL_CATCH_DATA = ${l_data}/GEOS5_10arcsec_mask_freshwater-lakes.nc
```
gmao-rreichle, line 267: @biljanaorescanin, I never quite understood what the deal was with this "mask" file.  Is the "freshwater-lakes" mask file different from the file without "freshwater-lakes" in the name?  If so, then this would presumably be a non-zero diff change to make_bcs, which is not ideal. 
Which file was used in the archived bcs?  
This probably requires more thought/discussion.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkCatchParam.F90
```diff
@@ -338,7 +338,7 @@ PROGRAM mkCatchParam
        inquire(file='clsm/ndvi.dat', exist=file_exists)
        if (.not.file_exists)  call gimms_clim_ndvi (nc,nr,gridnamer)
 
-       write (log_file,'(a,a,a)')'Done computing ', trim(LAIBCS),' vegetation climatologies .............4'
+       write (log_file,'(a,a,a)')'Done computing ', trim(LAIBCS),' vegetation climatologies ............4'
```
gmao-rreichle, line 341: What does the "4" mean at the end of this line?  Was this change meant to remove the "4", but the change inadvertently removed the dot?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
```diff
@@ -5754,7 +5754,7 @@ SUBROUTINE CLM45_fixed_parameters (nc,nr,gfiler)
        if(count_pix(i,3) > 0.) abm_int = NINT(abm  (i) / count_pix(i,3))
        if(count_pix(i,4) > 0.) hdm_r   = hdm   (i) / count_pix(i,4)
              
-       write (10,'(2I10, i3, f8.4, f8.2, f10.2, f8.4)' ) tid, cid, abm_int, peatf_r, gdp_r, hdm_r, field_cap(sc_com)  
+       write (10,'(2I8, i3, f8.4, f8.2, f10.2, f8.4)' ) tid, cid, abm_int, peatf_r, gdp_r, hdm_r, field_cap(sc_com)
```
gmao-rreichle, line 5757: @biljanaorescanin, are you sure this should be 2I8 and not 2I10?  Elsewhere in this file, the variables "tid" (presumably, tile ID) and "cid" (Catchment ID?) are written using 2I10.
Shorter fields here may get us in trouble with high-resolution tile spaces.

biljanaorescanin: So we have two options here, one we leave it as it is here and change format here: https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/develop/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/mk_GEOSldasRestarts.F90 line 1420 and 1477 or we change format here during BCS creation for clm4.5. Anyways this is bug fix, in either case since under current setup run will fail with CatchCNCLM45.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
biljanaorescanin: This is something we should discuss. Some archived BCS for EASE use this mask some don't. I will followup with you on this one.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
biljanaorescanin: I agree. We can remove this. I just wasn't sure do you want to keep that as it was.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkCatchParam.F90
biljanaorescanin: This is really just so everything is aligned in log file.  No science in this one. Once BCS are created we create mkCatchParam.log if we want all numbers to be one under another we have one extra space...

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
sanAkel: It seems that the entire block ⬇️ which selects the way Great Lakes and Caspian are to be treated should be done carefully!
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/7af27bac78690bcc6f3b94be7f9566ef63777953/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs#L235-L268

This specific change from @biljanaorescanin reg mXX, XX=1, 3, 9, 36, 25 sets it for the EASE grids, so I won't say anything about that! However, the blocks above, which set for O1, T2- T4, etc are highly restrictive. I guess _this_ PR isn't the place do anything about them. But we should do something to have a _cleaner_ logic ...

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkCatchParam.F90
sanAkel: I find `4` to be confusing!
When something is done, it is done!

I would recommend changing the print statement to something simple such as:

>print *, 'Done computing vegetation climatologies: ', trim(LAIBCS)

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/rmTinyCatchParaMod.F90
```diff
@@ -109,14 +109,14 @@ SUBROUTINE init_bcs_config (LBSV)
...
+     case ("NL5")
        LAIBCS  = 'MODGEO'
        SOILBCS = 'HWSD'
        MODALB  = 'MODIS2'
        process_peat = .true.
        jpl_height   = .true.
 
-    case ("DEF")
+    case ("DEV")
        LAIBCS  = 'MODGEO'
```
sanAkel, line 120: Rather than repeating:

> process_peat = .F.
jpl_height   = .T.

and so on for different cases,
it is simpler to set:

>
process_peat = .F.
jpl_height   = .F.

in the beginning and for each case, only set `.T.` those that need to be set,
for e.g.,
>case ("NL4")
       ...
       ...      
       jpl_height   = .true.

A similar approach can be taken for `MODALB` which is always being set to `MODIS2` except for `F25`.

I always like writing shortest code! 😄

sanAkel (dismissed):
Thanks for copying me on your PR!

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
gmao-rreichle: @biljanaorescanin,

>  Please implement this if it sounds ok to you.

I went ahead and changed the logic of how the default is set.  I can't promise that I got the syntax right.  We'll need to test everything once we're done with editing. 

Note that I also added some logic to let the user know if they entered an invalid three-character code for lbcsv.  I didn't see this elsewhere in the script but might have missed it.  A similar exit message might be needed for the other interactive inputs.  Please take a look and implement if needed.

biljanaorescanin:
I changed run setup. Is that what you had in mind @gmao-rreichle ? Now user runs from build directory and makes on screen choice or experiment dir.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
```diff
@@ -153,11 +160,15 @@ echo "                    Nevertheless, the impact of these differences on scien
...
-echo "              OR press ENTER to select NL3 (current default)."
+echo "              OR press ENTER to select $lbcsv (current default)."
 set dummy = `echo $<`
 set dummy = `echo $dummy | tr "[:lower:]" "[:upper:]"`
-set lbcsv =  NL3
-if( $dummy == 'F25'  | $dummy == 'GM4'  | $dummy == 'ICA'  | $dummy == 'NL3'  | $dummy == 'NL4'  | $dummy == 'NL5'| $dummy == 'DEV') set lbcsv = $dummy
+if( $dummy == 'F25'  | $dummy == 'GM4'  | $dummy == 'ICA'  | $dummy == 'NL3'  | $dummy == 'NL4'  | $dummy == 'NL5'| $dummy == 'DEV') then
+    set lbcsv = $dummy
+else
+    echo "You selected lbcsv=$dummy   !!!! Invalid Version, stopping !!!! "
```
biljanaorescanin, line 233: @gmao-rreichle This won't work because we always set default if nothing is chosen.  Do you want to remove default and have this invalid version message? But then we don't have this option just enter? If we do that ( remove enter) we can't use exclamation marks that is interpreted as history command we have to use like **** or something.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkCatchParam.F90
gmao-rreichle: I just added another commit with somewhat major revisions to the status ("done...") messages in mkCatchParam.F90.   For the most part, this should take care of the comment by @sanAkel.  I did my best not to change the logic of the code and only edit that status messages.  This will need testing.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
gmao-rreichle: In mod_process_hres_data.F90, ASCII files with "tid, cid" in the first two columns use mostly "i10,i8" (files: catchment.def, mosaic_veg_typs_fracs, soil_param.first, tau_param.dat) except "2i10" in subroutines ESA2CLM_45() and ESA2CLM(), and "2i8" in subroutine CLM_fixed_parameters().  Looks like the CatchCN stuff is inconsistent.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/rmTinyCatchParaMod.F90
gmao-rreichle: @sanAkel, shorter code is usually better, but in this case I think we should leave it as is.   It's easier to read what's in each case when we don't have to piece together the settings by whatever changes from a difficult-to-define default. 
Having said that, the switches themselves are somewhat confusing.  For example, "process_peat" really means that we want to use PEATMAP ancillary data and model parameters from Bechtold et al 2019(?).  The FP bcs (NLv3) do already have  peat tiles based on HWSD and Woesten et al soil parameters.  A more sensible approach would be to specify the ancillary data source for peat tiles, which then leaves room for an updated version of, say, PEATMAP or an entirely new peat dataset. 
I will "resolve" this sub-conversation because they have either been addressed or are showing up elsewhere (ie, the "mask" issue).

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
```diff
@@ -7,8 +7,7 @@
 #  
 #                             ---- USAGE ----
 # 1) STEP 1 : build the model.
-# 2) STEP 2 : create a link to install/bin from the working directory.
-# 3) STEP 3 : bin/make_bcs
+# 2) STEP 2 : ./make_bcs
```
gmao-rreichle, line 10: Shouldn't this 2 read something like:
#                             ---- USAGE ----
# STEP 1: Build the model.
# STEP 2: Run [install directory]/bin/make_bcs

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
```diff
@@ -27,8 +26,8 @@ if ( $1 == noland ) set SKIPLAND = YES
 #                     Set GEOS5/ESMADIR Variables
 #######################################################################
 
-set MAKE_BCS = `echo $cwd`/bin/make_bcs
-set  ESMADIR = `echo $cwd`/bin/ 
+set MAKE_BCS = `echo $cwd`/make_bcs
```
gmao-rreichle, line 29: Not sure what these environment variables do.  Are they needed so that the other executables in the ./bin directory can be found?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
```diff
@@ -59,15 +58,80 @@ set WHITE   = "\033[37m"
...
 
+if( -e $HOME/.make_bcs) then
+     set make_bcsinfo = `cat $HOME/.make_bcs`
+     set C1 = $make_bcsinfo[1]
+     set C2 = $make_bcsinfo[2]
+else
+     set C1 = $RED
+     set C2 = $BLUE
+endif
+     set CR = $RESET
```
gmao-rreichle, line 74: I had assumed that C1, C2 and CR are simply to show text for the interactive questions in different colors.  Why do we need a "hidden" (dot) `.make_bcs` directory?  Do the colors not work without this?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
```diff
@@ -59,15 +58,80 @@ set WHITE   = "\033[37m"
...
+#######################################################################
+#                     Build Directory Locations
+#######################################################################
+
+# Set Current Working Path to make_bcs
+# -------------------------------------
+setenv ARCH `uname -s`
+if ($ARCH == Darwin) then
```
gmao-rreichle, line 82: Is the architecture dependency really needed?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
```diff
@@ -59,15 +58,80 @@ set WHITE   = "\033[37m"
...
+   set FINDPATH = realpath
+else
+   set FINDPATH = 'readlink -f'
+endif
+set MAKE_BCS = `$FINDPATH $0`
```
gmao-rreichle, line 87: Why redefine MAKE_BCS here?  What about the definition of MAKE_BCS in line 29?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
```diff
@@ -59,15 +58,80 @@ set WHITE   = "\033[37m"
...
+if (! -x  GEOSgcm.x && ! -x  GEOSldas.x) then
+   echo "You are trying to run $0 from the wrong directory"
+   echo "Please run from the bin/ directory in your installation"
+   exit 1
+endif
+#######################################################################
+#                 Create Paths for EXP Directorie
+#######################################################################
+echo
+echo "Enter the ${C1}Boundary Conditions Directory Name${CR}:"
```
gmao-rreichle, line 99: Why do we need a "name" for the bcs directory?  Don't we just need a path so that make_bcs can create an appropriately-named sub-directory for whatever bcs the user chooses to create?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
```diff
@@ -59,15 +58,80 @@ set WHITE   = "\033[37m"
...
+set EXPID = $<
+
+# EXP Directory (for Output Data )
+# --------------------------------------------
+    set check =  FALSE
+while( $check == FALSE )
+  if( -e $HOME/.EXPDIRroot ) then
+         set EXPDIRroot = `cat $HOME/.EXPDIRroot`
```
gmao-rreichle, line 107: Same as above: Why do we need a dot file?  If we do want this functionality, why not add EXPDIRroot into .make_bcs?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
gmao-rreichle: I don't understand why this shouldn't work.  Note that we're checking the user-selected $dummy.  If $ dummy is not one of the pre-defined options spelled out in the if statement, the else statement will abort make_bcs.  At least that's the intent.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
mathomp4: For what this is doing, yes. macOS (being BSD-based) does not supply `readlink` but rather `realpath`. Now, will `make_bcs` ever run on macs...who knows?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
mathomp4: This is an odd way to set these things. For the GCM we actually use CMake to "insert" in the correct installation path where needed. Since CMake knows where things are installed, you can just have CMake add the right path and then it's "hard coded" as it were. 

The way this is set you probably can never run:
```
/path/to/install/bin/make_bcs
```
but rather must run as:
```
cd /path/to/install
./bin/make_bcs
```
I guess since it assumes it will only ever be run from the install dir directly.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
biljanaorescanin: After today's meeting we decided this will go back to mask that is used for EASE on develop. ( GEOS5_10arcsec_mask.nc ) that same mask is used in ltakacs/bcs/Icarus-NLv3/Icarus-NLv3_EASE/

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
biljanaorescanin: This was tested today and is now working.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
biljanaorescanin: This is working now.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
gmao-rreichle: Based on our discussion, there is no difference in the *.til (lower-case file extension) and all other files for bcs on the EASEv[x] grid.  
The mask does seem to impact the *.TIL (upper-case file extension) that seems to be created in post-processing for the coupled model, where some land and ocean tiles are changed to facilitate the dynamic ocean modeling (opening of straits etc).
Since the tentative change here is within an EASE grid if block, there is no impact on any bcs for the coupled atm-ocean model. 
I'm resolving the conversation.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/create_README.csh
```diff
@@ -41,7 +41,6 @@ set mygrid=`echo $myusage | cut -d'g' -f2 | cut -d '-' -f1`
 cvs status src/mkCatchParam.F90 > clsm/TagInfo
 #echo GMU_OCT10_SM >> clsm/TagInfo
 echo `head -7 clsm/TagInfo | tail -1` | cut -d':' -f2 | cut -d'(' -f1 > clsm/TagName
-set MYTAG=`head -1 clsm/TagName | tail -1`
```
gmao-rreichle, line 44: It make sense to remove this env variable that is meant to hold the CVS tag.  But why not also delete Lines 43 and 45?  Aren't those also just trying to process CVS tag info?

biljanaorescanin: Yes, I missed those. I can clean it up

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
```diff
@@ -2,19 +2,21 @@
...
+#  STEP 1 : Build the model (GCM or GEOSldas).
+#  STEP 2 : cd [install-path]/bin
+#  STEP 3 : source g5_modules
+#  STEP 4 : ./make_bcs 
+#               Answer the following interactive questions:
+#               a) Enter the Atmospheric Horizontal Resolution Code(s) to build.
+#               b) Select the Land BCs version to use. 
+#               c) Enter the Ocean Resolution Code(s) to build.
+#                  (Note this selection is not relevant to land-only EASE grids)
+#               d) Enter sponsor code.
```
gmao-rreichle, line 19: @biljanaorescanin:  I tentatively revised the usage/instructions.  Two things:
1. Do we need to source g5_modules before running make_bcs?  I added it just in case. 
2. One more idea re. the specification of the output path (and perhaps also EXPID): We could make the path and EXPID command line arguments of make_bcs.  If specified at the command line, they would overwrite the defaults and then we don't have to query for them.  This may be better than asking users to type (or cut-and-paste) a path in response to an interactive question.  Just an idea.

gmao-rreichle:
@biljanaorescanin: I added three commits:
Commits https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/517/commits/cfa88ac73253815d825fb1dfbf424b3f153c6dd9 and https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/517/commits/5c772388a6abbb63e9404bb8da21fafd2c099ec0 are very minor and should be self-explanatory.
Commit https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/517/commits/d0fc320ba7cf80b2a9cb9527c640be21f3522703 is another attempt at cleaning up the status messages in mkCatchParam.F90.  I found a few more issues with incomplete file checks and status reporting that I tried to fix, in addition to the clean-up that I had in mind.  The code compiled for me, but  I did not run it.  Maybe give it a try to see what we get and what may still need tweaking.

biljanaorescanin:
@gmao-rreichle this last commit is cleanup we talked about on Friday.

sanAkel:
I have following question:
is `topo_DYN_ave_90x540.data` generated somehow by the `make_bcs`?

If **yes** and as it seems this PR is intended to enhance documentation, then I would like to note that the `develop` version **does not** produce this file: `topo_DYN_ave_90x540.data` for me! Did it for @biljanaorescanin, @mathomp4, or whoever has _tested_? Isn't that a BUG??

biljanaorescanin:
@sanAkel no make_bcs package doesn't generate this file: topo_DYN_ave_90x540.data
These files are generated somewhere else and then linked to bc's. You can see in Larry's area how it is linked :
/discover/nobackup/ltakacs/bcs/Icarus-NLv3
NCAR_TOPO_GMTED -> ../Icarus-NL/NCAR_TOPO_GMTED

gmao-rreichle:
The obvious next questions are:

1. Which package creates the AGCM's topo_* files?
2. Why is the generation of the topo_* files is separate from make_bcs?
3. Shouldn't the two sets of files be created as part of the same package and (ideally) use the same topo source data?

In the past, I pointed out potential inconsistencies between the topo_* files and the topography data underlying the Catchment parameter files, but I'm not aware of any follow-up. 

Something for the modeling group to keep in mind.

sanAkel:
> The obvious next questions are:
> 
> 1. Which package creates the AGCM's topo_* files?
> 2. Why is the generation of the topo_* files is separate from make_bcs?
> 3. Shouldn't the two sets of files be created as part of the same package and (ideally) use the same topo source data?
> 
> In the past, I pointed out potential inconsistencies between the topo_* files and the topography data underlying the Catchment parameter files, but I'm not aware of any follow-up.
> 
> Something for the modeling group to keep in mind.

@gmao-rreichle Thanks for noting ⤴️ I didn't want to clobber the context of _this_ PR with seemingly _unrelated_ issues. Obviously it seems there are numerous issues. Should we start a new issue for as many of these as we can think of? Or _revive_ the discussion we already [started sometime ago](https://github.com/GEOS-ESM/GEOSgcm_App/discussions/193)?

gmao-rreichle:
> @gmao-rreichle this last commit is cleanup we talked about on Friday.

Thanks, @biljanaorescanin.  I revised the comments and messages a bit more for clarity.  I also made the indents prettier for the status messages from mkCatchParam.F90 (white-space changes only).  Here's the commit: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/517/commits/c1875688a0c5d5badf8397f5d4b42b962fc300ee

Please review and let me know if you have any edits.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
gmao-rreichle: @biljanaorescanin, 

I still don't understand why MAKE_BCS and ESMADIR are set here. 

It looks like MAKE_BCS is simply set so it can be echoed a few lines down.  Given that the instruction tell the user to cd to the install/bin directory and run make_bcs, this echo isn't very informative and is not needed.
ESMADIR is apparently used only to source g5_modules (after setting ESMADIR to the current working directory).  I don't see it being used otherwise.
We can probably remove all this and simply add "source g5_modules" as a new STEP 3 in the USAGE instructions.  Then the instructions for running make_bcs would be exactly the same as the instructions for running ldas_setup.

There are also 5 instances of "setenv ESMADIR $ESMADIR" in the BCJOB scripts that are created by make_bcs.  I don't see anything in these job scripts that obviously requires ESMADIR, so chances are these lines are also obsolete. 

Please remove the MAKE_BCS and ESMADIR pieces and try running make_bcs again (after sourcing g5_modules).  With any luck, everything will still work.  If not, we'll maybe learn why MAKE_BCS and/or ESMADIR are needed.

gmao-rreichle:
@biljanaorescanin, 
I simplified the logic for determining the BCS output directory in make_bcs: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/517/commits/348f9bfd54d740560f921c7cead1888a4bf24741
I hope I got my "setenv" and "set" commands right and didn't break the script.  I'm confident the overall logic is ok.  Please double-check.

mathomp4:
I'm going to cc in @atrayano and @bena-nasa as they might know who makes those topo files.

biljanaorescanin:
@gmao-rreichle was this what you had in mind for help message? I left usage text the same just moved it to be on screen if added "-h" .

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
```diff
@@ -59,143 +85,222 @@ set WHITE   = "\033[37m"
...
+   echo "Enter desired BCS ${C1}output directory${CR} (incl. full path)"
+   echo "   or press ENTER to use the default:"
+   echo "   [${C2}${EXPDIR}${CR}]"
+   set NEWEXPDIR = $<
+
+   if( $NEWEXPDIR != '' ) setenv EXPDIR $NEWEXPDIR
+
+   mkdir -p $EXPDIR
+
+   if($EXPDIR != '') then
```
gmao-rreichle, line 119: Needs work: this checks if the directory exists, but it should check if the directory is non-empty.

gmao-rreichle: Perhaps we can check for the specific (resolution) sub-directories of EXPDIR, rather than just check if EXPDIR is empty?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
biljanaorescanin: I've just pushed commit to resolve this, since it is getting lost in the comments.  We now source g5_modules as you suggested and don't need any of this.

biljanaorescanin:
@gmao-rreichle I've added revision so we can address the potential for specifying more than one set of bcs. It will just add to directory name all resolutions separated by "_" . I've removed left over module sourcing. 
I've changed one line in latlon option that I think it was wrong but we can talk about it after you take a look.

gmao-rreichle:
@biljanaorescanin:  
I finished the changes that I had on my list by adding two commits:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/3541a12a1246aea801b454d903359548a7bf40ad contains the nitty-gritty changes that fix several problems with bad user inputs in the interactive questions and the logic of processing the information.  There are also some revisions to the help text.
https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/d6179f9a1528a5c7dcb59d78342c3a34ee6e0ce6 changes the order of the interactive questions into something more logical and adds the BCs version to the default output path.  I put these changes into a separate commit so the xxdiff of the changes in the first commit still makes some sense. 

There are two things left to do:
1. If you haven't already done so, please proofread the bcs README file generated by `create_README.csh`.  We need to make sure that the README file correctly reflects how the bcs package works and what it generates. 
2. Test the branch to make sure everything still works as intended.

gmao-rreichle (approved):

sdrabenh (approved):

PR #518, opened 2022-01-14
removed redundant code: DQS calc, QC constraint, RA calc

gmao-rreichle:
This PR cleans up presumably redundant code (DQS calc, QC constraint, RA calc) that occurred twice in each Catch[CN] GridComp F90 file.  The QC constraint was also removed from catchment.F90 (where QC is known as "qa").
 
The PR was prompted by @amolod's observation that the code in question should not have been placed within an `if (LAND_FIX)` block.  After further investigation, it was discovered that if block code in question already occurs verbatim earlier in the subroutine and thus looks to be redundant.  

@amolod reports that Mike Manyin successfully ran a 0-diff test for the S2S model after removing the if (LAND_FIX) condition.

In the CatchCN GridComp F90 files, the two identical blocks of code are separated by ~1,000 lines.  An inspection of these lines suggests that the code blocks are redundant, but it is difficult to say from just going through the source code.
 
@biljanaorescanin: When you get a chance, please run the standard 0-diff tests for both the AGCM and GEOSldas.

cc: @rdkoster @wmputman @sdrabenh

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM40_GridComp/GEOS_CatchCNCLM40GridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM40_GridComp/GEOS_CatchCNCLM40GridComp.F90
@@ -6251,13 +6251,17 @@ subroutine RUN2 ( GC, IMPORT, EXPORT, CLOCK, RC )
            end do
         end if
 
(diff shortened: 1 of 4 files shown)


biljanaorescanin:
@gmao-rreichle  I've tested this PR for:
1. GEOSldas nightly tests and all are zero diff 
2. AGCM ( catchment) one day run was zero diff , wasn't sure was there need to test catchmentCN options as well?

gmao-rreichle:
Thanks, @biljanaorescanin.

> 1. GEOSldas nightly tests and all are zero diff.
> 2. AGCM ( catchment) one day run was zero diff , wasn't sure was there need to test catchmentCN options as well?

The LDAS tests include an offline CatchCN simulation (which passed).  I'm not aware of an AGCM test that runs CatchCN.  

As far as I can tell, the PR is ready for review by @sdrabenh.

cc: @amolod

biljanaorescanin (approved):

PR #521, opened 2022-01-19
Simple fix to prevent negative tauaer in RRTMG LW

dr0cloud:
RRTMG LW does not scatter and so it must extract the absorption aerosol optical depth. This should be taua_ext - taua_scat, which in the parlance of the un-normalized radiation variables in the RRTMG LW section is TAUA - SSAA.
Unfortunately this difference is sometimes a small negative value. **This is not the fault of the Radiation GC, since it gets these two variables directly by callback to the aerosol system. It should really be fixed on the aerosol side, but as a simple mitigation for now, we make the fix max(TAUA - SSAA, 0.).**


dr0cloud:
Its technically zero-diff label because any negative tauaer would have caused a crash so not run at all!

sdrabenh (approved):

Issue #469, opened 2021-10-14
Land Group Peat and Irrigation PR's


biljanaorescanin:
This issue collects links to draft PRs with land model developments that were closed without merge.

The closed PRs will be needed again when development is restarted.

1) Peat: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/417 
2) Irrigation: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/420
3) Boundary conditions for peat, irrigation, river routing and interannually varying vegetation: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/413

An even older set of PRs bundled several land model updates: 
https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/151
https://github.com/GEOS-ESM/GEOSgcm_App/pull/68

@gmao-rreichle @rdkoster @sdrabenh @gmao-jkolassa

biljanaorescanin:
Adding one more PR to this issue, this one is from GEOSgcm_App labeled [science improvements PR#68 ](https://github.com/GEOS-ESM/GEOSgcm_App/pull/68)

NOTE: This PR is not related to Peat or Irrigation updates.

gmao-rreichle:
Reopening this issue because items 2 for irrigation and 3 for bcs have not yet been addressed.
Item 1 for peat and peat-related bcs changes in item 3 have been addressed with PR #524.

PR #417, opened 2021-03-17
Feature/sm peat clm3

smahanam:
This draft is for an internal review by land group. 
Feel free to edit while the science tests are underway.

PR #524, opened 2022-01-25
Feature/borescan merge sm peat

biljanaorescanin:
Before Sarith left the land group he was working on several PR's. 
This PR is reproducing results he has shown for the update related to peatlands. 

Related issue and PR:
PR: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/417
Issue: https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/469

I'm able to get zero diff comparison for both history and restart files with respect to my "offline" run when:
1.  In "lfs" collection I remove SWnet variable ( we don't have SWnet in our collection anymore)
2.  For "lnd" collection I have to run experiment with no nbits.  

@gmao-rreichle @rdkoster 

---

UPDATED 9-Feb-2022 by @gmao-rreichle:  
The above 0-diff result is vs. Sarith's original PEATCLSM simulation:
/discover/nobackup/borescan/par/NL5_reproduce_NLv4p_M01_M07_nbits/

---
UPDATED 11 Feb 2022 by @gmao-rreichle:

**Summary of Changes**:
- Added revised hydrology for peatlands (PEATCLSM): New PEATCLSM hydrology is used when the model is run with "NLv5" boundary conditions, which identify peat tiles through hybridization of PEATMAP and HWSD ancillary data and assign a porosity of 0.93.  PEATCLSM hydrology kicks in for tiles with porosity >=0.9.
- New Catchment export variables: WATERTABLED, FSWCHANGE
- New Catchment diagnostic functions: catch_calc_zbar(), catch_calc_watertabled()
- Cleanup 

---

**Test Results** (at final commit [90f7439](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/524/commits/90f743995171c7039b680df24ea385f8618b76f1)):
- 0-diff vs. AGCM develop
- 0-diff vs. GEOSldas develop for all Intel and GNU tests (as expected, tests with “aggressive” builds have roundoff differences)
- 0-diff vs. Sarith's original PEATCLSM experiment **after reversing** the following fix that was needed to get 0-diff vs. GEOSldas develop: https://github.com/GEOS-ESM/GEOSgcm_GridComp/compare/f10f98e..59e3121
- Successfully reproduces bcs (verified for M36/NLv3 and M36/NLv5)

---

**Merge Info**:  The present PR already includes and should be merged after #536.

---

**Still To Do**:  See issue #539 for future PEATCLSM work.


biljanaorescanin:
Ah it can also be just my mistake. I will double check Sarith's original code but I think it was removed there, so I removed them here since I was coping what was done on that branch, not sure now. But as you said I only tested catchment not catchmentCN.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/catchment.F90
```diff
@@ -1765,28 +1784,91 @@ SUBROUTINE RZDRAIN (                                                     &
...
+             ! RUNSRF(N)=RUNSRF(N) + (CATDEF_PEAT_THRESHOLD - CATDEF(N))
+             ! runoff from AR1 for zbar>0
+             ! RZFLW_AR1 = RZFLW - RZFLW_CATDEF + (CATDEF_PEAT_THRESHOLD - CATDEF(N))
+             ! AR1=0.5 at zbar=0
+             ! SYsurface=0.5 at zbar=0
+             ! RUNSRF(N) = RUNSRF(N) + amax1(0.0, RZFLW_AR1 - 0.5*1000.*ZBAR1)
+             ! 
+             ! revised (rdk, 1/04/2021): take excess water from both 
+             ! soil and free standing water, the latter assumed to cover area AR1=0.5
+             RUNSRF(N) = RUNSRF(N) + CATDEF_PEAT_THRESHOLD-CATDEF(N) + 0.5*1000.*(-ZBAR1)
```
gmao-rreichle, line 1862: @rdkoster:  Are the units in this equation for RUNSRF ok?  If RUNSRF is a flux and CATDEF is a storage, I would expect a division or multiplication by the time step somewhere.   I might well be wrong, but thought it's best to double-check. 
cc: @biljanaorescanin

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/catchment.F90
```diff
@@ -1765,28 +1784,91 @@ SUBROUTINE RZDRAIN (                                                     &
...
+       ENDIF
 
-        IF(CATDEF(N) .LT. 0.) THEN
+       IF (POROS(N) >= PEATCLSM_POROS_THRESHOLD) THEN
+          ! MB: CATDEF Threshold at zbar=0
+          ! water table not allowed to rise higher (numerically instable) 
+          ! zbar<0 only occurred due to extreme infiltration rates
+          ! (noticed this only snow melt events, very few locations and times)
+          ! (--> NOTE: PEATCLSM has no Hortonian runoff for zbar > 0)            
+          CATDEF_PEAT_THRESHOLD = ((BF2(N))**2.0-1.e-20)*BF1(N)
```
gmao-rreichle, line 1851: @rdkoster:  Is there a reason why 1.e-20 is subtracted here?   It doesn't seem to be a protection against division by zero or ensuring non-negative values is necessary.  Perhaps I'm missing something.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/catchment.F90
```diff
@@ -2358,6 +2441,15 @@ SUBROUTINE energy4 (                                                     &
       DEDEA(CHNO)  = DEDQA(CHNO) * EPSILON / PSUR(CHNO)
       DHSDEA(CHNO) = DHSDQA(CHNO) * EPSILON / PSUR(CHNO)
 
+      IF (POROS(CHNO) < PEATCLSM_POROS_THRESHOLD) THEN
```
gmao-rreichle, line 2444: @rdkoster:  Do we need the equivalent of this difference in SWSRF4 between mineral and peat soil also in catchmentCN.F90?  In CatchCN, the "energy" subroutines are done differently, and I can't easily figure out how this if block would translate into CatchCN.  All of the other "if peat" block in catchment.F90 appear almost verbatim in catchmentCN.F90, so I thought it's best to double-check. 
cc: @biljanaorescanin @gmao-jkolassa

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/Scale_Catch.F90
```diff
@@ -393,6 +366,15 @@ end subroutine calc_soil_moist
...
   endif
 
+  ! PEAT CLSM - ensure low CATDEF on peat tiles
+  ! -------------------------------------------
+
+  where (catch(sca)%poros >= PEATCLSM_POROS_THRESHOLD)
+     catch(sca)%catdef = 100.
+     catch(sca)%rzexc  = 0.
+     catch(sca)%srfexc = 0.
+  end where
```
gmao-rreichle, line 376: @rdkoster, this block sets catdef=100. for peat tiles when restarts are scaled & regridded.  Two questions:
1. A value of 100. to initialize catdef in peat seems large to me (ie, the initialization seems too dry to me).  Initializing too wet would probably spin up more quickly. 
2. The value of 100. is used in the target (output) restart for peat regardless of what the source restart had.  If the source restart also had peat, wouldn't we want to keep the catdef close to what was in the source restart, perhaps adjusted for porosity differences (0.8 as in HWSD vs. 0.93 as in PEATMAP).

gmao-rreichle:
@biljanaorescanin, @rdkoster: 
For the record, it turned out that Sarith's branch (feature/SM_peatCLM3), which was Biljana's starting point for this PR, was **not** 0-diff against "develop".  I had not expected starting off with something that was already violating 0-diff and spent most of my weekend trying to understand why 0-diff testing failed after adding my revisions.  I finally fixed the problem as follows: 
https://github.com/GEOS-ESM/GEOSgcm_GridComp/compare/f10f98e..59e3121

FYI, in the associated GEOSldas branch (feature/borescan_merge_sm_peat), I removed the new variables WATERTABLED and FSWCHANGE from the default HISTORY file (GEOSldas_HIST.rc) for now.  This avoids "trivial" test failures when the new variables show up in the test output and makes it easier to interpret GEOSldas test results while we're still working on the peat branches.  We can add these variables back and rebaseline the tests later.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/Shared/catchmentCN.F90
```diff
@@ -2462,11 +2558,11 @@ subroutine gndtmp_cn(poros, zbar, ht, xfice, tp, FICE)
 
 ! calculate the boundaries, based on the layer thicknesses(DZGT)
 
-      zb(1)=-0.05    ! Bottom of surface layer, which is handled outside
-                    ! this routine.
+      zb(1)=-DZTC    ! Bottom of surface layer, which is handled outside
```
gmao-rreichle, line 2561: @rdkoster, @gmao-jkolassa:
I assume that the 0.05 number here is what is DZTC in Catchment.  In CatchmentCN, however, I thought that there are two kinds of "surface" temperatures:  "TC" (canopy) and "TG" (ground/soil). 
Note that DZTC is also used in subroutine FLUXES() in catchmentCN.F90.
In CatchmentCN, it's probably inconsistent to keep calling this constant DZTC.  Perhaps it should be renamed DZTG in CatchmentCN.  Does that make sense?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/Shared/lsm_routines.F90
```diff
@@ -1861,8 +2065,15 @@ subroutine gndtmp(dts,phi,zbar,thetaf,fh21,ht,xfice,tp, FICE)
...
-         enddo
-      xfice=xfice/((N_GT+1)-lstart)
+      enddo
+
+      IF (phi < PEATCLSM_POROS_THRESHOLD) THEN
+         xfice=xfice/((N_GT+1)-lstart)
+      ELSE
+         !PEAT
+         !MB: only first layer for total runoff reduction
+        xfice=AMIN1(1.0,fice(1))
```
gmao-rreichle, line 2075: @rdkoster, @gmao-jkolassa:
At the end of subroutine gndtmp(), xfice is calculated differently in PEATCLSM.
In subroutine gndtmp_cn() in catchmentCN.F90, xfice is also calculated at the very end, but there is no "if PEATCLSM" block in gndtmp_cn().  
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/e44b03ba48a35a6f9b036783f15b0d89165d97e0/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/Shared/catchmentCN.F90#L2618
Does the "if PEATCLSM" block need to be added in gndtmp_cn() as well?
In any case, what is the purpose of gndtmp_cn() as opposed to gndtmp()??  **Both** are used in CatchmentCN.  At the very least, a few comments in gndtmp_cn() about its purpose and relationship to gndtmp() are needed.  Ideally, the gndtmp subroutines would be restructured to avoid the code duplication and make the purpose of each variant more obvious.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
```diff
@@ -5628,6 +5651,10 @@ subroutine Driver ( RC )
         if(associated(RMELTBC002)) RMELTBC002 = RMELT(:,7) 
         if(associated(RMELTOC001)) RMELTOC001 = RMELT(:,8) 
         if(associated(RMELTOC002)) RMELTOC002 = RMELT(:,9) 
+        if(associated(FSWCHANGE )) FSWCHANGE  = FSW_CHANGE
+        if(associated(WATERTABLED )) then
+           WATERTABLED = MIN(SQRT(1.e-15 + CATDEF/BF1) - BF2, CDCR2/(1.-WPWET)/POROS/1000.)
```
gmao-rreichle, line 5656: @rdkoster, @biljanaorescanin:
The zbar calculation within the MIN() should be replaced with the catch_calc_zbar() function.
This will be a **non-zero-diff change for PEATCLSM** because of the inconsistent use of "1.e-15" and "1.e-20" across PEATCLSM. 
The same applies to the corresponding lines in GEOS_CatchCNCLM40GridComp.F90 and GEOS_CatchCNCLM45GridComp.F90.
Note that a vectorized version of catch_calc_zbar() is needed here.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/Shared/lsm_routines.F90
```diff
@@ -275,82 +278,197 @@ SUBROUTINE SRUNOFF (                                                  &
...
+               !  srun0 = srun0 + RZEQ(N)+RZEXC(N)-VGWMAX(N)
+               !  RZEXC(N)=VGWMAX(N)-RZEQ(N)
+               !  ENDIF
+               !(Commented out following lines to retain water balance -- rdk, 9/18/20)
+               !if (srun0 .gt. ptotal) then
+               !   srun0=ptotal
+               !   endif
+               RUNSRF(N)=RUNSRF(N)+srun0
+               QIN=PTOTAL-srun0
+               !SRFEXC(N)=amin1(SRFEXC(N)+QIN,srfmx(n))
```
gmao-rreichle, line 369: @rdkoster: Is this commented out intentionally?  It's different for non-peat tiles, where QIN is added to SRFEXC.  It's probably intentional, but I thought I should double-check.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/Shared/lsm_routines.F90
gmao-jkolassa: @gmao-rreichle The gndtmp_cn subroutine was implemented by Greg and there is not much information in his notes as to why. Maybe Randy is going to be able to shed a bit more light on this. What I can tell from the code is that gndtmp_cn is called by the GEOScatchCN_GridComp (catchmentCN only calls the original gndtmp) in order to compute the soil temperatures (tp) and the total column ice factor, since these are needed for the call to CN_Driver that occurs before the call to catchmentCN. gndtmp_cn is a simplified version of gndtmp that leaves out many of the heatflux calculations present in the original routine. I _think_ that there are two reasons for introducing this new and simpler routine: (1) not all of the outputs computed by the original gndtmp are needed for the call to CN_Driver, and (2) not all of the inputs needed to call the original gndtmp subroutine are available in the GridComp (several inputs are only defined in catchmentCN). Like I said, this assessment is just based on looking at the code. Randy may shed more light on the actual rationale for creating this routine. 

I can add some comments in the code to explain this. But is this something you want added as part of the peat PR?

As to whether the "if PEATCLSM" block needs to be added, I guess the answer is 'yes' if there are plans to run the peat model with CatchmentCN.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/Shared/lsm_routines.F90
rdkoster: Yes, this was intentional; for peat, SRFEXC is updated in the earlier IF blocks.  Probably for clarity the QIN=PTOTAL-srun0 line should be commented out as well (or both these lines could be removed), though it's not hurting anything as is.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
rdkoster: Sounds fine.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/Shared/lsm_routines.F90
rdkoster: Yes, the PEATCLSM IF block needs to be added to CatchmentCN; there shouldn't be any difference between Catchment and CatchCN in this regard.  I can't shed any new light on why Greg put in the gndtmp_cn routine (I don't think he told me.)  Jana's assessment makes sense.  I'll hunt down that shorter routine and ensure that the physics subset looks right.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/Shared/catchmentCN.F90
rdkoster: My only concern about the renaming is the possibility that Catchment and CatchmentCN may need to share a subroutine that requires the DZTC parameter.  In fact, they do -- didn't Jana say that both Catchment and CatchmentCN use gndtmp.f?  You wouldn't want to maintain two otherwise identical subroutines just to use different names for the variables, right?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/Scale_Catch.F90
rdkoster: 1. It wouldn't hurt to reduce the initial catdef value for peat.  My instinct tells me not to set it below 20 or 30, though.  (For most of the non-wetland world, a value of 100 is actually reasonably wet.)
2. Given the utterly different hydrology that Michel has formulated (e.g., with standing water that doesn't just run off), I wouldn't want to mine a source restart for a catdef derived with our original (non-peat) hydrology.  I say we just initialize wet (as in 1 above) and require some spin-up.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/catchment.F90
rdkoster: Yes, this is needed in CatchmentCN.F90.  The following code could be added at line 758.  (The value of SWSRF4 coming into this piece of code is the value for non-peat; this code would overwrite that value if we're at a peat tile.)
      IF (POROS(CHNO) > PEATCLSM_POROS_THRESHOLD) THEN
         ! PEAT
         ! MB: For ET calculation, AR4 surface wetness is set to WPWET
         SWSRF4(CHNO) = WPWET(CHNO)
      ENDIF

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/catchment.F90
gmao-rreichle: @rdkoster confirmed that 1.e-20 can be removed from this line.  To be implemented later, along with other non-zero-diff changes in PEATCLSM.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/Scale_Catch.F90
gmao-rreichle: Thanks, @rdkoster.   Re. item 1: I'll change the initial catdef value to 25.
Re. item 2: Note that I suggested mining a source restart for catdef **only if the source restart also had peat in the associated tile**.   The current code ALWAYS set the initial catdef value for peat to 100. (to be changed to 25.).  If, in the future, we have a spun-up restart file with PEATCLSM tiles in it, the perfectly good catdef values for the PEATCLSM tiles in the source restart file would be ignored altogether.  This doesn't make sense to me.   I think this needs more work.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/catchment.F90
rdkoster: You've found a problem upon onto a longstanding, ugly bug in the original Catchment model.  Fortunately, based on a bit of analysis with MERRA-2 dailies, it appears to manifest itself very rarely, if at all.  (I can't be sure based on the data I currently have access to.)  In a zero-diff sense, fixing it should only be problematic in the coupled system, wherein the ocean prognostics are affected by land-produced runoff -- again, if this bug manifests itself at all.  It would be highly useful to perform, e.g., a non-peat M36 Nature Run with the corrected Catchment and compare the computed runoffs against the control.
As you indicate, the corrected line should be 
             RUNSRF(N) = RUNSRF(N) + (CATDEF_PEAT_THRESHOLD-CATDEF(N) + 0.5*1000.*(-ZBAR1))/DTSTEP
Also, line 1875 should be corrected to
          RUNSRF(N)=RUNSRF(N)-CATDEF(N)/DTSTEP
This latter change corrects the problem that has existed for a long time in Catchment.  I'm keen on finding out the extent to which this problem has affected our runoff diagnostics.

rdkoster: Note that I expect minimal past damage from this bug given that in the early days, we tested Catchment (with this bug) extensively on a global scale and always achieved an accurate water balance.  This code was originally added to protect, perhaps unnecessarily, against something crazy happening.  It would be good to establish whether that crazy behavior could actually happen...

gmao-rreichle: @rdkoster: I went ahead and 0-diff tested the bug fix in the "develop" version of catchment.F90 (non-PEATCLSM) using the GEOSldas suite of tests.  It turned out to be 0-diff.  Arguably, the tests are just 1-day runs in a couple of different configurations and don't sample the full range of possibilities that we would encounter in, say, a 40-year reanalysis experiment.  Chances are that we also get 0-diff in the GCM tests (which are much coarser resolution and probably even shorter than the GEOSldas tests).  To be sure, @biljanaorescanin should test this for the GCM.  (Biljana, my GEOSgcm_GridComp branch is `bugfix/rreichle/rzdrain_runsrf`.)  
Assuming we also get 0-diff in the GCM tests, I suggest we just implement the bug fix as part of the PEATCLSM branch.  The GCM tests in any case don't consider changes in diagnostic outputs (such as runoff) a violation of 0-diff.  (For SMAP L4, I would try to avoid such changes if at all possible, but since GEOSldas tested 0-diff, it's ok with me.)

biljanaorescanin: I've did one day AGCM test and it is 0-diff.

gmao-rreichle: @rdkoster:  I implemented the bug fix in catchment.F90 and also in catchmentCN.F90.  I also added a missing "rdkoster" revision in catchmentCN.F90.
Please double-check: https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/99f1dc9

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/Scale_Catch.F90
gmao-rreichle: Initial catdef value changed from 100. to 25.  https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/99f1dc9
Still needs work pulling initial catdef from source restart if tile is peat in both the source and target restarts.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/Shared/catchmentCN.F90
gmao-rreichle: Renamed DZTC --> DZTSURF https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/b3aa6bd1820d1b8534bcc27b598bbaec47519e97

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/Shared/lsm_routines.F90
gmao-rreichle: Fixed missing if-block for CatchCN.   Clean up gndtmp_cn() still needed but I am not volunteering.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
gmao-rreichle: Provided vectorized version of catch_calc_zbar()  https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/30ef2dc910b3850ff56e0dcfbb28b4daf2f18291
Still needs to be turned on when non-zero-diff changes for PEATCLSM are merged.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/Shared/lsm_routines.F90
gmao-rreichle: @rdkoster:  To me it looks like commenting out `QIN=PTOTAL-srun0` would be a non-zero-diff change for PEATCLSM because QIN is used to compute QINFIL just before the subroutine ends.  In fact, if the QIN line is commented out, QIN is not defined when QINFIL is computed at the end.  If you agree, please "Resolve conversation" for this comment on GitHub.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/catchment.F90
gmao-rreichle: @rdkoster:  I hope the following commit fixes this: https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/666d16073d7d6bc041a88940ee3ba7e05fe5f955.    Please double-check.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/Shared/lsm_routines.F90
rdkoster: You're right; QIN=PTOTAL-srun0 should not  be commented out.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/catchment.F90
rdkoster: Looks good.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/Shared/lsm_routines.F90
gmao-rreichle: See new issue #530 re. cleanup of gndtmp_cn() that is not related to PEATCLSM.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/catchment.F90
```diff
@@ -1283,6 +1303,12 @@ SUBROUTINE CATCHMENT (                                                   &
         WCHANGE(N) = (WTOT(N)-WTOT_ORIG(N))/DTSTEP
         ECHANGE(N) = (ENTOT(N)-ENTOT_ORIG(N))/DTSTEP
 
+        !FSW_CHANGE IS THE CHANGE IN THE FREE-STANDING WATER, RELEVANT FOR PEATLAND ONLY
+        FSW_CHANGE(N) = 0.
+        IF(POROS(N) >= PEATCLSM_POROS_THRESHOLD) THEN
+           pr = trainc(n)+trainl(n)+tsnow(n)+tice(n)+tfrzr(n)
+           FSW_CHANGE(N) = PR - EVAP(N) - RUNOFF(N) - WCHANGE(N)
```
gmao-rreichle, line 1310: @rdkoster:  Is this now the complete land water balance of the AGCM:  
`WCHANGE = PR - EVAP - RUNOFF - FSWCHANGE + SPWATR ` ??
I wonder where could be documented.  It's obviously too early for a file specs doc, but it'd be good to make a note somewhere.

gmao-rreichle:
@biljanaorescanin, @rdkoster: 
FYI, I merged the RUNSRF bug fix (branch: bugfix/rreichle/Catchment_RUNSRF_units) into the PEATCLSM branch in anticipation of Scott merging PR #535 into develop.  This required manual resolution of conflicts.  See [commit 6904c78](https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/6904c78bf90c5279945ce0ee2c287bb6b578a13e).
If there are more conflicts to resolve after Scott's merger, I will take care of them.

gmao-rreichle:
Update on test results at latest [commit 3c4ee1b](https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/3c4ee1b2fd9b86b2c0701a2cdee869d917c7e298):

- 0-diff vs. GEOSldas develop in the following Intel tests: conus, global, globalcs, globalcnclm4, debugconus

- 0-diff vs. Sarith's original PEATCLSM experiment **after reversing** the following fix that was needed to get 0-diff vs. GEOSldas develop:  https://github.com/GEOS-ESM/GEOSgcm_GridComp/compare/f10f98e..59e3121
  
Sarith's original experiment located here:
/discover/nobackup/smahanam/MichelBechtold/EXPS/NRv8.3_M36_pCLSM

Sarith's run was reproduced 0-diff by Biljana using [commit 9daf5a0](https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/9daf5a0924f40e114632097384a7d8f8d42305a9)  (before my changes), with additional daily restart output:
/discover/nobackup/borescan/par/NL5_reproduce_NLv4p_M01_M07_nbits/

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/catchment.F90
gmao-rreichle: Additional fixes were needed, also on the develop branch, see #535.  These fixes were merged into the branch associated with this PR in  [commit 6904c78](https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/6904c78bf90c5279945ce0ee2c287bb6b578a13e).
RESOLVED.

gmao-rreichle:
Added non-zero-diff PEATCLSM cleanup: [commit 9dcb682](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/524/commits/9dcb6820f61d8eed540a5b508b557b90b3cfaef2).
These changes are non-zero-diff w.r.t. Sarith's original experiment (see comment above).

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
gmao-rreichle: Addressed with [commit 9dcb682](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/524/commits/9dcb6820f61d8eed540a5b508b557b90b3cfaef2).

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/catchment.F90
gmao-rreichle: Addressed with [commit 9dcb682](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/524/commits/9dcb6820f61d8eed540a5b508b557b90b3cfaef2).

gmao-rreichle (approved):

sanAkel:
@biljanaorescanin, @gmao-rreichle,
I will review on behalf of the @GEOS-ESM/ocean-team. Please help me gather a few things:

1. Once I build this branch: `feature/borescan_merge_sm_peat`
2. Re-generate the `BCs` using `make_bcs` for my favorite ocean resolution(s), but using `NLv3`
3. I expect the **SAME** or _almost_ same answer as I have from the `develop` version of `GEOSgcm_GridComp` ⬅️ **is this correct**? 

If ⬆️ checks out, that would be fine with me. 

I will try to get back to you by end of _this_ week. 🙏

gmao-rreichle:
@sanAkel, @mathomp4, @yvikhlya, @zhaobin74, @lcandre2, @sdrabenh: 

Many thanks, @sanAkel, for offering to review on behalf of the @GEOS-ESM/ocean-team, which made me realize that we must have touched files (probably GEOSsurface_GridComp/Utils to do with restarts and bcs) that require approval from the @GEOS-ESM/ocean-team and @GEOS-ESM/landice-team.  This is normally a good thing, but in this particular case it throws a major wrench into the works.  We need this PR to be merged immediately, before the big GOCART-2 changes. I've been in touch with @sdrabenh about this.

The PR introduces new peat physics that are important and can't be delayed for the next SMAP L4 product release.  My very strong preference would be for the ocean and landice teams (or @mathomp4 for both) to take my word for this and approve the PR so that @sdrabenh can proceed.  

The changes in this PR, incl. those in GEOSsurface_GridComp/Utils, only impact Catchment and **do not impact ocean or landice**. From the perspective of ocean and landice, the PR is trivially 0-diff.  Also, we verified that the PR is 0-diff, which @sdrabenh is also verifying on his end.  Changes in bcs are only cosmetic and limited to land (improved documentation and improved consistency of mask inputs for EASE-grid land bcs, which aren't used by ocean or landice).

Sorry I didn't catch the ocean and landice connections earlier and failed to give everyone a heads-up.  And I now also realize that the ocean and landice teams were probably bombarded with a ridiculous number of GitHub emails from this PR.  Sorry!!!

sanAkel:
@gmao-rreichle, NP! Following is _comforting_ (I still feel the _burn_ from `NLv3` default! 😞 ) 

> The changes in this PR, incl. those in GEOSsurface_GridComp/Utils, only impact Catchment and **do not impact ocean or landice**. From the perspective of ocean and landice, the PR is trivially 0-diff. Also, we verified that the PR is 0-diff, which @sdrabenh is also verifying on his end. Changes in bcs are only cosmetic and limited to land (improved documentation and improved consistency of mask inputs for EASE-grid land bcs, which aren't used by ocean or landice).
> 

We (@GEOS-ESM/ocean-team) co-own GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90 https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/7bae5aa00dfac25726f3b9dea658d3cace7b7f2f/.github/CODEOWNERS#L36 (though I am not sure if these reviews are strictly being enforced!) and indeed _your_ changes for it look okay, the only thing I would like to **note** for now is probably to start paying attention to the added expense of `LocStreamTransform` ... Otherwise, all seems okay! 

Thanks again!

gmao-rreichle:
Thanks, @sanAkel, for pointing out the details of ownership.  (Among the changed files, there's also overlap with landice for SurfParams.F90, but there are only whitespace differences.)
Re. the LocStreamTransform: I thought this only kicks in when the new (peat-related) diagnostics are written out in HISTORY, but I might be wrong.  In any case, there are only 2 new diagnostics, which are important going forward, while there are many more (pre-existing) diagnostics associated with dirty snow that are useless at this time.  If there is a cost to keeping these around, a (separate) cleanup of the dirty snow elements would provide a more fruitful approach to realize savings.

sanAkel:
> Thanks, @sanAkel, for pointing out the details of ownership. (Among the changed files, there's also overlap with landice for SurfParams.F90, but there are only whitespace differences.) Re. the LocStreamTransform: I thought this only kicks in when the new (peat-related) diagnostics are written out in HISTORY, but I might be wrong. In any case, there are only 2 new diagnostics, which are important going forward, while there are many more (pre-existing) diagnostics associated with dirty snow that are useless at this time. If there is a cost to keeping these around, a (separate) cleanup of the dirty snow elements would provide a more fruitful approach to realize savings.

@gmao-rreichle if these _extra_ ones are for `HISTORY`, then, of course, there won't be any addition if these are not present in the HISTORY. (However, in the past, some of us - incl. myself! added our developments and corresponding additions to HISTORY, which ultimately got _tagged_ along and included in _stock_ versions, and sadly enough nobody cared else cared for those fields! With this github way, hopefully we will use such versions of HISTORY in our local branches, instead of committing to the the `develop` version of the `App`.)

sdrabenh (approved):

gmao-rreichle:
Many thanks, @sdrabenh, for moving ahead with the peat PR.  Much appreciated.  It beats me, though, why this was possible without approval from the @GEOS-ESM/landice-team.  @mathomp4: Would this be something to take a closer look at?

lcandre2:
> > Many thanks, @sdrabenh, for moving ahead with the peat PR. Much appreciated. It beats me, though, why this was possible without approval from the @GEOS-ESM/landice-team. @mathomp4: Would this be something to take a closer look at?
> 
> Huh. As far as I can see from the settings, it should have. Very odd. I'll have to research this and see.

I did come to look at/approve this this morning and didn't see any approval needed.

PR #546, opened 2022-02-17
Run History in predictor

bena-nasa:
This was a request by @christophkeller to allow some possible optimization for future chemical reanalysis. The desire was to run History in the predictor loop when running a replay. I have enabled this by instantiating a 2nd instance of History in GCM gridded component if doing replay and only if the user requests output by providing the appropriate RC that is distinct from the main History instance in cap. This is zero diff and should not conflict as it is off by default. 
Note this will require MAPL v2.18


sdrabenh:
@bena-nasa is this PR finalized yet? Or should ignore for now with the DNA label?

sdrabenh (approved):

PR #550, opened 2022-02-22
Option to use MODIS Snow Albedo Information 

biljanaorescanin:
This branch enables the use of the new snow albedo information. The new snow albedo info is based on 20Y of observations from MODIS Terra. 
To deliver the new snow albedo, MOD10A1 product was first obtained from the LIS group. A two-step approach is used to get the product implemented here. In step1, we use MODIS snow over information to determine areas with 60% or more of snow cover. Then, in step2 we use the top 10th percentile of the snow albedo over these regions. The final product is not changing over time, i.e. the new snow albedo information is static. 

Files need to be copied from: 
/discover/nobackup/borescan/for_AGCM_test/nc4_format
to:
/discover/nobackup/projects/gmao/osse2/stage/BCS_FILES/MODIS_snow_alb

Note: Bill created this location when the binary format is used in the testing phase. 

Comparison between one-day AGCM runs, using the new .nc files vs. the old .bin files, confirmed there are no changes in the output. 

Since this implementation is under flag "snow_first =.false."  in GEOS_CatchGridComp.F90 it will maintain zero diff vs. develop. To use this new snow albedo information user has to recompile with flag true. ( In that case one day AGCM test, EXP vs. develop, is not zero diff due to the changes in the albedo information. )

The results with this dataset were presented at the modeling meeting and approved to be added to develop. 

-------------
A downside of this data set is its limitation to 60S-90N latitudes (the dataset obtained from the LIS group provides only data north from 60S.) In the future development, the land group will use the raw data to generate full coverage and implement snow albedo information so it becomes part of make_bcs processing.  The processing of 22Y of the raw data will take time, as will the implementation to the make_bcs and testing the framework, hence the release of this middle PR.


gmao-rreichle:
> Since this implementation is under flag "snow_first =.false." in GEOS_CatchGridComp.F90 it will maintain zero diff vs. develop. To use this new snow albedo information user has to recompile with flag true.

I think what's needed for now is a runtime switch in the rc file that is off by default but easy to flip to on (i.e., use new snow albedo) without recompiling.  This can probably be accomplished by simply adding the corresponding rc variable into GEOS_SurfaceGridComp.rc.  Then the user could readily edit the rc file manually after setup.  If Bill and Scott want to include it in gcm_setup, it would require a  matching PR for the GCM.  Or perhaps Bill and Scott are ok with the recompile implementation.

cc: @sdrabenh @wmputman @rdkoster

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
```diff
@@ -4855,6 +4859,58 @@ subroutine Driver ( RC )
...
+           call MAPL_LocStreamGet(LOCSTREAM,  NT_GLOBAL=NT_GLOBAL, TILEGRID=TILEGRID, RC=STATUS) ; VERIFY_(STATUS)
+           call MAPL_TileMaskGet(tilegrid,  mask, rc=status)                                     ; VERIFY_(STATUS)
+
+           ALBEDO_FILE="/discover/nobackup/projects/gmao/osse2/stage/BCS_FILES/MODIS_snow_alb/MODIS_snow_alb_" // trim(GRIDNAME) // "_nel_Global.nc"
+           call InFmt%open(ALBEDO_FILE,pFIO_READ,rc=status) ; VERIFY_(status)
+           InCfg=InFmt%read(rc=status)                       ; VERIFY_(status)
+           variables => InCfg%get_variables()
+           var_iter = variables%begin()
+           
+           allocate(global_tmp_incr(NT_GLOBAL),source =0.0)
```
gmao-rreichle, line 4883: The re-use of the variable "global_tmp_incr" here is highly risky.  This variable is intended to ingest LDAS increments, which isn't turned on by default but is needed for the coupled land-atm DAS (LADAS).  If the variable is allocated here, the repeated allocate statement when LDAS increments are read is likely to create problems.  I suggest changing this variable name to something else, otherwise you would need to test this PR in the LADAS configuration.

sdrabenh:
> I think what's needed for now is a runtime switch in the rc file that is off by default but easy to flip to on (i.e., use new snow albedo) without recompiling.

I agree - we don't really want to recompile to enable the use of a feature. We typically read from the rc files to do this.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
bena-nasa: Rolf, In Fortran, any allocatable variables in local scope get automatically deallocated so if the same variable is allocated each time a subroutine is called this is not a memory leak as the variable would have been deallocated automatically.
What I do see is that global_tmp_incr is allocated at line 4883 then there is another allocate statement at 5354 allocating the same variable... So if it's already allocated early in the subroutine on the same execution cycle of that subroutine that would cause an error.
Also I would really break this up into subroutines, and get rid of ifdefs and use runtime conditionals, this is rather hard to parse between the long code blocks and the use of ifdefs should really be avoided unless there's absolutely no other way, which most of the time there is. I'm having a hard time understand if this code would trip based on two allocations in one call all the conditionals

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
```diff
@@ -4855,6 +4859,58 @@ subroutine Driver ( RC )
...
+           if(.NOT. allocated (MODIS_SNOW_ALBEDO)) then
+             allocate(MODIS_SNOW_ALBEDO(NTILES), source = 0.0)
+           endif
+           
+           do while (var_iter/=variables%end())
```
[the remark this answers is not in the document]
bena-nasa: I'm not sure why that iterator over the variable names there is there, you can just do
```
if (MAPL_AM_I_Root(VM)) then
                call MAPL_VarRead ( InFmt,"Snow_Albedo"), global_tmp_incr)
  endif
 call ArrayScatter(MODIS_SNOW_ALBEDO, global_tmp_incr,tilegrid, mask=mask, rc=status) ; VERIFY_(STATUS)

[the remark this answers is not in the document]
Comment (bena-nasa) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90:4888]: Unless you need to first see if "Snow_Albedo" is in the file? Is that a use case, that the file exists but would not have "Snow_Albedo"

**Comment by bena-nasa** [comment]
> > I think what's needed for now is a runtime switch in the rc file that is off by default but easy to flip to on (i.e., use new snow albedo) without recompiling.
> 
> I agree - we don't really want to recompile to enable the use of a feature. We typically read from the rc files to do this.

Yes, ifdefs should only be used if it can't be accomplished via a logical conditional incode that may be obtained by parsing an external input file. So pretty much there are very few cases ifdefs should be used. They just make code hard to read and as you say, you have to rebuild...

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90 (hunk 1 above)
Comment (biljanaorescanin) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90:4883]: I agree. I will rename this variable. I did change this in my other sand box but it just stayed here.

**Comment by biljanaorescanin** [comment]
Thanks everyone for the comments, I will need a bit more time to address everything.  I do agree the code can be simplified.

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90 [hunk 3]
```diff
@@ -4855,6 +4860,58 @@ subroutine Driver ( RC )
...
+              vname => var_iter%key()
+              if (MAPL_AM_I_Root(VM)) then
+                call MAPL_VarRead ( InFmt,trim(vname), global_alb)
+              endif
+              if ( trim(vname) == "Snow_Albedo" ) then
+                call ArrayScatter(MODIS_SNOW_ALBEDO, global_alb, tilegrid, mask=mask, rc=status) ; VERIFY_(STATUS)
```
gmao-rreichle, line 4895: This read and scatter seems to work only if the land domain is "global". In offline simulations with GEOSldas, however, we routinely use non-global domains such as CONUS or Australia.  

I think there is a step missing that maps the global tile space to the local tile space, as in the "Catchment increments" example elsewhere in this file:
   `call ArrayScatter(local_tmp_incr, global_tmp_incr, tilegrid, mask=mask, rc=status)`
(It looks like the code here was just adapted from the "increments" read.)

Having said that, I wonder if the "increments" read in the Catchment GC would actually work with non-global domains.  I doubt this was ever tested because the target use case for the "increments" read is the land-atmosphere DAS, which is necessarily always on the global domain.  But for the "snow albedo," we must have the option to run on non-global domains.  

For reference, leaf-area index (LAI), for example, is read in the Surface GridComp, I think, where the mapping to the tile space is handled.  
GEOSldas doesn't have GEOS_SurfaceGridComp.F90 and handles the sub-domain stuff for variables like LAI in GEOSldas-specific code.

I assume this can be fixed with the help of the SI team, but I'm starting to wonder if it isn't simply too rushed to force the "snow albedo" stuff into develop without properly modifying make_bcs.

cc: @wmputman @sdrabenh @bena-nasa @weiyuan-jiang @rdkoster @saraqzhang

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/GEOS_SurfaceGridComp.rc
```diff
@@ -245,3 +245,9 @@
 # GEOSagcm=>SCALE_ALBFPAR: 0
 # GEOSldas=>SCALE_ALBFPAR: 0
 
+# ---- NEW SNOW ALBEDO INFORMATION
+#    .FALSE.  : Default; uses current snow albedo arbitrary information 
+#    .TRUE.   : New; uses product based on 20Y of MODIS Terra 
+#  
+# GEOSagcm=>NEW_SNOW_ALBEDO: .FALSE.
```
gmao-rreichle, line 252: Can we use something like "SNOW_ALBEDO_MODIS" instead of "NEW_SNOW_ALBEDO"?  What do we do if there's another revision, then "new" becomes "old"?
Perhaps better still, leave some room for future versions and use something like:

```
# ---- SNOW ALBEDO
#
#    0  : Snow albedo look-up table (default) 
#    1  : Snow albedo derived from MODIS Terra Collection XXX (20xx-20xx)
#  
# GEOSagcm=>SNOW_ALBEDO: 0
# GEOSldas=>SNOW_ALBEDO: 1
```

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
weiyuan-jiang: @gmao-rreichle The read and scatter here is not exactly from "global" to local. It is more like from "full" ( GEOSldas's concept) domain to local processes.  In GEOSldas' offline CONUS simulation,  "CONUS" is a 'full' domain and the root scatters tile data to all the other processes. So we should be fine with it. The change should be in our GEOSldas setup.  The GEOSldas setup maps the global to "CONUS" and tricks GEOSldas.x to think it is a "globall" domain with corresponding mapped tile data file.

weiyuan-jiang: @gmao-rreichle @biljanaorescanin This hard coded file will not work work with GEOSldas offline CONUS simulation

gmao-rreichle: @weiyuan-jiang:  I assume you're referring to the fact that the file name incl the full path is hard-coded.  @biljanaorescanin, if I understand correctly, to make this in GEOSldas we would have to be able to subset the albedo file in pre-processing and place the file into the ../input directory.  A potential solution would be to change the rc variable back to specifying the path and file name of the snow albedo input.  If the rc variable is an empty string (default), use the lookup table values, if there's a path and file name, read the snow albedo from there.  
@weiyuan-jiang, this would be somewhat similar to what we do with the veg opacity in GEOSldas.  Could this work?

weiyuan-jiang: Yes, we need a reference to this file from rc configure file

mathomp4: I'm just seeing this now in my attempts to run @wmputman 's new code (which apparently has the ALBEDO_FILE). This definitely needs to be in an RC file or something.

Even better if we could refer to it a la ExtData with a non-specific path (though I could "if NAS, do sed" something in the gcm_setup).

biljanaorescanin: @mathomp4 as I said when we just talked, for .bin files there is nothing in RC file, you have to turn it off during compilation ( that is setup Bill has in his old branch that uses **.bin** files). In this PR it is in Surface RC file but uses **.nc4** format of files. General consensus was that this is not so important to be gcm_setup. ( we had that as option at one point and removed it in favor of having in in surface RC. But maybe I didn't get the NAS reference here...) We are working on adding this to the make boundary conditions package ( make_bcs) so we can create all needed resolutions from the raster file for snow albedo. It is just taking a bit longer since I have to download, stitch and reprocess 22Y of newest v6.1 collection to make raster and then work on adding to the bcs package. We left this PR for now just in case Bill wants to use .nc4 format that is only reason.

weiyuan-jiang:
@biljanaorescanin I am confused here. Have we talked about adding this variable into the restart file ?

biljanaorescanin:
> @biljanaorescanin I am confused here. Have we talked about adding this variable into the restart file ?

@weiyuan-jiang this was just a transition PR for Bill to have option to use `nc4` files and not `bin` files. This PR will be obsolete once we have added branch and made a new PR I've shared with you. 
This is the same variable. 
@gmao-rreichle wanted this done the right way with boundary conditions package and restart files.

biljanaorescanin:
Closing this PR since this will be fully addressed with #618.

PR #551, opened 2022-02-22
Remove Ocean Grid Comp

mathomp4:
This PR removes the Ocean Grid Comp for a move to its own repo. See:
- https://github.com/GEOS-ESM/GEOSgcm/issues/377
- https://github.com/GEOS-ESM/GEOSgcm/pull/379

This will utterly fail CI due to removal of files and chicken-egg problem in CI (it checks out `main` of GEOSgcm)


mathomp4:
All tests show the split-off ocean is zero-diff.

tclune (approved):
poof

sanAkel:
Nice! Thank you @mathomp4. 
Just want to be sure your test:
>All tests show the split-off ocean is zero-diff.

is with the **coupled model**, right?

yvikhlya:
Honestly, I would object this step if I saw this earlier. Ocean is a very thin component with sub components below. Too many thin repos can create maintenance problems in the future. But what is done is done. No tragedy.

sanAkel:
> Honestly, I would object this step if I saw this earlier. Ocean is a very thin component with sub components below. Too many thin repos can create maintenance problems in the future. But what is done is done. No tragedy.

Maybe you should layout what _problems_ you anticipate? 

( @tclune and @mathomp4 do not have any problems.)

Also, FYI: 
1. what you see _now_ may not remain so, in terms of repository content - it can only _grow_!
2. There are other repositories in GEOS-ESM of this nature.

yvikhlya:
> Maybe you should layout what _problems_ you anticipate?

First thing that comes to my head are cross repository PRs. We will have more of them, because ocean component is strongly dependent on saltwater, sea ice, obio, orad. Multi repo approach is good and I'v been using it since fortuna times, but it works only when repos are more or less independent from each other, which is not the case with ocean component. By isolating it, we will have more work to do, not less. And I am a lazy guy, I don't like more work :)

sanAkel:
> > Maybe you should layout what _problems_ you anticipate?
> 
> First thing that comes to my head are cross repository PRs. We will have more of them, because ocean component is strongly dependent on saltwater, sea ice, obio, orad. Multi repo approach is good and I'v been using it since fortuna times, but it works only when repos are more or less independent from each other, which is not the case with ocean component. By isolating it, we will have more work to do, not less. And I am a lazy guy, I don't like more work :)

Thanks for laying out your concern(s)!

Reg
>because ocean component is strongly dependent on saltwater, sea ice, obio, orad.

That just happens to be the case _until_ now! To elaborate, so far `ocean component` was used nothing other than for coupled modeling.

>By isolating it, we will have more work to do, not less.

Of course, we want to **get more work done**, not less!

sanAkel (approved):
All looks like as _expected_!

tclune:
One element of the overall strategy that may address some of these concerns is the next generation of mepo called "hepo".   This version will be much more hierarchical and will semi-automatically maintain the information that is currently in the components.yaml file.

There is some debate whether hepo will actually simplify all workflows, but the philosophy is that the gatekeeper of Ocean should decide what is "under-the-hood" for that component.  Not something way up in the fixture.     This will at the very least shift who and when updates to versions are made at various levels of the repository.

Note that this is neither a note of support nor a note of resistance to this as a separate repository.   I defer such decisions to those actively involved in the relevant components (and ultimately @wmputman if a consensus cannot be reached).  Possibly in the future the Software Council will provide clearer guidance, but that will be months at best.

sanAkel:
I would like to put _opinions_ aside, let's stick to the facts and nature of work that needs to be done.

(First the _resistance_ started with _thin_ layered repository, then shifted to connectivity of components - which is prevalent in GEOS. While such arguments can always be constructed and as @tclune said ⬆️ _hepo_ would help remedy them, it is what happens in this repository that will ultimately have the _final_ word!)

Connectivities aside, this _split_ allows for possibility of adding features and capabilities that _this_ org (GEOS-ESM) needs for coupled DA, cc: @danholdaway, @rtodling.

yvikhlya:
I would understand if whole ogcm was separated, this would make at least some sense. Separating ocean alone does not make any sense to me and will result in cross repo PRs almost every time we update ocean.

sdrabenh (approved):

bena-nasa (approved):

Issue #169, opened 2020-01-08
Add DataLake mode in the lake component


atrayano:
The current lake model does not adequately predict the surface temperatures for certain lakes. In AMIP mode we use a workaround and label Great Lakes and Caspian Sea as being part of the ocean, and thus the SSTs are provided by DataSea component. This approach cannot be used in the coupled model, and therefore we keep 2 sets of tile file, and also creates restart incompatibility issues between AMPI and coupled restarts.

We need to add a new mode (DataLake) in the lake code which will provide SSTs over certain lakes (for example Great Lakes and Caspian Sea) for observed SSTs (either Reynolds or Ostia). The choice of which lakes are affected is controlled by a mask. provided in a separate file.

sanAkel:
@atrayano please take a look at the code: https://github.com/GEOS-ESM/GMAO_Shared/pull/112 Thank you!

atrayano:
@sdrabenh I am wondering if you had a chance to test this PR. The code is already "protected" so the default is not to use dataLake, and it should be zero diffs. For this test you do not need any adidtional files

sanAkel:
> @sdrabenh I am wondering if you had a chance to test this PR. The code is already "protected" so the default is not to use dataLake, and it should be zero diffs. For this test you do not need any additional files

Someone also needs to test this feature by turning it "ON"! I understand when it is "OFF" i.e., default- it gives the same answers.

sdrabenh:
@atrayano yes this is 0-diff as anticipated. I would like to turn this on for more testing. Do we have an appropriate mask yet similar to the existing mode (Great Lakes and Caspian Sea)? Or anything I can test?

atrayano:
Thanks! I do not have any mask file at the moment, but I could try to generate one in the next couple of days

From: Scott Rabenhorst ***@***.***>
Reply-To: GEOS-ESM/GEOSgcm_GridComp ***@***.***>
Date: Tuesday, March 15, 2022 at 10:28 AM
To: GEOS-ESM/GEOSgcm_GridComp ***@***.***>
Cc: "Trayanov, Atanas L. (GSFC-610.1)[SCIENCE SYSTEMS AND APPLICATIONS INC]" ***@***.***>, Mention ***@***.***>
Subject: [EXTERNAL] Re: [GEOS-ESM/GEOSgcm_GridComp] Add DataLake mode in the lake component (#169)


[...]

PR #559, opened 2022-03-04
Feature/ltakacs/r21c observed precip

lltakacs:


sdrabenh:
Requires companions PR [#169](https://github.com/GEOS-ESM/GEOSchem_GridComp/pull/169)

gmao-rreichle (dismissed):
Looks ok to me.  One question, though: Should we change the default latitude band for tapering of precip corrections?  With the IMERG-based corrections in R21C, the latitude band should be 50-60 N/S (rather than 42.5-62.5 N/S as in M-2).  Below is a link to the MAPL_GetResource() hard-coded defaults.  
The same numbers are probably also in a GCM rc file somewhere. 
cc: @gmao-qliu 

Precip corr Tech Memo: https://gmao.gsfc.nasa.gov/pubs/docs/Reichle1344.pdf

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/3a337d43f120cb81e2f60456e85b779a04f7336f/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90#L5957

lltakacs:
> Looks ok to me. One question, though: Should we change the default latitude band for tapering of precip corrections? With the IMERG-based corrections in R21C, the latitude band should be 50-60 N/S (rather than 42.5-62.5 N/S as in M-2). Below is a link to the MAPL_GetResource() hard-coded defaults. The same numbers are probably also in a GCM rc file somewhere. cc: @gmao-qliu
> 
> Precip corr Tech Memo: https://gmao.gsfc.nasa.gov/pubs/docs/Reichle1344.pdf
> 
> https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/3a337d43f120cb81e2f60456e85b779a04f7336f/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90#L5957

Yes, Rolf, that is a good idea.  I will try to merge that into my feature branch.

sanAkel (commented):
Does this impact:
1. Coupled model configuration? I can't tell!
2. What is the difference (from what I get now) over ocean (and sea ice; even in atmosphere-only mode)?
     - liquid precip
     - frozen precip
     
If anything !=0 why?

lltakacs:
> Does this impact:
> 
>     1. Coupled model configuration? I can't tell!
> 
>     2. What is the difference (from what I get now) over ocean (and sea ice; even in atmosphere-only mode)?
>        
>        * liquid precip
>        * frozen precip
> 
> 
> If any difference !=0 why?

If you are NOT using an Observed Precip file, this code is zero-diff.  Otherwise, the code now uses a MAPL-2 interface to read the Observed Precip file.  The method of partitioning the observed precip has not changed.  For the case of R21C, the Observed Precip file is associated with IMERG data and is defined on the cubed-sphere.  The default values used for the latitudinal tapering have changed from (42.5-62.5) to (50-60).  The MAPL-2 interface may give round-off differences if reading the Lat/Lon MERRA-2 observed precip file.  I did not check this.

mathomp4:
Random question: is this a configuration I should be running in my nightly tests for any reason? Or is it the sort of thing where it works or it doesn't rather than 'might change nightly'?

sdrabenh:
@mathomp4 from an AGCM standpoint, I don't think this needs to be a nightly test at this point. The default is off.

zhaobin74:
Doesn't affect landice. Good to go.

sanAkel:
>> If you are NOT using an Observed Precip file, this code is zero-diff. 

Thank you @lltakacs, that's great!

Based on what @sdrabenh says:
>@mathomp4 from an AGCM standpoint, I don't think this needs to be a nightly test at this point. **The default is off.**

There is an _option_ 👍 

Thanks!

sanAkel (approved):

sanAkel:
> Doesn't affect landice. Good to go.

@zhaobin74 then please hit "approve" 😄

zhaobin74 (approved):

sdrabenh (approved):

PR #576, opened 2022-04-20
further cleanup of make_bcs utility and CatchmentCN routines

gmao-rreichle:
An inconsistency in NLv5 Catchment model parameters was discovered that impacts 6 tiles (out of ~1.6 million) on the EASEv2 M09 grid.   The 6 tiles in question end up with some PEATCLSM parameters (incl. porosity=0.93), even though they (probably) should have mineral soil properties (which is still reflected in some parameters associated with the 0-30 cm soil layer).  An investigation is ongoing.

During this investigation, a few minor problems with the make_bcs script were discovered and fixed in the branch associated with this PR.  More fixes might be needed.

The PR will be zero-diff for all model runs.  It is not yet clear how the inconsistency in the Catchment model parameters will be resolved.  We will probably need to introduce a new bcs version to fix the parameter inconsistency, which probably means that will the PR will also be zero-diff for the make_bcs package.  

cc: @rdkoster @gmao-jkolassa @biljanaorescanin @weiyuan-jiang @gmao-qliu 

UPDATED 11 Jun 2022: This PR will be **merged without addressing the above-mentioned parameter inconsistency**, which requires further study.  The bulk of the PR consists of long overdue cleanup.

UPDATED 7 May **2026**:  Catchment model parameter inconsistency was addressed in #1196


gmao-rreichle:
**UPDATE (5/5/22):**
The subroutines dealing with HWSD/PEATMAP processing and the Woesten soil parameters were partially cleaned up and some documentation was added.  The changes are **not** meant to be final.  The latest commit [ef69e84](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/576/commits/ef69e8496be5415424864a9a4a7f13c131855dd2) successfully tested 0-diff for the generation of NLv4 and NLv5 bcs in EASEv2_M09 resolution.  Additional tests for other bcs versions and resolutions will be needed.

It remains unclear how the tiles identified by Sebastian end up with peat as profile soil class.  The attached pdf and xls files provide additional information.
[SMAP_L4_Meeting_20220504_NLv5_bcs_only.pdf](https://github.com/GEOS-ESM/GEOSgcm_GridComp/files/8636213/SMAP_L4_Meeting_20220504_NLv5_bcs_only.pdf)
[soilparam_peat_tiles_rr_20220504.xlsx](https://github.com/GEOS-ESM/GEOSgcm_GridComp/files/8636220/soilparam_peat_tiles_rr_20220504.xlsx)

cc: @rdkoster @gmao-jkolassa @biljanaorescanin @weiyuan-jiang @gmao-qliu

gmao-rreichle:
UPDATE (23 May 2022):

Additional cleanup was merged from the following supporting PRs:

[remove generation of obsolete bcs file CLM4.5_veg_typs_fracs #586](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/576/commits/1b3ace58eb4db8f7d387bf93f8d68f3bb78c26fe)

[cleanup of incomplete CatchCN albedo scaling #589](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/576/commits/cad95dde43155ca26bd5d2124caf6f4e537b46c3)

#589 touches the CatchmentCN GridComp and requires 0-diff verification of the GEOSldas Catchment**CN** tests.

Also, make_bcs requires testing to see if the branch successfully reproduces the existing sets of bcs (keeping in mind that the obsolete bcs file `CLM4.5_veg_typs_fracs` is no longer generated and a couple of IDL plots were renamed).

Since the (standard) Catchment model was not touched, the PR is trivially 0-diff for all GCM tests and most GEOSldas tests (except those involving CatchmentCN as noted above).

In its current form, this PR does **not** address the inconsistent soil parameters identified by Sebastian.  This is left for future work, provided someone can figure out the logic of subroutine `soil_para_hwsd()` and how we end up with the inconsistent parameters.

@biljanaorescanin, please test this PR as outlined above and report the results here.  If the tests pass, remove the "contingent" label and convert from "draft" to "ready for review" to indicate that the PR is ready to be merged into develop.  

cc: @rdkoster @gmao-jkolassa @weiyuan-jiang @gmao-qliu @sdrabenh

biljanaorescanin:
1. All Nightly tests passed. 
2. I've run one bcs test (EASE M09), files are same as from run when using develop branch with exception to file we removed (file was obsolete), some wording changes in README file and out log file and some plots that visualized information in removed file were removed as well.

gmao-rreichle:
> I've run one bcs test (EASE M09)

Thanks, @biljanaorescanin.  Which bcs version did you run?  Given the extent of the changes in this PR, we need to run multiple versions at multiple resolutions.  At the very least, we need to run NLv4 and NLv5 on M09 and M36, and we need to run NLv3 at c720, c360, c180 and c90.

biljanaorescanin:
@gmao-rreichle I did NLv3. Sure, I will run other cases and add a comment once I am done.

biljanaorescanin:
@gmao-rreichle @weiyuan-jiang I've added last commit so we don't have warning messages in catchcn_params.nc4 and catch_param.nc4 . I am testing it now to confirm.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkCatchParam.F90
```diff
@@ -244,7 +242,15 @@ PROGRAM mkCatchParam
           write (log_file,'(a)')'Skipping step for EASE grid. '
        endif
        write (log_file,'(a)')' '
-       
+
+        open (10, file = 'clsm/catchment.def', form = 'formatted', status = 'old', &
```
gmao-rreichle, line 246: @biljanaorescanin, two things:
1) I'm not sure this fix will work for non-EASE grids.  If you're creating tiles in the cube-sphere space, the block immediately above seems to skip the creation of catchment.def.  Perhaps for the non-EASE grids the file catchment.def is created earlier?  If that's the case, the changes here should be ok. 
2) Can you please rename the variable "maxcat" to "N_tile" (assuming there's no conflict with another "N_tile" variable in the scope).  The variable name "maxcat" does not reflect the meaning of the variable.

gmao-rreichle:
I added the "contingent" label again.  Not sure why this was removed.  We are not quite done with this PR.  Hopefully final touches and testing underway now.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkCatchParam.F90
biljanaorescanin: > @biljanaorescanin, two things:
> 
> 1. I'm not sure this fix will work for non-EASE grids.  If you're creating tiles in the cube-sphere space, the block immediately above seems to skip the creation of catchment.def.  Perhaps for the non-EASE grids the file catchment.def is created earlier?  If that's the case, the changes here should be ok.
> 2. Can you please rename the variable "maxcat" to "N_tile" (assuming there's no conflict with another "N_tile" variable in the scope).  The variable name "maxcat" does not reflect the meaning of the variable.

1.  This fix works for non-EASE grids
2. I renamed maxcat to n_tile as you suggested

gmao-rreichle:
The previous two commits fix a netcdf dimension error that had been present for a long time in the catch_params.nc4 and catchcn_params.nc4 files generated by make_bcs.  The data in the files was successfully verified 0-diff after the commits. 
Owing to the bug in the old version of these nc4 files, nccmp and cdo do not work properly for comparing the catch_params.nc4 files before and after the dimensions bug fix.  I verified 0-diff for data across the old and new nc4 files using matlab.  

@biljanaorescanin successfully completed final developer tests today (GEOSldas runs that use bcs generated with the branch associated with this PR).

gmao-rreichle:
PR is ready for final verification and merging by GCM gatekeeper. 

PR only touches make_bcs and CatchmentCN and is trivially 0-diff for all GCM tests.

Suggested log entry:

```
make_bcs and CatchmentCN cleanup:
- removed generation of unused/obsolete bcs file CLM4.5_veg_typs_fracs
- cleanup of incomplete CatchCN albedo scaling 
- partial cleanup and improved documentation of subroutines dealing with HWSD/PEATMAP processing and Woesten soil parameters 
```

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
```diff
@@ -4458,198 +3888,319 @@ SUBROUTINE soil_para_hwsd (nx,ny,gfiler)
...
+ 
+        ! NOTE: For PEATMAP, oc_sub was cut back to 8./sf above:
+        !       "! move HWSD sub-surface peat to peat-rich mineral Group 3 because merged surface peat defines sub-surface peat"
+        !       "where (oc_sub*sf >= cF_lim(4))                                                                                "
+        !       "    oc_sub = NINT(8./sf)                                                                                      "
+        !       "endwhere                                                                                                      "
+        !       For PEATMAP, the sub-layer weight of 2.33 should only count towards cFamily(1:3), and in most cases the 
+        !       maxloc statement below should therefore result in o_clp = 1, 2, or 3 only.  However, if the top-layer orgC
+        !       is peat for most contributing raster grid cells and the sub-layer orgC values are relatively evenly spread 
+        !       over orgC classes 1, 2, and 3, then maxloc(cFamily) can result in o_clp=4.
```
mbechtold, line 3900: I thought this is what we don't want.
o_clp=4 should only be allowed if o_clp makes up more than 0.5 of a grid cell.
This may explain that Sebastian found 6 'buried' peat tiles which were not peat in the top layer but in the profile layer.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
```diff
@@ -4458,198 +3888,319 @@ SUBROUTINE soil_para_hwsd (nx,ny,gfiler)
...
 
 	if (sum(cFamily) == 0.) o_clp = 1
-	if (sum(cFamily)  > 0.) o_clp = maxloc(cFamily, dim = 1)
+	if (sum(cFamily)  > 0.) o_clp = maxloc(cFamily, dim = 1)        
+
+        ! ----------------------------------------------------------------------------------------
+        ! 
+        ! Determine *top* layer mineral/organic soil class of tile n 
 
         if(o_cl == 4) then
```
mbechtold, line 3909: Wouldn't it make sense to include in this 'if-statement' a line like
o_clp = 4
We want all o_cl to become full peat tiles.
With this we would get rid of those inconsistent tiles with top soil being peat but not profile being peat.

gmao-rreichle:
> We may have to examine why it checks table 2 of pix2 but use table 1 of pix1

Thanks, @weiyuan-jiang, for the sleuthing.  I've never looked at this program and cannot quickly make sense of what's going on here.  I'm still unclear what happens when we run the (usual) build with optimization, and what a fix would look like.  If we also get Pix1=-999 with the optimized build, then lines 344-347 should result in garbage, which is concerning.  @atrayano, are you familiar with the code in question?  [In brief, when we run this code for make_bcs using a debug build, the run stops with an error message that an index value is negative.  No such stop and error message is produced when we run the (usual) optimized build.  See previous comment by @weiyuan-jiang.]

cc: @biljanaorescanin @rdkoster

atrayano:
@gmao-rreichle Yes, I am familiar with this code , it was written by Max Suarez many years ago. I have seen the negative (-999) Pix error once in a while, and usually this is a result of the rasterization process not covering completely the Earth

weiyuan-jiang:
The good news is I only saw one bad pix. Maybe we can throw away that one ? @gmao-rreichle @biljanaorescanin @atrayano

atrayano:
@weiyuan-jiang Unfortunately, we cannot throw away the "bad" Pix. My first impulse is to write a piece of code to check the raster(s) for -999 and possibly do something to re-assign a "good" value (for example, assign the value from a neighbor)

biljanaorescanin:
@sdrabenh I made an issue for things we didn't address  https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/596, we are ready to proceed with this PR as it is.

biljanaorescanin (approved):

sdrabenh (approved):

bena-nasa (approved):

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
gmao-rreichle: Addressed with #1196.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
gmao-rreichle: Addressed with #1196.

PR #584, opened 2022-05-05
fix fvg for clm45

weiyuan-jiang:
@gmao-jkolassa Do you think this is a bug fix?


weiyuan-jiang:
After I made such fix, getids does not converge any more. That means this line will not jump out and keep looping 
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/46010ac6324d04d895881ed7e4aefa305916f830/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/getids.F90#L331

That makes me wonder why clmc45_xxx are calculated but never used.  Has  this bc file been verified?
 https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/46010ac6324d04d895881ed7e4aefa305916f830/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/mk_GEOSldasRestarts.F90#L1444

@gmao-jkolassa @gmao-rreichle

gmao-jkolassa:
@weiyuan-jiang Thanks for looking into cleaning this up. I think I mentioned this already in a previous thread when we talked about this part of the code, but the use of the file `CLM4.5_veg_typs_fracs` is obsolete. CNCLM40 and CNCLM45 were initially supposed to have different plant functional type (PFT) distributions, hence the reason why different files were read. At some point the decision was made to use the same PFT distribution for both models, which is why the code was changed to only pass the variables called `CLMC_XXX` (and not `CLMC45_XXX`). The read statements for `CLM4.5_veg_typs_fracs` and the treatment of the variables called `CLMC45_XXX` was just never cleaned up in the process. 

This means that the parts of the code that read `CLM4.5_veg_typs_fracs` and all calculations involving `CLMC45_XXX` are obsolete and can be removed. This also means that the bug fix you implemented is not needed, since `ITY` and `FVG` are the same for CNCLM40 and CNCLM45. Please let me know if any of this does not make sense!

weiyuan-jiang:
@gmao-jkolassa Thanks for the clarification. I will clean that up in our new program.

PR #589, opened 2022-05-16
cleanup of scaling albedo with fpar

biljanaorescanin:
Removing model/ experiment specific scaling with fpar. This option was by default set to zero, scaling was not used, this cleanup is zero diff. In boundary conditions package, option to create these parameters was commented out as well, so change is zero diff for boundary conditions as well. 
@gmao-rreichle @gmao-jkolassa @rdkoster


gmao-rreichle:
Thanks, @biljanaorescanin, for putting this together.  I don't have time to look at the changes in detail until later this week or next. 
However, I'm wondering if it would not be better to split this PR into two. The plan for #576 was to have only changes in ./Raster.  The make_bcs clean up (changes in ./Raster)  from the present PR can be merged into  https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/576 as set up in the present PR.  
But if we also merge the changes in the model code from this branch into #576, then the scope of #576 broadens and #576 also requires testing for zero-diff model simulations.   Something to keep in mind for now.  We can decide later.

Issue #596, opened 2022-06-10
debug mode exposed an error that exists in optimized mode 


biljanaorescanin:
Testing of bcs package in debug mode for NLv3 with EASE grid M09 exposed an error:
https://github.com/GEOS-ESM/GEOSldas/issues/552

Summarizing comments left on https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/576 here, so we can merge PR where comments were left:
> [GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/CombineRasters.F90](https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/9c3baa128d377b8a2c673f8bf8df31e3b1773b95/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/CombineRasters.F90#L343-L347)
> 
> Lines 343 to 347 in [9c3baa1](/GEOS-ESM/GEOSgcm_GridComp/commit/9c3baa128d377b8a2c673f8bf8df31e3b1773b95)
When line 344 is executed, Pix1 = -999 which leads to crash. It is exposed by debug mode but exists in optimized mode ( just does not crash) .

We may have to examine why it checks table 2 of pix2 but use table 1 of pix1
from @gmao-rreichle 
https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/576#issuecomment-1149117456
from @atrayano:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/576#issuecomment-1150086274

weiyuan-jiang:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/1046239f980616615bc494d54228e35cbc38eba6/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/rasterize.H#L141-L153

We have seen a lot of print out like "undefined pixels".  If it succeeds,  "Rasterization completed successfully" would be printed out. Unfortunately, we didn't see that message for tag NL3 and M09 grid . @biljanaorescanin @gmao-rreichle @atrayano

biljanaorescanin:
[...]

Issue #539, opened 2022-02-11
PEATCLSM outstanding issues


gmao-rreichle:
Summary of things not yet addressed in PR #524:

1. Document revised water balance equation:  WCHANGE = PR - EVAP – RUNOFF - **FSWCHANGE**
2. Revisit re-gridding (re-tiling) of restarts for peat tiles ([..]/Utils/mk_restarts/Scale_Catch[CN].F90).  
3. Revisit hybridization of PEATMAP and HWSD in bcs generation.  Is the peat area fraction [PMAP] ok for non-EASE tile spaces?  Close enough for EASE tile spaces?  Note pmap_thresh=0.5 in mod_process_hres_data.F90.
4. Revisit relationship between peat area fraction [PMAP] from PEATMAP/HWSD consistent with that used for fire (“peatf”) in CatchmentCNCLM45 input file  (CLM4.5_abm_peatf_gdp_hdm_fc)
5. Investigate snow water balance print statements in StieglitzSnow.F90 (~line 971), which were called out in a comment by Michel Bechtold in an early version of PEATCLSM.

biljanaorescanin:
I'm adding few comments left on PR https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/576 related to PEATCLSM:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/576#discussion_r887939241. -- Addressed with #1196.
https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/576#discussion_r887940679 -- Addressed with #1196.

PR #601, opened 2022-06-24
zero-diff bug fixes in make_bcs package  

weiyuan-jiang:
This PR addresses issue #596 .  

Without passing the range into LRRasterize(), the "completed successfully" message was never printed out for EASE grid. https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/1046239f980616615bc494d54228e35cbc38eba6/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/rasterize.H#L153

PR contains additional cleanup and bug fixes.  Without the fixes, array indices were out of bounds and array elements were assigned garbage numbers, hence the non-zero-diff results for some intermediate files generated in M09 and M36 bcs.  These intermediate files do not seem to be used and are no longer created in the present branch, so effectively the PR is 0-diff for make_bcs. 

The PR is trivially 0-diff for running the GCM or GEOSldas.

Adds functionality to point out non-0-diff results between:
- bcs generated with the current code and
- archived bcs used in ops that were generated with an older compiler.  
In this case, the default string for the output dir is appended to clarify expected non-0-diff results from make_bcs.


GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkSMAPTilesPara_v2.F90
```diff
@@ -776,8 +776,8 @@ SUBROUTINE mkEASEv2Raster
         allocate (xs ( nc_smap+1, nr_smap+1))
         allocate (ys ( nc_smap+1, nr_smap+1))
         
-        do  j = 1, nr_smap
-           do i = 1, nc_smap
+        do  j = 1, nr_smap+1
+           do i = 1, nc_smap+1
               x = real(i-1)        -0.5
               y = real(nr_smap - j)+0.5
```
gmao-rreichle, line 782: @weiyuan-jiang : I just noticed that here the latitude index might indeed be turned around from the original convention.  As I said, I don't have enough time to assert which way things are oriented.  It may all be correct as is.

weiyuan-jiang: I have written it out before the pushing. So it should be fine

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkSMAPTilesPara_v2.F90
```diff
@@ -785,20 +785,9 @@ SUBROUTINE mkEASEv2Raster
...
-        where (ys < -90.)
-           ys = -90.D0
-        endwhere
-        where (xs > 180.)
-           xs = 180.D0
-        endwhere
-        where (xs < -180.)
-           xs = -180.D0
-        endwhere
+
```
gmao-rreichle, line 788: @weiyuan-jiang, I agree that this check should be unnecessary, but there's a slight chance that dropping it will result in a non-zero diff change.  The longitude values on the grid edge come from the easeV2_inverse() calculation, and it's conceivable that roundoff will result in something like 180.000001, which would then no longer be reset to 180.0.  Did you or @biljanaorescanin test the PR yet?  If it's 0-diff without this unnecessary check, then I'm ok with merging it.

gmao-rreichle:
@weiyuan-jiang, I have one more question: I can see how passing the min/max lat/lon into LRRasterize() is important for the EASE and EASEv2 grids because these grid do not fully cover the globe.  However, in #596 @biljanaorescanin reports that she hasn't seen the "Rasterization completed successfully" message for the **CF** grids either:  https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/596#issuecomment-1165706310
Since the CF grids should provide full global coverage and (I assume) their min/max lat/lon values match the defaults [-90,90], [-180,180] that are hardcoded in LRRasterize(), the PR should not change anything for the CF grids.  Consequently, we still would not see the "successfully completed" message for the CF grids.  Then the question becomes if that's ok or if we're missing a related issue for the CF grids.
In any case, the key is for @biljanaorescanin to 0-diff test the PR for at least a few representative EASEv2 and CF tile spaces.

weiyuan-jiang:
I think this PR is non-zero diff for EASEv2 grid. For CF grid, it should be zero-diff. I have printed out the x values, it is exact 180.000000 and -180.000000. Of course it is highly possible it could be 180.000001 since the last digit is not reliable.  But this PR it is already non-zero diff anyway.

weiyuan-jiang:
> 

We could not see the message  "Rasterization completed successfully" because this message is redirected to >/dev/null . I tested NL3 C24 case. If I remove ">/dev/null" in the job script, the successful message would be printed out. @gmao-rreichle @biljanaorescanin

weiyuan-jiang:
This line is obviously a bug. tmpstring is not assigned at this point. We probably can just remove it.
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/1046239f980616615bc494d54228e35cbc38eba6/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkSMAPTilesPara_v2.F90#L752

gmao-rreichle:
@weiyuan-jiang, thanks for the explanation and adding the non-0-diff label.  If the PR is indeed not 0-diff for EASEv2, figuring out how to proceed will need more thought.  We would lose the 0-diff backward compatibility for all existing EASEv2 bcs datasets, including the brandnew NLv5 datasets.  
We also still need to fix the issue in item 6. of #539, for which the solution may or may not be 0-diff.
Re. the following comment: 

> This line is obviously a bug. tmpstring is not assigned at this point. We probably can just remove it.
> 
> https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/1046239f980616615bc494d54228e35cbc38eba6/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkSMAPTilesPara_v2.F90#L752

I agree.  We should be able to just remove this line.

mathomp4:
Note from the Gallery, but if you are updating this file, I might recommend making the following changes:

- `iargc()` → `command_argument_count()`
- `getarg()` → `get_command_argument()`
- `getenv()` → `get_environment_variable()`
- `system()` → `execute_command_line()`

The commands on the left are all Fortran extensions that some compilers (e.g., NAG) do not support because, well, Fortran can do it now.

I have a WIP PR (https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/500) trying to change this everywhere in GCM GC, but that's something I only go back to every so often (low priority task). 

But I figure this is one file where the changes can be tested to make sure they work! 😄

biljanaorescanin:
Nightly tests say this is zero diff as it is now. It is not zero diff for boundary conditions is that what you wanted to say @weiyuan-jiang ?

gmao-rreichle:
Thanks for the update, @biljanaorescanin.  Yes, the "non-0-diff" aspect should only apply to creating bcs, and it should only apply to EASEv2 bcs (but not sure about the latter).  The important test is to generate a bcs datasets and see to what extent the output differs from what we generated with a recent tag (or "develop").  I suggest starting with NLv3 for c180 and EASEv2 M36.

gmao-rreichle:
> [...] I might recommend making the following changes:
> 
> * `iargc()` → `command_argument_count()`
> [...]

Fine with me, although at this point it's hard to predict when we can merge this PR, and it's unlikely to be 0-diff for bcs, so throwing in an unrelated change that should be 0-diff may not be ideal.

biljanaorescanin:
> I suggest starting with NLv3 for c180 and EASEv2 M36.

@gmao-rreichle I will tests this as soon as bcs package runs again we are still debugging at this moment.

weiyuan-jiang:
I pushed more changes to avoid crash in debug mode. It is slow for debug mode. I think we don't have to call "create_mapping" every time. We should be able to improve it.

biljanaorescanin:
@weiyuan-jiang there are still errors, this is from C24 NLv3 CS run: 
Screen Shot 2022-06-30 at 1 43 16 PM

weiyuan-jiang:
> @weiyuan-jiang there are still errors, this is from C24 NLv3 CS run: Screen Shot 2022-06-30 at 1 43 16 PM
> 
> For EASE grid M09 NL3 Screen Shot 2022-06-30 at 1 45 48 PM

I saw it and fixing it...

weiyuan-jiang:
@gmao-rreichle @biljanaorescanin Now I can finish make_bcs with NL3 and m9 choices. It is time to check the difference.

biljanaorescanin:
@weiyuan-jiang @gmao-rreichle with last commit we are able now to run both ease grid case and cf case with no errors. As we suspected it is not zero diff. 

**My control/develop EASE m36 NLv3 case from May:** `/discover/nobackup/borescan/brisi/develop_3/SMAP_EASEv2_M36`
**This branch:**
`/discover/nobackup/borescan/brisi/test_pr_601/try_4/bcs_ease/SMAP_EASEv2_M36`

**Files that differ are:**
 ar.new, bf.dat, catchment.def, cti_stats.dat, soil_param.dat, ts.dat , SMAP_EASEv2_M36_964x406.til , Pfafstetter.til, SMAP_EASEv2_M36_964x406.til, SMAP-EASEv2-M36-Pfafstetter.til and SMAP-EASEv2-M36.til file. 

I can take a closer look at all the differences next week. But sole fact that .til  and catchment.def are different make all past bcs files in comparison non zero diff.  Some differences are not related to this branch, but to peat update like soil_param.dat has 5 tiles that are different. Maybe I should just rerun control to isolate just this PR impact, I didn't realize I didn't run these cases after last peat change. 

For C90 NLv3 CS case similar differences:
**Control/develop from May:**
`/discover/nobackup/borescan/brisi/develop_3/NL3/CF0090x6C_CF0090x6C`
**This branch:**
`/discover/nobackup/borescan/brisi/test_pr_601/try_4/bcs_cf/CF0090x6C_CF0090x6C`

Same files are affected as in ease case summary.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/rmTinyCatchParaMod.F90
```diff
@@ -6215,9 +6215,9 @@ SUBROUTINE FUNCIDEP(                                         &
             endif
          enddo
          locmax=MAX(3,indimax10)
-
-      endif      ! if (nmax .ge. shift+1) 
-                  
+         ! add protection here in case nmax <3 . why 3 ?
+         if (locmax > nmax) locmax = nmax
+      endif      ! if (nmax .ge. shift+1)
```
gmao-rreichle, line 6220: @weiyuan-jiang: It turns out this code was not written by @rdkoster.  My sense it that it was written by Agnes Ducharne, who was helping @rdkoster with the development of the Catchment model in the mid/late 90s.  There are some telltale signs in the code that suggest Agnes was the last person to touch this code, see comments at the beginning of subroutine tgen().
In any case, the code is too complex for me to understand within a reasonable amount of time why the number 3 is hardwired here and below.  Since the protection seems sensible and does seem to work when running the debug build, I think we should just leave the protection as is now.
cc: @biljanaorescanin

gmao-rreichle:
@biljanaorescanin, when you get a chance, please add a commit that addresses the directory (version) name of the existing bcs (i.e., use "NL3.2" instead of "NL3") in the main script (`make_bcs`).  Once you committed this change, I'll refine the interactive comments printed from `make_bcs`.

gmao-rreichle:
@biljanaorescanin, many thanks for your last commit addressing the make_bcs reproducibility issues.  I mulled this over a bit more and added another commit that makes the changes less invasive.   In a nutshell, rather than define new BCs versions within the code, only two things are really needed: 1) in the interactive questions part, let the user know if the selected BCs version cannot reproduce the archived copy of the BCs and 2) if so, change the name of the default output dir accordingly.  
This new solution should be easier to maintain going forward.  Please take a look and let me know if you have any comments or concerns.
If everything looks ok to you, please go ahead and generate the following BCs:
NL3 c360
NL4 m36
NL4 m9
NL5 m9
and compare them to the archived BCs.  With any luck, we can isolate the differences to a few tiles.  I'm keeping my fingers crossed!

biljanaorescanin:
During testing branch didn't want to take user defined path, commit I pushed is fix for that. 

As far as results:
NL3 c360 - has just minor log differences ; no files differ

NL4 m36, NL4 m9 ,NL5 m9 - All ease grid runs have difference in subdirectory `/til` , different number of tiles when compared to what is on develop. Those are not main .TIL file we use that one is the same. 

Files that differ are:
```
til/SMAP-EASEv2-M09-Pfafstetter.til
til/SMAP-EASEv2-M09_SMAP-EASEv2-M09-Pfafstetter.ind
til/SMAP-EASEv2-M09_SMAP-EASEv2-M09-Pfafstetter.TIL
til/SMAP-EASEv2-M09.til
``` 
I will take a closer look at these files to see where they are used and what will be affected.

gmao-rreichle:
> During testing branch didn't want to take user defined path, commit I pushed is fix for that.

@biljanaorescanin, I tried to understand the problem and fix, but I'm still a bit confused.  Why did we use "setenv" for EXPDIR in the first place (in the develop branch, line 315)?  
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/d725913dda446e4fdbd6036e0141471ff4f637e0/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs#L315
I couldn't find an instance in `./Raster/` where EXPDIR is needed as an environment variable.  Should all instances of `setenv EXPDIR ...` be changed to `set EXPDIR ...`?  Perhaps there's some other code from some other GridComp (maybe MAPL) that requires EXPDIR as an environment variable?  But it looks like here the only use of EXPDIR is as a local variable.

biljanaorescanin:
Before it worked, it is like that on develop. Now, on this branch, you added one `IF` loop and with that something changed and now it has to be `SET` or it doesn't take users path. You can try it in your sandbox and see, you just print `EXPDIR` around that section. 

Looking at these files it looks like `if( $TRIPOL_OCEAN == TRUE ) then` they will make a change. So I will add a tripol option to testing to see how it really affected those boundary conditions.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
```diff
@@ -51,11 +51,11 @@ if ( "$1" == "-noland" ) set SKIPLAND = YES
...
 
-   setenv NCPUS `/usr/bin/lscpu | grep '^CPU(s)' | cut -d ':' -f2 | head -1 `
-   @ NCPUS = $NCPUS / 4
-   @ NCPUS = $NCPUS * 3
-   set NCPUS = 20
-   setenv OMP_NUM_THREADS 1
+   #setenv NCPUS `/usr/bin/lscpu | grep '^CPU(s)' | cut -d ':' -f2 | head -1 `
+   #@ NCPUS = $NCPUS / 4
+   #@ NCPUS = $NCPUS * 3
+   setenv NCPUS           20
```
gmao-rreichle, line 57: @biljanaorescanin: 
I'm not sure why NCPUS was hardcoded here, overwriting the earlier setting(s).
I'm also not sure if NCPUS is needed as a global environment variable.  Maybe we could just use
`set NCPUS = 20` ?
For now, I'm leaving it as an env variable, but if NCPUS is just a local variable, we should use "set", not "setenv".

biljanaorescanin: Yes, we can use: 
 set NCPUS = 20 
I am still trying to figure out why we use 20 and not more.

biljanaorescanin: I talked to Matt and he thinks it was either memory issue or it stopped scaling around then.

biljanaorescanin:
If we remove from make_bcs script this section: 
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/e91a8ad0fda513a16645fae91171b2689203174a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs#L1315-L1321

We get zero diff for EASE grid cases m9 and m36. So without "if loop" section on this branch vs when compared to develop.

Two things:
 1. we never run EASE grid with AGCM
 2. we always run bcs script for EASE grid options with O1 ocean
 
So question is why did we ever have this section?  Maybe there was preparation to run some cases with different oceans? 

I am able to run GEOSldas without these files and experiments are zero diff if I use this branch produced m36 vs develop branch produced m36 bcs files. ( those /til files are never used in experiment setup)

gmao-rreichle:
> If we remove from make_bcs script this section:
> 
> https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/e91a8ad0fda513a16645fae91171b2689203174a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs#L1315-L1321
> 
> We get zero diff for EASE grid cases m9 and m36. So without "if loop" section on this branch vs when compared to develop.
> 
> Two things:
> 
> 1. we never run EASE grid with AGCM
> 2. we always run bcs script for EASE grid options with O1 ocean
> 
> So question is why did we ever have this section? Maybe there was preparation to run some cases with different oceans?
> 
> I am able to run GEOSldas without these files and experiments are zero diff if I use this branch produced m36 vs develop branch produced m36 bcs files. ( those /til files are never used in experiment setup)

@biljanaorescanin : Are you saying we could and should remove the lines in question?  If that's the case, we should remove this block.

biljanaorescanin:
With this clean up we are zero diff to develop created ease grid m36 and m09. 

This branch will not create files in 2 subdirectories /til and files in subdirectory /rst. 

These files are in /rst :
image
And in /til :
image

This result is now same as we have for M03 resolution.

gmao-rreichle:
> > [...] I might recommend making the following changes:
> > 
> > * `iargc()` → `command_argument_count()`
> >   [...]
> 

> Fine with me, although at this point it's hard to predict when we can merge this PR, and it's unlikely to be 0-diff for bcs, so throwing in an unrelated change that should be 0-diff may not be ideal.

@biljanaorescanin, @mathomp4: Since the PR turned out to be 0-diff for make_bcs after all, it would be ok to now add the changes suggested by @mathomp4.  If the final tests unexpectedly show non-0-diff, we'd have to undo the changes, but that's unlikely.  So if you're still interested in cleaning this up, please go ahead.

biljanaorescanin:
@mathomp4 I've changed all in "Raster" directory.

image
I can change on some other branch all in "mk_restarts"  (since this PR is just raster related).

gmao-rreichle (approved):

gmao-rreichle:
@sdrabenh, I just approved this PR for the land group.  It's trivially 0-diff for running the GCM because the changes are confined to the ./Raster (bcs generation) directory.  We successfully tested the bcs generation.  Please let me know if you have any questions.

PR #604, opened 2022-07-07
WIP: Updates for SCM - Bill Branch

mathomp4:
These are some updates for the SCM found by myself and @narnold1 necessary to run the SCM with the @wmputman development branch. They are:

1. Add `VARFLT` import to datmo
2. Reorder the setup routines in Moist so that microphysics setup runs first because binary restarts expect `Q` first
3. Pass in an MPI communicator to `gfdl_cloud_microphys_init` because in SCM, this is the first call to `fms_init()` and it was crashing without a comm (usually FV3 initialize FMS)

There is still an outstanding issue due to this code:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/df20a679456bae71b8b9656dd2c8f56fcbbec08d/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90#L4869-L4881

which expects a MODIS Snow Albedo file for every grid. But for SCM this becomes:
```
 /discover/nobackup/projects/gmao/osse2/stage/BCS_FILES/MODIS_snow_alb/MODIS_snow_alb_XY1x1-C_34N_100W_38N_95W_nel_Global.bin
```
which does not exist.

It is up to @wmputman and @gmao-rreichle to figure out the best way to handle this. (@atrayano might have some thoughts.)

This also should be used with https://github.com/GEOS-ESM/GEOSgcm_App/pull/337


gmao-rreichle:
@mathomp4, 
@biljanaorescanin is still working on the snow albedo inputs.  The direct read from a file is meant to be a temporary crutch, prompted by @wmputman's request to be able to run "develop" immediately with the revised snow albedo.  Eventually, the revised snow albedo parameters are meant to become part of the bcs and enter the model through InternalSpecs in the restarts, but you've probably heard that make_bcs has lots of issues.  I wouldn't expect a quick solution, which means that some other hack may be needed.  Or perhaps the revised snow albedo isn't necessary for SCM runs and can just be turned off.  I'll defer to @wmputman on what's sensible and needed here.  But again, implementing the proper solution will take lots of time. 
cc: @rdkoster

gmao-rreichle:
Following up on my earlier comment, after talking with @biljanaorescanin, I understand now that @wmputman's branch here has the use of the "new snow albedo" hardwired into it.  This could be avoided by updating @wmputman's branch as in #550.  The latter PR has been ready to use since Feb but may now be somewhat outdated.  In any case, I think the solution is to avoid forcing the use of the new snow albedo in Bill's branch.  Then for running the SCM, the config should use the old snow albedo formulation and everything should work.   I don't think that for SCM runs the new snow albedo is important.  Let me know if that's not the case.

mathomp4:
Close in favor of #725

PR #618, opened 2022-08-10
Add Option to Use MODIS-based Snow Albedo Climatology 

biljanaorescanin:
**Overview:**

- Enables use of MODIS-derived, static snow albedo in Catchment ("land") tiles.  **Not** (yet) implemented for CatchCN.
- Adds new land bcs versions (v06, v07, v08 and v09). 

**Details:** 

The new snow albedo is read into the GCM or GEOSldas via the Catchment restart file.  Prior to that, the data are processed in two steps: 

1) Preprocess MODIS snow albedo ('MOD10A1.061') for 2000-2022:  First, MODIS snow cover information is used to determine times & locations with snow cover fraction greater than 60%.  Second, for these snow-covered times & locations, the average of the top 10th percentile snow albedo is calculated at 30 arc sec "raster" resolution and stored in 36-by-18 regional files.  The global mean over "land" (i.e., non-glaciated land) is computed (weighted by the raster grid cell area) and used to fill raster grid cells for which MODIS does not provide a snow albedo.  
This preprocessing step is **not** captured in the present PR; this will need to be addressed in the ["make_bcs cleanup" project]( https://github.com/GEOS-ESM/GEOSgcm_GridComp/projects/4)

2) The bcs generation package reads the preprocessed (filled) data from step 1), maps them into GEOS tile space, and adds them to the clsm/catch_params.nc4 bcs file.  

**To do in future versions**: The raster-to-tile remapping in step 2) should be based on the tile ID raster grid.  Here, this mapping is implemented through simple averaging of all raster albedo values within each tile's bounding box (min/max lat/lon), which results in some loss of resolution and is particularly inaccurate along coastlines.

New land bcs versions:

- v06 : NLv3 + JPL veg height + PEATMAP + MODIS snow alb
- v07 : NLv3 + PEATMAP
- v08 : NLv3 + MODIS snow alb
- v09 : NLv3 + PEATMAP + MODIS snow alb

Dependencies:
- requires new MAPL release after merge of https://github.com/GEOS-ESM/MAPL/pull/1825
- related PR: add warning message for users  [GMAO_Shared#304](https://github.com/GEOS-ESM/GMAO_Shared/pull/304)

---------------

Notes:
- Supersedes #550 (closed 29 Aug 2022)
- This PR needs successful merge of https://github.com/GEOS-ESM/GMAO_Shared/pull/276  (done 29 Aug 2022)
- And in GEOSldas merge of https://github.com/GEOS-ESM/GEOSldas/pull/551  (done 2 Sep 2022)


biljanaorescanin:
@weiyuan-jiang here is my branch with just processing and writing this information to catch_param.nc4 files. Next step is to add this to restart file and after reading this information from it.

weiyuan-jiang:
@biljanaorescanin Now, SNOWALB is added to CatmentRst.F90. If this variable is in the boundary file catch_params.nc4, it will be added to catchment's restart file.

biljanaorescanin:
@gmao-rreichle I've changed NLv6 to be V06 and added 2 more options for testing. We can remove them if it turns out we don't need separate options for just peatlands and other for just snow albedo. I've cleanup some comments and indents as well. 
PR is tested and now our catch_params.nc4 is zero diff when compare to land options that don't use snow albedo. 

There is no need for PR to be draft at this point but I am leaving contingent label since we still need merge of GMAO_shared and GEOSldas PR's listed under notes for this PR to work. 

I will close #550 since at this point it is obsolete. There is no more need for that intermediate step. I feel it is just confusing people at this point.

biljanaorescanin:
Since we merged both needed PR's one in GMAO_shared and other on GEOSldas side there is no need to keep this PR draft anymore. 
@gmao-rreichle remove contingent label when you think PR is ready for Scott.

gmao-rreichle:
This PR had not yet gone through a formal code review by the land team.  After taking a closer look today, I think some additional cleanup and improved documentation are needed, which will then also require re-testing.   
I'll add changes and comments as soon as I can and will report back when I've done my bit.  
I'm also confused why the PR fails the CI "build-test"/"run-GCM-on-gfortran", among other CI failures.
Changing back to "draft" status for now.

gmao-rreichle:
> I'll add changes and comments as soon as I can and will report back when I've done my bit.

I added the easy changes in https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/618/commits/d58f3460d2335aa2277a2c887136e0e9a46943f5 .  Should build and be 0-diff (to be verified at a later time).  I'll add more changes next week.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/GEOS_SurfaceGridComp.rc
```diff
@@ -114,6 +114,14 @@
 #
 # GEOSagcm=>MODIS_DVG: 0                     # NOT yet used in GEOSldas
 
+# ---- SNOW ALBEDO  
+#
+#    0  : Snow albedo parameterization based on look-up table (default)
+#    1  : Snow albedo derived from MODIS Collection MOD10A1.006 (2000-2022)  [NOT YET AVAILABLE FOR CatchCN]
```
gmao-rreichle, line 120: @biljanaorescanin : The data period for MOD10A1.006 cannot possibly extend through all of 2022.  This needs to be clarified, as in [month]/2000-[month]/2022.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
```diff
@@ -2986,6 +2992,235 @@ END SUBROUTINE hres_gswp2
...
+  ! Read min/max lat/lons to use when locating snow albedo grids in 
+  ! the stitched MODIS albedo file
+  allocate (min_lon(1:maxcat))
+  allocate (min_lat(1:maxcat))
+  allocate (max_lon(1:maxcat))
+  allocate (max_lat(1:maxcat))
+  allocate (snw_alb(1:maxcat))
+
+  ! Start by setting all snow albedo values to missing
+  snw_alb(:)=-9999.0
```
gmao-rreichle, line 3039: @biljanaorescanin: If I'm not mistaken, the no-data-value for snow albedo in the nc4 output will be -9999.0.  
By construction, the fields in the nc4 file except snow albedo do not have missing values.  But snow albedo does, so we need to introduce proper no-data-handling.
1) Change the no-data-value that is used in catch_param.nc4 to MAPL_UNDEF (that is, 1.e15, the standard GEOS no-data-value). 
2) Add the FillValue Attribute to catch_param.nc4 so that the proper no-data-value is identified inside the nc4 file. 
3) Make sure the same is true for Catchment restart files generated with CatchmentRst.F90.
4) Double-check that 1.e15 works as the no-data-value for snow albedo in the model.  (It should be ok, based on the "if >0 and <=1" statements in GEOS_CatchGridComp.F90.) 

PS: Note that in an earlier commit I removed a redundant line "snw_alb(n)=-9999.0" that was present within the loop through tiles below.

gmao-rreichle:
@biljanaorescanin, @weiyuan-jiang,
Related to my [inline comment](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/618#discussion_r966172453) above, are the changes in this PR compatible with regrid.pl and remap_restart.py? 
The regrid/remap utilities could potentially be tripped up in two ways:
1) The presence of the field for snow albedo in the Catchment restart file depends on the bcs version.
2) The snow albedo field has no-data-values (which requires careful interpolation onto a new tile space).

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/GEOS_SurfaceGridComp.rc
biljanaorescanin: I just pushed clarification for this. 
We used data from Feb 24th 2000 till March 24th 2022. I will omit the date and just leave month. 
Also I had to fix collection you've put it was 006 but we used updated version 061. 
https://nsidc.org/data/mod10a1/versions/61#anchor-1

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs
```diff
@@ -141,9 +141,9 @@ if ( $HELPMODE != YES ) then
        $dummy == 'NL3'  | \
        $dummy == 'NL4'  | \
        $dummy == 'NL5'  | \
-       $dummy == 'v06'  | \
-       $dummy == 'v07'  | \
-       $dummy == 'v08') then
+       $dummy == 'V06'  | \
+       $dummy == 'V07'  | \
+       $dummy == 'V08') then
```
gmao-rreichle, line 146: I don't understand this change.  I thought we intentionally moved to the lower-case "v"?  Maybe there's another place that needs adjusting to the lower-case "v"?

biljanaorescanin: Change is for lower case to work. Let me explain offline

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
```diff
@@ -896,6 +903,17 @@ subroutine SetServices ( GC, RC )
...
   VERIFY_(STATUS)
 
+  call MAPL_AddInternalSpec(GC                  ,&
+    LONG_NAME          = 'snow_albedo'               ,&
+    UNITS              = '1'                         ,&
+    SHORT_NAME         = 'SNOWALB'                   ,&
+    FRIENDLYTO         = trim(COMP_NAME)             ,&
+    DIMS               = MAPL_DimsTileOnly           ,&
+    VLOCATION          = MAPL_VLocationNone          ,&
+    RESTART            = MAPL_RestartRequired        ,&
```
weiyuan-jiang, line 913: Should not make it required. Otherwise all the internal restart file without SNOWALB will fail

biljanaorescanin: @mathomp4 just sent me email saying same thing. I suppose I should than make it 
RESTART            = MAPL_RestartOptional 
We don't want this anywhere else except in run using bcs v06 that has snow albedo.

mathomp4: Note that you then have to be careful with any references to `SNOWALB` later as it might not always exist. But it does look like `SNOW_ALBEDO_INFO` might be that protection!

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
biljanaorescanin: @gmao-rreichle No, you are not mistaken – the no-data-value for snow albedo was set to -9999.0; and yes - snow albedo does have missing values. At the time being (see the latest commit) the comments are addressed as follows:
1.	The no-data-values is changed into MAPL_UNDEF (1.e15)
2.	MAPL_AddInteralSpec does not allow '_FillValue' attribute. Maybe @mathomp4 or @weiyuan-jiang know what is the right argument (or function) to use. I tried with DEFAULT as well - did not work. Currently, catch_param.nc4 holds correct no-data-values, but does not provide this info via the attribute. 
3.	Catchment restart files follow the above: correct no-data-values are now implemented, but no info in the .nc file via the argument. (attempted to use both the string and actual variable when assigning to the '_FillValue' argument via 'call var%add_attribute'). 
4.	Yes, the 1.e15 does work in the model.

biljanaorescanin:
> 

This doesn't work directly from `remap_restart.py` since we don't have option to choose `m36` grid going from lets say C180. 
But manually it does work, @weiyuan-jiang I believe said this will be part of the clean up. But for now we still can run this only by manually forcing `yaml`. 
For example to run it I have to go and manually change line in `remap_catchANDcn.py`  
line `in_tilefile` can't be `=  /*-Pfafstetter.til` since ease grid doesn't have that extension for its tile file so it becomes just `/*.til` 
Every time I've run this I get confused so I ask Weiyuan for help.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
weiyuan-jiang: LDAS_is_nodata actually takes care of -9999.0 or MAPL_UNDEF

weiyuan-jiang:
> > @biljanaorescanin, @weiyuan-jiang, Related to my [inline comment](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/618#discussion_r966172453) above, are the changes in this PR compatible with regrid.pl and remap_restart.py? The regrid/remap utilities could potentially be tripped up in two ways:
> > ```
> > 1. The presence of the field for snow albedo in the Catchment restart file depends on the bcs version.
> > 
> > 2. The snow albedo field has no-data-values (which requires careful interpolation onto a new tile space).
> > ```
> 
> This doesn't work directly from `remap_restart.py` since we don't have option to choose `m36` grid going from lets say C180. But manually it does work, @weiyuan-jiang I believe said this will be part of the clean up. But for now we still can run this only by manually forcing `yaml`. For example to run it I have to go and manually change line in `remap_catchANDcn.py` line `in_tilefile` can't be `= /*-Pfafstetter.til` since ease grid doesn't have that extension for its tile file so it becomes just `/*.til` Every time I've run this I get confused so I ask Weiyuan for help.

You don't have to change the code if you can change the yaml and edit the catch_tilefile. But I should change it to *.til.  I though I had changed it, bit I hadn't.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
gmao-rreichle: @biljanaorescanin, @weiyuan-jiang, 
Thanks for changing the no-data-value to MAPL_UNDEF, @biljanaorescanin.
Since this variable is part of the GCM, we cannot rely on LDAS_is_nodata(), @weiyuan-jiang.  For the GCM, the no-data-value is MAPL_UNDEF.
We do need to add the FillValue attribute into the nc4 file somehow.  If MAPL does not provide a simple way via Internal specs, then we have to use the generic nc4 interface.  I recognize that this is a bit of a pain, but if we stick with no-data-values in the restart, we have to identify them through the FillValue attribute.  The only other solution would be to make up a static snow albedo value for the tiles that aren't covered by MODIS processing, rather than use the old look-up table approach (which is structurally very different because the albedo depends on snow mass and therefore changes with time, so we can't use the "old" snow albedo values as a time-invariant parameter in the restart file).

gmao-rreichle:
>> [...]  The regrid/remap utilities could potentially be tripped up in two ways:
>> 
>>     1. The presence of the field for snow albedo in the Catchment restart file depends on the bcs version.
>> 
>>     2. The snow albedo field has no-data-values (which requires careful interpolation onto a new tile space).

On 2nd thought, item 2. is probably irrelvant because the (new) snow albedo in the restart file is a time-invariant parameter, so it should not be interpolated (like, say, snow mass).  That is, the target restart should just receive the tile-space snow albedo values from the target bcs, without spatial interpolation. 

Item 1. is still valid, though.  Regrid.pl and remap_restarts.py will need to be able to handle the presence or absence of snow albedo in the source and target restart files depending on the selected source and target bcs. 

> This doesn't work directly from `remap_restart.py` since we don't have option to choose `m36` grid going from lets say C180. 

I don't understand how this answer relates to my earlier comment above.  If remap_restart.py cannot map from c180 to m36, that's a problem that has nothing to do with snow albedo.  Does regrid.pl have the same limitation?  I suspect the limitation is because the EASE grid  is still not fully integrated in MAPL?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
bena-nasa: @gmao-rreichle When we create the checkpoints at the end of any MAPL component which all go throug the same codepath, currently we do not set a _FillValue, rather we set _NoFill = true.
It has been like this for a decade. The rational was probably that you should not have missing values in the state written via checkpoint, these are not diagnostic but the internal states used in computation.
If there is a need for the checkpoints to have _FillValue (which for the time being would be MAPL_UNDEF) please let me know. We can certainly do this, but then your code that is using such a checkpoint which becomes the restart for the next segment had better be able to protect against such values getting into a calculation.

I would advocate in any offline tool when creating restarts set the _NoFill to true. If you are using our PFIO wrapper to NetCDF we may not have an interface to set this in the NetCDF file. If needed we can add it.

gmao-rreichle: @bena-nasa, thanks for the clarification.  The restart variable in question is a spatially distributed, time-invariant parameter, not really a model "state" (as in model prognostic variable).  Personally, I never quite understood why GEOS repeatedly saves all of the model's time-invariant parameters in every restart file, although I recognize it's convenient.  

I'm afraid the situation is a bit of a catch-22.  If we enforce _nofill=true, we have to make up a static snow albedo value where we didn't have one before (because the old parameterization was structurally different, see above).  But using the old parameterization to backfill the new one doesn't seem to be consistent with the GEOS assumptions for restart files.

I suppose we could read the (gappy) snow albedo field directly from the bcs file at the beginning of the simulation?  This is (somewhat) akin to the temporary snow albedo fix that Bill has in his current "development" tag, except that we now have a way (through the present PR) to properly generate the snow albedo field for each resolution within a given set of bcs.  (For Bill's fix, Biljana had to manually create the input files.  Also, Bill's fix re-reads the snow albedo file at each and every time step, I think.)  In this scenario, the snow albedo input would become something like an "ExtData" file.  To me, this seems like a lot of overhead for a static model parameter that (almost) looks like any one of the other 50-ish land model parameters. 

Maybe we should just make up a default time-invariant snow albedo value to backfill the MODIS-derived values?  

cc: @rdkoster

weiyuan-jiang:
> 

The remap_restart.py will recognize the extra SNOWALB variable depending on the BCS, but not the regard.pl

gmao-rreichle:
FYI, I merged the latest develop branch into the branch associated with this PR, which required resolution of a minor conflict.  Should be ok now.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
gmao-rreichle: UPDATE: We revised the preprocessing strategy for the MODIS snow albedo on the raster grid, see this PR's introductory comment above.  The assumption is now that all raster grid cells will have valid snow albedo values and consequently all land tiles will also have valid snow albedo values.  A safety check for no-data-values should still be present, just in case something went wrong or future snow albedo raster files violate this assumption.  Resolving conversation for now.

gmao-rreichle:
> The remap_restart.py will recognize the extra SNOWALB variable depending on the BCS, but not the regrid.pl

This is tricky.  The idea is to finish this PR by the end of next week, and based on Bill's interest in having this functionality, the assumption is that the SNOWALB restart variable will be used by the modeling group immediately.  However, the modeling group has yet to make the transition from `regrid.pl` to `remap_restart.py`. 

One option out of this would be to add a stop into regrid.pl when Catch restarts that  include  SNOWALB  are involved.  This would encourage users to migrate to remap_restart.py and avoid developing the outdated regrid.pl.

cc: @sdrabenh @weiyuan-jiang @biljanaorescanin @rdkoster

biljanaorescanin:
@gmao-rreichle this is ready for you. 
Preprocessing files are here: `/discover/nobackup/projects/gmao/bcs_shared/make_bcs_inputs/land/albedo/snow/MODIS`
v1 - noFill ( has missing values) 
v2- FillVal ( where no data values were filled with global land average Snow_Albedo=0.56 ) 
I've tested new raster files and run bcs package for few diff grids. I also updated few comments but we can revise that. Only major changes are in preprocessing files.

biljanaorescanin:
@gmao-rreichle `NLv3` - C720 O2, `NLv4` - EASE M09 and `NLv5`-  EASE M09 are all zero diff ( only changes are in readme file, log and error file and running script file). 
Last commit is adding case `v09`, this case includes both peatlands and snow albedo.

biljanaorescanin:
@gmao-rreichle  After we added the fill value for snow albedo, issue on the dateline was reveled for high res CF bcs runs. At first I pushed previous commit as a fix but realized that shouldn't be necessary if indexing was right. Problem in short, zero index doesn't exist in fortran. Issue is fixed and all was rerun to confirm fix,  except I will have to redo spinup that was ongoing.

gmao-rreichle:
> The remap_restart.py will recognize the extra SNOWALB variable depending on the BCS, but not the regrid.pl

@weiyuan-jiang, @biljanaorescanin: We still need a stop in regrid.pl that aborts execution of regrid.pl if the snow albedo restart variable is in either the source or the target restart file.  We cannot assume that GMAO users will know that they cannot use regrid.pl when the snow albedo restart variable is present somewhere.  Adding a stop into regrid.pl should be a lot simpler than adding the snow albedo functionality into regrid.pl.

gmao-rreichle:
@biljanaorescanin: Per your email message from this morning, I removed the obsolete reading of "tile_id" in subroutine MODIS_snow_alb().  It looks like the tile_id was indeed not used in the subroutine that maps the snow albedo into tile space.  Rather, it looks like the min/max lat/lon of each tile is used for a cruder mapping of the snow albedo from the raster grid into tile space. It's not clear to me if this really what we wanted.  If the snow albedo raster file is on the same grid as the tile ID info, then shouldn't we use the tile ID info to aggregate the snow albedo into tile space?  It's too late in the day for me to go through subroutine MODIS_snow_alb() in detail and compare to what is done in other, similar subroutines.  We'll have to revisit this next week. 
@rdkoster @weiyuan-jiang

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
```diff
@@ -2991,6 +2993,224 @@ END SUBROUTINE hres_gswp2
...
+             
+             ! Generate sums and counts using current tile corresponding indices
+             sno_alb_sum= sno_alb_sum +                                              &
+                  sum(stch_snw_alb(hhtil,vvtil,imin:imax,jmin:jmax),                 &
+                  stch_snw_alb(hhtil,vvtil,imin:imax,jmin:jmax).gt.0.0 .and.         &
+                  stch_snw_alb(hhtil,vvtil,imin:imax,jmin:jmax).le.1.0)
+             
+             sno_alb_cnt= sno_alb_cnt +                                              &
+                  count(stch_snw_alb(hhtil,vvtil,imin:imax,jmin:jmax).gt.0.0 .and.   &
+                  stch_snw_alb(hhtil,vvtil,imin:imax,jmin:jmax).le.1.0)
```
gmao-rreichle, line 3146: @biljanaorescanin: Why do we need the check for unphysical values here? The pre-processor with the 0.56 backfill value should give us a snow albedo raster file that has only "good" albedo values.  
To be safe, we should check that the global min and max of the input raster data are indeed in the range [0,1] immediately after the raster data are read.  But assuming we pass this check, there shouldn't be any more >0 and <1 conditions needed in the mapping from raster to tile space.
cc: @rdkoster

biljanaorescanin: @gmao-rreichle  If we remove the check for unphysical values, we can't run for the files that are not filled. We still have those files as option to use.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
```diff
@@ -2991,6 +2993,224 @@ END SUBROUTINE hres_gswp2
...
+             
+          end do ! vvtil
+       end do ! hhtil
+       
+       ! Calculate snow albedo for the current tile
+       snw_alb(n) = sno_alb_sum / max(1.0,sno_alb_cnt)
+       if (snw_alb(n) .le. 0.0 .or. snw_alb(n) .gt. 1.0 ) snw_alb(n)=MAPL_UNDEF !1.E15
```
gmao-rreichle, line 3153: @biljanaorescanin: With the backfilled snow albedo raster data, could we ever have snow_alb(n) outside of the range [0,1] except through roundoff?  That is, we may need to cut back the roundoff error (from the summation and division) to get the tile space snow albedo to not violate [0,1] but there shouldn't be anything funny if all of the individual raster space snow albedos are within [0,1].
cc: @rdkoster

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
```diff
@@ -2991,6 +2993,224 @@ END SUBROUTINE hres_gswp2
...
+                
+                sno_alb_cnt2= sno_alb_cnt2 +                                                &
+                     count(stch_snw_alb(hhtil,vvtil,imin2:imax2,jmin2:jmax2).gt.0.0 .and.   &
+                     stch_snw_alb(hhtil,vvtil,imin2:imax2,jmin2:jmax2).le.1.0)
+                
+             end do ! vvtil
+          end do ! hhtil
+          
+          snw_alb(n) = sno_alb_sum2 / max(1.0,sno_alb_cnt2)
+          if (snw_alb(n) .le. 0.0 .or. snw_alb(n) .gt. 1.0 ) snw_alb(n)=MAPL_UNDEF !1.E15
```
gmao-rreichle, line 3192: same comment as above - do we really need the checks for [0,1] here?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/CatchmentRst.F90
```diff
@@ -611,6 +623,20 @@ subroutine add_bcs_to_rst(this, surflay, DataDir, rc)
...
+
+        if (meta_%has_variable('SNOWALB')) then
+           if ( .not. allocated(this%snowalb)) allocate(this%snowalb(ntiles))
+           call MAPL_VarRead ( CatchFmt ,'SNOWALB', this%snowalb, __RC__)
+           if ( .not. this%meta%has_variable('SNOWALB')) then
+              var = Variable(type=pFIO_REAL32, dimensions='tile')
+              call var%add_attribute('long_name', 'snow_albedo')
+              call var%add_attribute('units', '1')
+              call this%meta%add_variable('SNOWALB', var)
+           endif
```
gmao-rreichle, line 637: @biljanaorescanin, @weiyuan-jiang: Why are we writing the metadata for snow albedo here and not in function create_meta(), where the meta data for all other restart variables are handled?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
gmao-rreichle: @biljanaorescanin: The files that are not backfilled are not meant to be used for land bcs. These "raw" files exist so that the landice group can decide how they want to fill the map for landice, where the backfill value should be very different from what we use over non-glaciated land.  
cc: @rdkoster

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
biljanaorescanin: @gmao-rreichle Correct. If all is as it should be, the `snw_alb(n)` range should be [0-1]. Meaning, the line 
3153 could be rewritten into: 
`if (sno_alb_sum .ne. 0.0 .and. sno_alb_cnt .ne. 0) snw_alb(n)=min(1.0,max(0.0,sno_alb_sum/sno_alb_cnt))`
This is now implemented in the last commit.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
biljanaorescanin: @gmao-rreichle After implementing the above, this is not an issue anymore and it is removed.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/CatchmentRst.F90
biljanaorescanin: @gmao-rreichle I've used `remap_restart.py` to remap restart with snow albedo from c180 to c720, it didn't have any issues. Is that enough of the test or you have something else in mind?

biljanaorescanin:
> @biljanaorescanin: Per your email message from this morning, I removed the obsolete reading of "tile_id" in subroutine MODIS_snow_alb(). It looks like the tile_id was indeed not used in the subroutine that maps the snow albedo into tile space. Rather, it looks like the min/max lat/lon of each tile is used for a cruder mapping of the snow albedo from the raster grid into tile space. It's not clear to me if this really what we wanted. If the snow albedo raster file is on the same grid as the tile ID info, then shouldn't we use the tile ID info to aggregate the snow albedo into tile space? It's too late in the day for me to go through subroutine MODIS_snow_alb() in detail and compare to what is done in other, similar subroutines. We'll have to revisit this next week. @rdkoster @weiyuan-jiang

Yes, tile_id is not used and should be removed. It stayed in the code as a leftover from the original attempt to bring in the SNOWALB from the tile-space raster (when it was stored in soil_param.dat file). This idea was abandoned. Snow albedo raster is on regular 30 arc sec grid. Therefore, we use min/max lat/lon to populate /aggregate the raster info in the tile space.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
biljanaorescanin: @gmao-rreichle only usage of filled files is now in code. I wasn't sure should I remove/cleanup comment in code, path to the original files noFill ones. For now it is there.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/CatchmentRst.F90
weiyuan-jiang: > 

I think the regridding  does not know if it should add the variables at the beginning. After it checks the boundary,  it knows.

weiyuan-jiang:
I think there maybe duplication of assignment of the bcs values.  In the call "re_tile", it re-arrange the values, then in add_bcs_to_rst, it reads again from bcs.  Do you think it is right ? @gmao-rreichle @biljanaorescanin

weiyuan-jiang:
Ok, I figure out there is another logic behind this. In the old code, you don't need out_bcs directory but still can do regrinding. Now the new code can not do that. Do we need to add this function?  @gmao-rreichle @biljanaorescanin

gmao-rreichle:
@biljanaorescanin, @weiyuan-jiang, 
I went through the comments above.  Here's what I think still needs addressing:

1.	We still need a stop in regrid.pl that aborts execution of regrid.pl if the snow albedo restart variable is in either the source or the target restart file. We cannot assume that GMAO users will know that they cannot use regrid.pl when the snow albedo restart variable is present somewhere. Adding a stop into regrid.pl should be a lot simpler than adding the snow albedo functionality into regrid.pl.

2.	It looks like the min/max lat/lon of each tile is used for a relatively crude mapping of the snow albedo from the raster grid into tile space.  E.g., along the coastline, raster grid cells over the ocean are backfilled with 0.56, and these values then contribute to the tile-avg snow albedo.  It's not clear to me if this really what we wanted. Is the snow albedo raster file is on the same grid as the tile ID info?  If so, then shouldn't we use the tile ID info to aggregate the snow albedo into tile space?  How are other, similar parameters mapped from raster to tile space?  (cc: @rdkoster)

3.	From @weiyuan-jiang: In the old code, you don't need out_bcs directory but still can do regrinding. Now the new code can not do that. Do we need to add this function? 

4.	In CatchmentRST.F90, why isn't snow albedo included in read_shared_nc4() and read_GEOSldas_rst_bin() when the other "optional" variables do appear in these subroutines?  I suppose in read_shared_nc4(), there’s no harm in not reading snow_alb because the target snow_alb values come from bcs, not the restart file?  But then most of the other variables that are read in read_shared_nc4() would also seem to be obsolete?  As for read_GEOSldas_rst_bin(), probably the same logic applies, that is, most variables are not needed, but in any case we could not easily add a new variable to a binary restart.  Is the binary restart file obsolete?

weiyuan-jiang:
> 

For 1), I am adding regrinding of SNOWALB to regard.pl
For4). It needs to read those variables then retile them if there is no out_BCS. If we don't want this capability, we can skip the readings of bc values from restart.

gmao-rreichle:
> 1. We still need a stop in regrid.pl that aborts execution of regrid.pl if the snow albedo restart variable is in either the source or the target restart file...

Added by @weiyuan-jiang in latest commit on #634 -- but this would need to be added into #618.

> 2. It looks like the min/max lat/lon of each tile is used for a relatively crude mapping of the snow albedo from the raster grid into tile space... 

After conversation with @rdkoster, it was decided to leave the refinement of the raster-to-tile remapping for future work. 

> 3. In the old code, you don't need out_bcs directory but still can do regrinding. Now the new code can not do that. Do we need to add this function?
>
> 4. In CatchmentRST.F90, why isn't snow albedo included in read_shared_nc4() and read_GEOSldas_rst_bin() when the other "optional" variables do appear in these subroutines?...

This will be addressed in future cleanup (#669).

@weiyuan-jiang, please:
- also add the snow_alb functionality for regrid.pl into this PR #618

@biljanaorescanin, please:
- review the changes in my most recent commit (note revised "if" conditions in line 3138 of mod_process_hres_data.F90)
- test @weiyuan-jiang's updated "regrid.pl" when the source and/or target restart includes snow albedo
- create an "issue" that reminds us to further refine the raster-to-tile mapping (item 2 above)

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/CatchmentRst.F90
gmao-rreichle: Cleanup of read_shared_nc4() addressed in #670.  Comments resolved.

biljanaorescanin:
@gmao-rreichle I've tested PR again for different bcs options and remap_restarts.py with different options.  
All results are as before. 

@sdrabenh Contingent label will stay until we have new MAPL tag, after that this PR is ready for merger. (what we need from MAPL is already on develop).

gmao-rreichle (approved):
Approving for the land team.

mathomp4:
Note: [MAPL 2.31.0](https://github.com/GEOS-ESM/MAPL/releases/tag/v2.31.0) has been released and it should have the change from @weiyuan-jiang. I've also made a PR in GEOSgcm to update the MAPL version, see https://github.com/GEOS-ESM/GEOSgcm/pull/490

PR #634, opened 2022-09-09
Further clean up and re-organize make_bcs

weiyuan-jiang:
First step 1) combine easeV1 and easeV2.  mkSMAPTilesPara.F90 is removed from building. It will be replaced by mkSMAPTilesPara_v2.F90

PR #639, opened 2022-09-12
Feature/wjiang/combine easev1 v2

weiyuan-jiang:
**OBSOLETE!  Addressed with #634.**

Combine easev1 and v2.  Some public interfaces in EASE_conv.F90 will be removed after mkSMAPTilesPara.F90 is removed.


weiyuan-jiang:
Obviously a typo here. Will lead to non-zero diff in the future when it is corrected.
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/e52d0919a2f2fdd50ac2153f153be8e6b5dbbeaa/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkSMAPTilesPara.F90#L604

tclune (commented):
CMake changes ok

gmao-rreichle:
> Obviously a typo here. It is v1, so it should be just 'EASE' . It will lead to non-zero diff in the future when it is corrected.
> 
> https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/e52d0919a2f2fdd50ac2153f153be8e6b5dbbeaa/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkSMAPTilesPara.F90#L604

This correction is fine.  It's only changing a string in a file header and should be easily identifiable as a bug fix.  My understanding is that the current branch already fixes this typo.

gmao-rreichle:
@weiyuan-jiang, a note on my latest commit: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/639/commits/6e467b422c480f9ae6df8408072f7fe5213d78d0
Before I made this commit, I approached this PR in isolation and did not refer back to the [make_bcs cleanup project](https://github.com/GEOS-ESM/GEOSgcm_GridComp/projects/4), and I promptly forgot that we have `LDAS_EASE_conv.F90` in the GEOSldas repo.  I now realize that your new `EASE_conv.F90` was just a copy of `LDAS_EASE_conv.F90`.  (It would have been very helpful if this equivalence had been pointed out in the commit messages or in the PR comments.)
In any case, I'm not 100% sure that `LDAS_EASE_conv.F90` (or, equivalently, `EASE_conv.F90` before my latest commit) is 0-diff against easeV1_conv.F90 because of the difference in the definition of PI for the EASEv1 case.  This should be fixed in my latest commit, along with some minor cleanup of comments and clarification of global constants.  We'll need to decide which version of `EASE_conv.F90` is best going forward.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/EASE_conv.F90
```diff
@@ -0,0 +1,694 @@
...
+  
+  private
+
+  public :: ease_convert
+  public :: ease_inverse
+  public :: easeV1_convert
+  public :: easeV1_inverse
+  public :: easeV2_convert
+  public :: easeV2_inverse
+  public :: easeV2_extent
```
gmao-rreichle, line 87: Ideally, the public interface here should only be the general subroutines.  That is, the version-specific subroutines should not be public.  This may require changes in other files that call the version-specific subroutines, incl. in the  GEOSldas repo.  
We also need a general, public subroutine ease_get_params().
Likewise for ease_extent, if that's needed somewhere. (It may be possible to use ease_get_param() instead.)

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkSMAPTilesPara_v2.F90
```diff
@@ -65,19 +65,21 @@ PROGRAM mkSMAPTilesPara_v2
...
       include 'netcdf.inc'
 
       N_args = command_argument_count()
 
       if(N_args < 1) then
-        print *,'USAGE : bin/mkSMAPTiles -smap_grid MXX'
-	print *,'Allowed SMAP grids are: M01 M03 M09 M25 M36'
+        print *,'USAGE : bin/mkSMAPTiles -smap_grid MXX -v LBCSV -pfaf_til x -ease_version xx'
+        print *,'Allowed SMAP grids are: M01 M03 M09 M25 M36'
+        print *,'Allowed EASE_VERSION are: EASE EASEv2'
```
gmao-rreichle, line 77: Would it be cleaner to ask for EASELabel here?  That is, merge the `-smap_grid` and `-ease_version` arguments.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkSMAPTilesPara_v2.F90
```diff
@@ -65,19 +65,21 @@ PROGRAM mkSMAPTilesPara_v2
...
         stop
       end if
 
+      EASE_Version = 'EASEv2'
```
gmao-rreichle, line 81: I think EASE_Version should be initialized to 'UNDEF', so that the user is forced to input the ease_version via the command line argument.  (There's no documentation of the fact that EASEv2 is the default.  Unless we say somewhere what the default is, we can't just set a default.)
This change would require changing lines 1320-1348 in the make_bcs script (which would also clean up this bit of make_bcs).

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkSMAPTilesPara_v2.F90
```diff
@@ -96,77 +98,92 @@ PROGRAM mkSMAPTilesPara_v2
...
+      if (MGRID /= 'M25' .and. EASE_version == 'EASE') then
+         stop ("EASEv1 only supports M25")
+      endif      
       
-      call execute_command_line('cd data/ ; ln -s /discover/nobackup/projects/gmao/ssd/land/l_data/LandBCs_files_for_mkCatchParam/V001/ CATCH')  
-      call execute_command_line('cd ..')
-      
+      ! WY noted: should do it in the script that calls this program 
+      !call execute_command_line('cd data/ ; ln -s /discover/nobackup/projects/gmao/ssd/land/l_data/LandBCs_files_for_mkCatchParam/V001/ CATCH')  
+      !call execute_command_line('cd ..')
```
gmao-rreichle, line 122: Can we just comment out these lines without changing the script that calls this program?  I didn't see any changes in make_bcs, so I'm not sure these lines can just be dropped.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkSMAPTilesPara_v2.F90
```diff
@@ -96,77 +98,92 @@ PROGRAM mkSMAPTilesPara_v2
...
+            CELL_km = 25.0252600081    ! nominal cell size in kilometers       
+            nc_smap = 1388
+            nr_smap = 584
+            gfile     = trim(EASElabel)//'_1388x584'
+         else 
+            CELL_km = 25.067525         ! nominal cell size in kilometers
+            nc_smap = 1383
+            nr_smap = 586
+            gfile      = trim(EASElabel)//'_1383x586'
+         endif
```
gmao-rreichle, line 158: The EASE grid parameters should **not** be defined again in this program.  They should be obtained with a call to `ease_get_params()` [which is a missing public interface in EASE_conv.F90, see comment elsewhere].  
Moreover, the `gfile` string should not have hard-coded numbers for Nx-by-Ny.  These numbers are part of the parameters provided by `ease_get_params()`.
The units and names of these parameters differ a bit between EASE[v1] and EASEv2, but the info is definitely the same as what is hardcoded here, so it should not be difficult to clean this up. 

This comment applies to lines 129-182.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkSMAPTilesPara_v2.F90
gmao-rreichle, line 77: Could we use "EASEv1" instead of "EASE" for the label of the original version?  I'm a bit lost as to what is used where.  EASEv1 would definitely be clearer, but it may require changes in many places.

gmao-rreichle:
Two more comments on `mkSMAPTilesPara_v2.F90`:
1. I'm confused what subroutine `mkEASEv2Raster()` does, and why this is not done for EASEv1. 
2. Also, subroutine `write_tilfile()` seems obsolete.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkSMAPTilesPara_v2.F90
weiyuan-jiang: We can clean this up after combining mkSMAPTilesPara

weiyuan-jiang:
Do not comment on this PR. It will be replaced by #634

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkSMAPTilesPara_v2.F90
weiyuan-jiang: Combined!

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkSMAPTilesPara_v2.F90
weiyuan-jiang: Done

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mkSMAPTilesPara_v2.F90
weiyuan-jiang: Yes, now everything is SMAP_EASEvx_Mxx

gmao-rreichle:
**OBSOLETE!  Addressed with #634.**  Closing.

PR #640, opened 2022-09-12
Remove unused NCEP code from Gravity Wave Drag grid component

darianboggs:
Gravity Wave Drag grid component has NCEP code located in `if (USE_NCEP_GWD) then .. end if` blocks which never runs since `USE_NCEP_GWD` is initialized to `.false.`, and it never changes. This pull request removes that code as well declarations for variables that are not used outside the NCEP blocks.


darianboggs:
Forgot to add flags and assignment.  Also, I ran the code and compared (`cmpscratch`) it to `develop` with 0 diff.

sdrabenh (approved):

PR #655, opened 2022-10-19
Feature/aoloso/gwd hybrid

aoloso:
GWD GridComp was modified for thread safety. Three derived type variables that are potentially modified from time step to time step were made thread private. The variables are beres_band, beres_desc, and oro_band. Also parameters retrieved with MAPL_GetResource were packed into a newly created GWD private state GEOS_GwdGridComp for thread safety. All "save" statements were removed. Finally all "RC=STATUS & VERIFY_(RC)" pairs were replaced with the macro "_RC".


tclune (approved):
CMake changes ok, but I am concerned that this introduces a new RC file.  Maybe I've misunderstood.  It aligns well with the long term plan, but if people don't know about an RC file for this component, it might need some discussion to increase awareness.

PR #657, opened 2022-10-24
bug fixes to address errors that caused fail of start-stop regression test for most Catchment-CN4.5 land tiles

weiyuan-jiang:
The Catchment CN 's continuous run is non zero-diff from stop-and-continue run. This PR is trying to fix the problem.

Since CatchCN is not included in the GCM suite of tests and since there is no stop&restart regression test in GEOSldas, the PR is **zero-diff** w.r.t. the existing test suites.  
The PR is **non-zero-diff** for users when existing CatchCN runs are reproduced with different stop&restart times.


weiyuan-jiang:
@gmao-jkolassa . I think CNCLM40 is also non zero-diff except there is no cncol or cnpft in the restart file.

gmao-jkolassa:
@weiyuan-jiang Thanks for putting this together! I will re-run my experiments with this branch.

` I think CNCLM40 is also non zero-diff except there is no cncol or cnpft in the restart file`

CNCLM40 **does** have CNCOL and CNPFT in the restart file. When I ran the continuous vs stop-and-continue run for CNCLM40 there were no differences in the restart files after 2 months. So as far as I can tell CNCLM40 was never affected.

I will re-run this test for both CNCLM40 and CNCLM45 just in case.

gmao-jkolassa:
`It is not just concol, there are are other variables.`

@weiyuan-jiang I am not sure I am following, could you please clarify?

CN_init (among other things) maps the CNCOL and CNPFT entries to CLM variables. CN_exit then maps the CLM variables back to CNCOL and CNPFT before they are written to the restart file. I went through the routines in the CNCLM45 CN_DriverMod.F90 and as far as I can tell all variables that are mapped from CNCOL and CNPFT in CN_init are mapped back into CNCOL and CNPFT in CN_exit. Can you please let me know which other variables you mean? Thanks!

gmao-jkolassa:
In its current form, this branch still leads to differences in the restart files for a continuous vs stop-and-continue experiment for CNCLM45. I have been talking offline with @weiyuan-jiang about this and am currently running some additional tests to better isolate the cause.

gmao-jkolassa:
I have run several more tests and I think I have found the issue. There are actually two separate issues, which are both related to the following code block in the `GEOS_CatchCNCLM45GridComp.F90` that resets three variables to 0 on the first time step. 

     `if(first) then
      ! variables used for summing CN inputs over multiple land model calls; not saved on restart 
      ! fzeng: run must end on a CN call step
      ! -----------------------------------------------------------------------------------------
      allocate( lmrsunm(ntiles,nveg,nzone) )
      allocate( lmrsham(ntiles,nveg,nzone) )
      allocate(             runsrf(ntiles) )

      lmrsunm = 0.
      lmrsham = 0.
      runsrf = 0.

      first = .false.

    endif`

These variables are sums over the CN interval, which is why they should be reset after each call to the CN subroutines (that happen at a  timestep of 1.5 hours).

`lmrsunm` and `lmrsham` are reset to 0 after each CN_driver call, meaning the stop-and-continue and continuous experiment are identical for these if the stop-and-continue run is stopped directly after a call to CN_Driver (as indicated by Fanwei's comment in the code). Having the "break" in the stop-and-continue experiment at 0z means that the resetting to 0 of `lmrsunm` and `lmrsham `occurs prior to the 0z CN_Driver call in stop-and-continue and after that call in the continuous experiment.

In the case of `runsrf` I think that there is a bug here. It is never reset after the CN_Driver call, so this variable will always be different between the continuous and stop-and-continue experiments after the "break" irrespective of when the break occurs.

**Regarding solutions:** 
All other variables that are sums over the CN interval are passed as restart variables meaning they to not need to be initialized to 0 on the first time step and they have the same value in the stop-and-continue and continuous experiments irrespective of when the break occurs. So I think it might be easiest to just include `lmrsunm` and `lmrsham` as restart variables here.

For `runsrf`, it should be equal to `RUNSURF`, which is already passed as restart variable making the `runsrf = 0.` statement redundant.

I apologize for the lengthy answer. I am happy to talk about this in person if that is easier.

weiyuan-jiang:
After the PR #658 is merged, I will change CatchmentCNRst.F90 again to make sure the two variables also need re-tile

gmao-rreichle:
@gmao-jkolassa, has there been any progress on this PR?

gmao-jkolassa:
> @gmao-jkolassa, has there been any progress on this PR?

@gmao-rreichle There has been some progress on this. I have identified a bug in the fire module that incorrectly initialized a variable needed to start the agricultural fire season. The effect of this was that stopping and re-starting a run lead to no agricultural fires occurring in the run for the rest of the year. I have fixed this bug by adding 'BURNDATE' as an additional restart variable for Catchment-CN4.5. This has removed the differences between the two regression experiments for most of the points.

However, there are still a few locations with differences. I have traced this bug to the subroutine that computes the stomatal conductance, but have not yet identified the ultimate cause. So this PS is still work-in-progress.

gmao-jkolassa:
This PR in its current form implements several fixes for bugs that caused a stop-start regression test to fail for a majority of land tiles in Catchment-CN4.5. The code changes implemented up to this point fix this issue for most land tiles. There is still one outstanding issue that causes the start-stop regression test to fail in a few (~10) land tiles. The causes for this is as of yet unidentified.

It has been decided to go ahead and merge this PR as is so that the bug fixes already implemented can take effect. A new issue will be created for the remaining bug and the fix to this remaining bug will be addressed in a separate PR. 

As of the time of writing, the branch is up to date with the current 'develop'.

gmao-rreichle (approved):

sdrabenh (approved):

PR #668, opened 2022-11-16
restored for mapl3 by removing adding all exports from child

weiyuan-jiang:
Adding all exports from a child during setservice will not work for MAPL3. So the call is removed and restored to the old ways to add exports.


gmao-rreichle (approved):
Looks ok to me.  Bummer the "all exports" option doesn't work in MAPL3.  
Approving for the land team.

mathomp4:
> Looks ok to me. Bummer the "all exports" option doesn't work in MAPL3. Approving for the land team.

Yeah. It was clever and clean! I hope @atrayano and @tclune can figure out a way for MAPL3 to do something similar.

Thanks @weiyuan-jiang for figuring this out

tclune:
Just to be clear for those that have not been in the voice discussions:   The basic issue is that certain operations are deferred from SetServices into a new Initialize phase.   This is mostly necessary to be compatible with how NUOPC does things.    So where in MAPL2 a component (in its SetServices) can look inside a child component to decide on its own imports/exports, such analysis must be deferred in MAPL3.    Instead, what we can have in MAPL3 is a simple call that expresses the "desire" to do what Weiyuan has implemented.  Then the generic layer can do the actual detail work in the subsequent phase.

The real fault here is mine.   My original intent was to get to MAPL3 by minor increments.    I eventually found that this was too difficult and am now building an entirely new, independent generic layer.    Unfortunately, vestiges of the original approach are in MAPL3 already, but at the time did not break any GEOS components.        I would prefer not to have to implement some kludge in this intermediate implementation, but it will also be some time until the new layer is mature enough to do this the right way.

I had hoped that if I looked away, a new solution would magically appear.   :-)

What we should do, IMO, is design a new procedure (interface) to express the intent.   Then, we (i.e., @atrayano ) will need to implement something in the existing GenericInitialize() that uses this signal to implement what is desired.   In theory not terribly difficult.

tclune:
Opened MAPL issue:  https://github.com/GEOS-ESM/MAPL/issues/1819

mathomp4:
MAPL3 runs again! Huzzah! Unblocking.

sdrabenh (approved):

PR #673, opened 2022-11-18
remove snowalb when target bc does not have

weiyuan-jiang:
remove snowalb from restart file  when target bc does not have. It depends on the PR https://github.com/GEOS-ESM/MAPL/pull/1825 from MAPL

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/CatchmentRst.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/CatchmentRst.F90
@@ -523,6 +523,7 @@ contains
       this%TSA2  = DP2BR      
       this%TSB1  = DP2BR      
       this%TSB2  = DP2BR      
+      this%GNU   = DP2BR      
       this%COND  = DP2BR      
       this%WPWET = DP2BR      
       this%POROS = DP2BR
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/CatchmentRst.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/CatchmentRst.F90
@@ -624,6 +624,9 @@ contains
         CLOSE (24, STATUS = 'KEEP')
         CLOSE (25, STATUS = 'KEEP')
         CLOSE (26, STATUS = 'KEEP')
+
+        if (this%meta%has_variable('SNOWALB'))  call this%meta%remove_variable('SNOWALB')
+
       endif
 
       do n=1,ntiles
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/CatchmentRst.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/CatchmentRst.F90
@@ -581,8 +581,9 @@ contains
               call var%add_attribute('units', '1')
               call this%meta%add_variable('SNOWALB', var)
            endif
+        elseif (this%meta%has_variable('SNOWALB')) then
+           call this%meta%remove_variable('SNOWALB')
         endif
-
         call CatchFmt%close()
       else
         open(unit=21, file=trim(DataDir)//'/clsm/mosaic_veg_typs_fracs',form='formatted')
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/mk_CatchRestarts.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/mk_CatchRestarts.F90
@@ -578,10 +578,12 @@ contains
        call OutFmt%create(OutFileName,__RC__)
 
        if (allocated(snowalb)) then
-          var = Variable(type=pFIO_REAL32, dimensions='tile')
-          call var%add_attribute('long_name', 'snow_albedo')
-          call var%add_attribute('units', '1')
-          call OutCFG%add_variable('SNOWALB', var)
+          if (.not. OutCFG%has_varibale('SNOWALB') then
+            var = Variable(type=pFIO_REAL32, dimensions='tile')
+            call var%add_attribute('long_name', 'snow_albedo')
+            call var%add_attribute('units', '1')
+            call OutCFG%add_variable('SNOWALB', var)
+          endif
        endif
 
        call OutFmt%write(OutCfg,__RC__)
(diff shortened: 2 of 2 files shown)


gmao-rreichle (approved):
@weiyuan-jiang, thanks for putting this together.  Here's another suggestion, if that would be easier or more elegant -- I think we could keep SNOW_ALBEDO in the target restart even if there isn't SNOW_ALBEDO in the corresponding ("old land") bcs, as long as SNOW_ALBEDO=MAPL_UNDEF in the target restart file. Having SNOW_ALBEDO set to the no-data-value (instead of 0) would be safe and clear and could possibly avoid the dependence on MAPL (which might complicate getting the snow albedo PR merged into GEOS GCM GC soon).  Either  way is fine with me,  it's just another option.  I'm approving for now but will let you decide about merging this PR.  
In any case, @biljanaorescanin still needs to test before we can ask Scott to merge the snow albedo PR into develop.

biljanaorescanin:
@weiyuan-jiang this didn't work. We get some weird values for snow albedo:
    59 :-8681-08-02 18:20:00       0   195136       0 :      0.0000     0.38401      37.696 : SNOWALB

biljanaorescanin:
I'm marking this PR draft until we resolve issue with "old land" restarts.

weiyuan-jiang:
> 

Can you show me the directory you tested?

weiyuan-jiang:
@biljanaorescanin I guess I should remove it when it goes to "else" block

biljanaorescanin:
@weiyuan-jiang now it works. I used new branch from MAPL with this PR: [bugfix/wjiang/rm_var_from_meta](https://github.com/GEOS-ESM/MAPL/tree/bugfix/wjiang/rm_var_from_meta)

I suppose what is left is to add "exit" in regrid.pl if someone tries to regrid restarts with snow albedo or we have to add new land option with new boundary conditions to options in regrid.pl. Not sure what option we want to proceed with @gmao-rreichle ?

gmao-rreichle:
> I suppose what is left is to add "exit" in regrid.pl if someone tries to regrid restarts with snow albedo or we have to add new land option with new boundary conditions to options in regrid.pl. Not sure what option we want to proceed with @gmao-rreichle ?

@biljanaorescanin, @weiyuan-jiang : I thought we no longer need the "exit" from regrid.pl because @weiyuan-jiang added this functionality to regrid.pl: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/618/commits/8020e32b0cabee822085555a6e50c2eb95f00c19
I'm not sure, though, if the above commit also works when the source restart uses snow_albedo but the target uses "old land" bcs (ie, no snow_albedo).  Would regrid.pl omit snow_albedo from the target restart or write zeros or do something else?

biljanaorescanin:
> > I suppose what is left is to add "exit" in regrid.pl if someone tries to regrid restarts with snow albedo or we have to add new land option with new boundary conditions to options in regrid.pl. Not sure what option we want to proceed with @gmao-rreichle ?
> 
> @biljanaorescanin, @weiyuan-jiang : I thought we no longer need the "exit" from regrid.pl because @weiyuan-jiang added this functionality to regrid.pl: [8020e32](https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/8020e32b0cabee822085555a6e50c2eb95f00c19) I'm not sure, though, if the above commit also works when the source restart uses snow_albedo but the target uses "old land" bcs (ie, no snow_albedo). Would regrid.pl omit snow_albedo from the target restart or write zeros or do something else?

@gmao-rreichle When we run `regrid.pl` we give restarts to start from and we are asked for boundary conditions used when those restarts were created, there is list to choose from. `regrid.pl `has no option of boundary conditions that created restart with snow albedo. If we want to use regrid.pl to regrid/remap catch restart we would have to add those boundary conditions as options. This works with `remap_restats.py` since there we option where we give our own boundary conditions path we don't have that option with `remap.pl`.  We can talk more offline.
These are options in regrid.pl: 

Sample GCM tags
---------------
F14  : Fortuna-1_4  .........  Fortuna-1_4_p1
F20  : Fortuna-2_0  .........  Fortuna-2_0
F21  : Fortuna-2_1  .........  Fortuna-2_5_pp2
G10  : Ganymed-1_0  .........  Ganymed-1_0_BETA4
G10p : Ganymed-1_0_p1  ......  Ganymed-1_0_p6
G20  : Ganymed-2_0  .........  Ganymed-2_1_p6
G30  : Ganymed-3_0  .........  Ganymed-3_0_p1
G40  : Ganymed-4_0  .........  Heracles-5_4_p3
ICA  : Icarus  ..............  Jason
GITOL: 10.3  ................  10.18
INL  : Icarus-NL  ...........  Jason-NL
GITNL: 10.19  ...............  10.23

weiyuan-jiang:
My understating is that regard.pl won't be able to access the new bcs with snowalb.  Now the question is "Do we need to update regard.pl so it can use the new bcs? " @gmao-rreichle

gmao-rreichle:
Ok, let me try again... Sorry I'm a bit confused as to what was and wasn't done.  There are too many ongoing threads.

For remap_restarts.py, snow_albedo is ok when it's in the target restart.  Also, we can handle the case when snow_albedo is in the source restart but not in the target restart per #673.  
We still need to merge #673 but are waiting for https://github.com/GEOS-ESM/MAPL/pull/1825.

**Question 1**: Is this summary for remap_restarts.py correct? 

For regrid.pl, we can use added the following commit
https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/618/commits/8020e32b0cabee822085555a6e50c2eb95f00c19
but this commit is useless because regrid.pl does not allow any bcs versions that have snow_albedo (for source or target restart).  

**Question 2**: Is this summary for regrid.pl correct? 

**Question 3**: Rather than an "exit", don't we just need a simple comment in the interactive portion of regrid.pl that explicitly states that "bcs with MODIS snow albedo cannot be remapped with regrid.pl and remap_restarts.py should be used instead" or something to that effect?  

I faintly remember that we already talked about this last item before, but I can't find it on github.  It may be buried somewhere in Teams chats.

gmao-rreichle:
PS: We may as well take the opportunity to put a message into regrid.pl that alerts the user to the fact that regrid.pl is being phased out and replaced with remap_restarts.py.  @weiyuan-jiang or @biljanaorescanin, please create a draft GMAO_Shared PR that adds these messages into regrid.pl.

biljanaorescanin:
1. Correct
2. Correct
3. I will make a PR on GMAO_Shared and add these comments to regrid.pl
`"bcs with MODIS snow albedo cannot be remapped with regrid.pl and remap_restarts.py should be used instead"`
and `regrid.pl is being phased out and replaced with remap_restarts.py`
4. I agree, seems like a right way is to remove that commit since we anyway can't use it if we don't change regrid.pl more. @weiyuan-jiang what do you think? Will you remove it or me if you agree? https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/8020e32b0cabee822085555a6e50c2eb95f00c19  ?

gmao-rreichle:
@biljanaorescanin, @weiyuan-jiang:
This PR only goes into Biljana's branch and can be merged now, I think.  We don't want to have two GCM GC PRs for the snow albedo changes.  This should leave us with the following to-do items:

1. Merging this PR will make #618 dependent on https://github.com/GEOS-ESM/MAPL/pull/1825, so this needs to be documented in #618.
2. We still need to remove https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/8020e32b0cabee822085555a6e50c2eb95f00c19  from #618 
3. We need to document on #618 that it is related to https://github.com/GEOS-ESM/GMAO_Shared/pull/304

Issue #667, opened 2022-11-15
Avoid duplication of constants definition in CatchmentCNRst.F90


gmao-rreichle:
CatchmentCNRst.F90 includes several hard-wired, orbit-related and CatchCN-specific constants:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/80c7acdba5d7b3518c44bdf14152826d76c16b7e/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/CatchmentCNRst.F90#L10-L25
These constants exist in other places, e.g., 
https://github.com/GEOS-ESM/MAPL/blob/8bb4b4ae158e8f26400dadd4a5052d0814220d8e/generic/MAPL_Generic.F90#L4119
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/80c7acdba5d7b3518c44bdf14152826d76c16b7e/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM40_GridComp/CLM40/clm_varpar.F90#L43

If at all possible, these constants should not be redefined in CatchmentCNRst.F90

weiyuan-jiang:
Before moving things around, let's see how MAPL is going to solve this issue https://github.com/GEOS-ESM/MAPL/issues/1818

weiyuan-jiang:
I suggest we move catchCN related parameters to LandShared just like catch @gmao-rreichle

gmao-jkolassa:
For the Catchment-CN specific parameters that are originally parameters from CLM and that live in the modules clm_varpar.F90 and clm_varcon.F90, I would suggest to keep them there. Going forward, the plan is to keep the Catchment-CN implementation as modular as possible, so that when new CLM version are available it is less time-consuming to implement them in Catchment-CN. The CLM routines load their parameters from clm_varpar and clm_varcon, so in the interest of not having to change each CLM-routine, I would advocate for keeping the CLM-parameters in those two modules. For inside CatchmentCNRst.F90, we should be able to just import the CLM-parameters through the appropriate use-statements (@weiyuan-jiang I can take care of adding these in if you want).

gmao-rreichle:
[...]

PR #682, opened 2022-12-08
Cleanup of Catchment-CN constants in CatchmentCNRst.F90

gmao-jkolassa:
This PR addresses parts of issue #667 . It removes the duplication of Catchment-CN specific constants in CatchmentCNRst.F90 and instead imports them through use-statements.

I tested this branch in a 2-month experiment for Catchment-CN4.0 and Catchment-CN4.5 and verified that restart files generated by ldas_setup and written out at the end of the 2-month period were identical to the equivalent restart files from experiments run with the 'develop' tag.

**TO DO:** So far, this branch only fixes the issue of the duplicated Catchment-CN constants, but does not address the duplication of orbital parameters (and the use of outdated orbital routines pointed out by Peter Norris). @weiyuan-jiang , do you want me to also make the changes for the orbital parameters or would you prefer to do that yourself?

cc: @gmao-rreichle 

**Update 03/03/2023:**

This PR has been expanded and now includes the following. 
It removes the duplication of Catchment-CN specific constants in CatchmentCNRst.F90 and instead imports them through use-statements. It also removes the duplication of orbital constants and replaces the use of hard-coded orbital subroutines with the relevant MAPL subroutines.

This PR should be zero-diff for Catchment and is zero-diff for Catchment-CN when used with `RESTART: 1` (restarts are of the same resolution as model run). It is not zero-diff for Catchment-CN used with `RESTART: 2 `(restarts are of a different resolution than model run). When running Catchment-CN with `RESTART: 2`, there are differences in the day length variable (dayx) with maximum differences on the order of 10 seconds.

Because changes had to be made to MAPL, this PR is dependent on using MAPL tag 2.35.0


gmao-rreichle:
Thanks, @gmao-jkolassa, for this first part.  Re. the orbital parameters: It's probably better if you take a first crack at it, assuming that Peter talked with you about what needs to be done (in a nutshell: use the day-length calculation from MAPL, rather than doing a custom calculation).  Does that make sense?

gmao-jkolassa:
> Thanks, @gmao-jkolassa, for this first part. Re. the orbital parameters: It's probably better if you take a first crack at it, assuming that Peter talked with you about what needs to be done (in a nutshell: use the day-length calculation from MAPL, rather than doing a custom calculation). Does that make sense?

Sure, I'll get started on that and will check back with Peter if anything is unclear.

weiyuan-jiang:
The  CI-build fails now because it needs develop branch of MAPL.

gmao-jkolassa:
Just wanted to comment that I have been using the MAPL develop branch ever since Peter's changes got pushed to it. I was updating to MAPL develop after cloning the repo. The model has successfully built since updating, but I do see differences in some variables (as expected) and need to establish that these differences are within reason.

gmao-jkolassa:
In the prior version there was still a bug where the latitude information was passed to MAPL_SunGetDaylightDuration() in degrees instead of radians, which led to an unreasonable geographic distribution of day length values. My latest commit fixes this issue. Compared to the implementation in develop, the **maximum** differences in the day length (`dayx`) are now on the order of 10s, which is within the expected range. All other variables are zero-diff. This is the case for CatchCNCLM40 and CatchCNCLM45. So from my end, this is now good to go.

Note: As of right now, this branch is up to date with the current develop branch. However, it does have to be used with the current develop branch of MAPL.

gmao-rreichle (approved):
Approving for Land Team and Surface Preproc Team.

mathomp4 (approved):
Approve for CMake

sdrabenh (approved):

PR #687, opened 2022-12-13
enhancements for MODIS-based snow albedo over land

biljanaorescanin:
This PR makes two enhancements to the MODIS-based snow albedo functionality:
1) For a given land tile, make_bcs computes the MODIS-based snow albedo by averaging over exactly the **land** raster grid cells that make up the tile. (The earlier version of make_bcs computed values by averaging over all raster grid cell values in the vicinity of a land tile, including raster grid cells associated with lake, landice, and ocean).
2) CatchCN can now also use the MODIS-based snow albedo. 

The PR also includes substantial cleanup and improved documentation of Fortran code used for mapping and parameter processing in make_bcs.

Details: 
The input snow albedo is backfilled using MODIS-based climatology at 30-arcsec grid, (i.e., the snow albedo input does not contain no-data values).
Snow albedo assigned to each tile is the average value of 30-arcsec MODIS grid cells associated with a given tile, per the 30-arcsec raster file associated with the tile space. 
Unlike the `MODIS_snow_alb` subroutine on develop branch, the tile-average snow albedo computed here does not include snow albedo values from neighboring land, water, and/or land-ice tiles. In this PR we use `rmap`, the pre-computed mapping of a 30-arcsec raster file, to fill the tile space.
This new feature should work for both catch and catchcn models.


GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
```diff
@@ -2991,9 +2991,192 @@ SUBROUTINE hres_gswp2 (nc_data,nr_data,rmap, gfiler,lai_name,merge)
...
+          jLL=(jx-1)*1200+1 ! element which has coordinate [-179.5,-89.5]
+
+          ! Read the data
+          status = NF_GET_VARA_REAL (ncid,NC_VarID(NCID,'Snow_Albedo'),(/1,1/),(/nc_10,nr_10/),stch_snw_alb); VERIFY_(STATUS)
+
+          do j = jLL,jLL + nr_10 -1  ! j is index in global space; apply the offset
+             do i = iLL, iLL + nc_10 -1 ! i is index in global space; apply the offset
+
+                ! the element has a valid (non-missing value) proceed to calculating mean albedo
+                if(stch_snw_alb(i-iLL +1 ,j - jLL +1) .gt. 0 .and. stch_snw_alb(i-iLL +1 ,j - jLL +1) .le. 1.0) then
```
gmao-rreichle, line 3081: Shouldn't this subroutine also assume, just like MODIS_snow_alb(), that the snow albedo on the raster grid has been filled?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
```diff
@@ -2991,9 +2991,192 @@ SUBROUTINE hres_gswp2 (nc_data,nr_data,rmap, gfiler,lai_name,merge)
...
+                   endif
+                 endif ! if not missing
+             enddo ! i-loop
+          enddo ! j-loop
+
+          ! In order to reduce the time taken by the gap filling procedure,
+          ! creating a 1.-degree gridded dataset to use it for filling the gaps seems practical/manageble.
+          !-----------------------------------------------------------------------------------------------
+
+          ! populate global grid/map (e.g. 1x1deg) of snow albedo values to use it later to fill in the gaps
```
gmao-rreichle, line 3104: We should not need to "fill" anything later.  See comment above.

gmao-rreichle:
@biljanaorescanin, I added one commit with some very minor cleanup, but I'm not convinced the new subroutine MODIS_snow_alb_tileid() is ready.  See inline comments.  Perhaps most importantly, I don't understand why MODIS_snow_abl_tileid() looks so different from MODIS_snow_alb() in virtually all aspects.  The only difference between the two should be how the MODIS snow albedo is mapped from the raster grid to the tiles.  Reading the snow albedo on the raster grid should not look different between the two subroutines.  Also, in MODIS_snow_alb() the snow albedo is read into a 4d array that effectively holds the global map.  Not sure why there's a 4d array when a 2d global map would have been simpler (given that there seems to be enough memory to store the whole thing at once).  This could probably be simplified but was missed during the earlier PR (#618).  The logic of the new subroutine MODIS_snow_alb_tileid() still escapes me.  I might be missing something.  
cc: @rdkoster

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
```diff
@@ -2991,9 +2991,192 @@ SUBROUTINE hres_gswp2 (nc_data,nr_data,rmap, gfiler,lai_name,merge)
...
+  open (10,file=fname,status='old',action='read',form='formatted')
+  read(10,*) N_tile               ! # of tiles 
+  allocate (tile_lon(1:N_tile))  
+  allocate (tile_lat(1:N_tile))
+
+  do n = 1, N_tile                ! catchment.def provides min/max lat/lon. Find lat/lon of the center of a tile
+     read (10,*) tindex1,pfaf1,minlon,maxlon,minlat,maxlat
+     tile_lon(n) = (minlon + maxlon)/2.
+     tile_lat(n) = (minlat + maxlat)/2.
+  end do
```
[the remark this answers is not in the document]
biljanaorescanin: Thanks. 
The lat/lon is used only when checking for any missing values and populating by the nearest neighbor value. Why is this used? It simply follows what’s done in other modules (eg hres_gswp2) when reading similar variables.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/mod_process_hres_data.F90
biljanaorescanin: Here the ‘fill’ refers to having a tile that fell in between two consecutive raster grids points. If tile space is at finer resolution than 30arcsec, we may get a tile with no snow value assigned to it, right?

biljanaorescanin:
The MODIS_snow_alb_tileid() follows the method used in hres_gswp2(). I am not sure who created that one. It’s logic was not clear to me neither, and this is why I created the MODIS_snow_alb() in first place. You are correct that the global map in MODIS_snow_alb() could be a 2d instead of 4d, but the size (i.e. memory usage) is the same, and having it as a 4d array eliminates room for an error when reading in the data (I don’t have to concatenate data along two dimensions). If in 2d, the jLL and iLL offsets must be calculated. If 4d, the hh and vv indices must be calculated. It’s the same level of complexity, except in the 4d case I do not have to manipulate the data when reading in.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90
```diff
@@ -3001,9 +3001,192 @@ SUBROUTINE hres_gswp2 (nc_data,nr_data,rmap, gfiler,lai_name,merge)
...
+
+          ! populate global grid/map (e.g. 1x1deg) of snow albedo values to use it later to fill in the gaps
+          do j = ceiling(1.*jLL/QSize),ceiling(1.*jLL/QSize) -1 + nr_10/QSize    
+             do i = ceiling(1.*iLL/QSize),ceiling(1.*iLL/QSize) -1 + nc_10/QSize
+                ! take a subset of indecies
+                QSub  => stch_snw_alb((i-1)*QSize+2-iLL :i*QSize-iLL+1, (j-1)*QSize+2-jLL :j*QSize-jLL+1)
+                ! if the QSub got a .gt. zero value, get the mean (account only non-missing vlaues)
+                if(maxval (QSub) > 0) data_grid(i,j) = sum(QSub, QSub>0)/(max(1,count(QSub>0)))
+             enddo
+          enddo
```
gmao-rreichle, line 3122: @biljanaorescanin: Is this still needed?  Isn't the snow albedo on the input raster grid already gap-filled?  See comments at the top of the subroutine.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90
```diff
@@ -3001,9 +3001,192 @@ SUBROUTINE hres_gswp2 (nc_data,nr_data,rmap, gfiler,lai_name,merge)
...
+     end do ! ix-loop over 10x10deg files
+  end do ! jx-loop over 10x10deg files
+
+  ! calculate mean value out form the sum and count (this is what I am after)
+  where (count_snow_alb > 0.) snw_alb = snw_alb/count_snow_alb 
+
+  ! Filling gaps
+  !-------------
+  DO n =1,N_tile ! loop over all elements of the output array
+      if(count_snow_alb(n)==0.)  then ! if an element is zero, take action
```
gmao-rreichle, line 3137: @biljanaorescanin: Can there be any gaps?  See comment above.  Rather than having code here to fill gaps that shouldn't exist, I think we should just double-check that count>0 for every tile and exit cleanly if this condition is violated.

biljanaorescanin: Very true. With gap-filled snow albedo raster, and the assumption of no runs with resolution finer than 30 arc sec, the gap-filling is unnecessary, and it should be safe to remove this section. A simple check on whether there are any tiles with no snow albedo values is an easy and fast way to check for existing gaps. No gaps are expected; however, a hard stop would be a good option if such a scenario occurs. I will remove the gap-filling sections of the code and run a zero-diff test.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90
gmao-rreichle: > With snow albedo raster at 30 arc sec, we could still start seeing gaps (i.e., tiles with no snow albedo assigned) if the model run is of a resolution finer than 30 arc sec. 

I'm confused about this statement.  The raster grid with the tile info is at 30-arcsec (per the hard-wired processing associated with the 10-arcsec mask file), and the MODIS snow albedo raster is also on the 30-arcsec grid (per the hard-wired MODIS snow albedo preprocessing).  For "gaps" to appear we'd have to have model tiles at a finer resolution than the 30-arcsec raster, which makes little sense because we would just running sets of identical tiles within 30-arcsec raster grid cells.  Gaps could only appear if we change the resolution of the tile ID raster file, but that would require revisiting all of the processing steps in mkCatchParam.F90, not just the snow albedo.  

We may still be missing a proper check that the raster file that is the basis for rmap() is on the same grid as the pre-processed MODIS snow albedo.

gmao-rreichle:
Testing update at commit 2f79132, tested make_bcs for EASEv2_M36 and c180/O1:
- NL3 is 0-diff vs. v17.131
- v10 is 0-diff vs. 0f19286 (most recent by @biljanaorescanin)

Remaining issues:
- The remapping from science data to tile space could use a lot more cleanup.
- It may be sensible to clean up mixed mode arithmetic and quirky calculations at this time, even if this means that the code can no longer reproduce existing bcs versions identically.
- For make_bcs (csh), the SBATCH job summary (wall time etc) is not captured, which makes it very difficult to assess the run-time of the job.  This may or may not be resolved in make_bcs.py.
- make_bcs (csh) creates the bcs in a top-level "resolution" dir, which is not consistent with the new output dir format in bcs_shared/fvInput.  This may or may not be resolved in make_bcs.py.

gmao-rreichle:
Merged branches from two PRs that are expected to be merged before the present PR:
#751: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/687/commits/eabdc668beb0d5f6dfe020572f081d6ed1f21751
#763: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/687/commits/76117e019fd034440a2a8b910315bfd11773d4d3

gmao-rreichle:
@biljanaorescanin, I added three comments to your most recent commit here:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/b8cd62cd2fda01853c74123879f34b74392d2666#comments
I realized afterwards that @weiyuan-jiang added another commit on top, which I think makes my comment about the variable name (catchcn_internal vs. CATCHCN_INTERNAL_STATE) obsolete.  But I think the other two comments still apply.  Note that the comment about the wrong order of the calculations **also** applies to GEOS_CatchCNCLM**45**GridComp.F90.  
At a later time I'll still need to edit the documentation in the Surface rc file.

gmao-rreichle (approved):

biljanaorescanin:
@sdrabenh this PR is now fully tested and it is zero diff.
1. GEOSldas Nightly tests zero diff  
2. Science runs - plots verified at weekly meetings 
3.  (1 day) AMIP, reply, non inc replay all zero diff 
4. regression and regression layout tests both passed

gmao-rreichle:
> @sdrabenh this PR is now fully tested and it is zero diff.
> 
> 1. GEOSldas Nightly tests zero diff
> 2. Science runs - plots verified at weekly meetings
> 3. (1 day) AMIP, reply, non inc replay all zero diff
> 4. regression and regression layout tests both passed

Thanks, @biljanaorescanin.  Is the testing of the make_bcs functionality up-to-date?  That is, are we reproducing bcs (old versions and the new tileID-based snow albedo) as expected?

biljanaorescanin:
I've tested two sets of v06, which are zero different from what Bill used or what I've used during science testing.  
Two sets of v11 are zero different from last month's runs. I don't have runs from before for this set. Bill did not use this set and v11 was not used during long science testing. v11 was validated with a comparison between the v11 and v06.

gmao-rreichle:
> I've tested two sets of v06, ...

Thanks, @biljanaorescanin.   All set for @sdrabenh, then.  Python changes in this PR are trivial.

PR #706, opened 2023-02-17
Boundary Conditions cleanup

biljanaorescanin:
Clean up of Raster directory with additional subdirectories. 
This should make it more clear what is used with boundary conditions package and what is not. 

I've tested it for: ICA C90 O2,NL3 C12 T4MOM6, NL4 M36, NL5 M09 , v06 C180 O2 . 

@gmao-rreichle this is draft for now until you confirm this is what you envisioned for this area.
@weiyuan-jiang you can take a look and let me know if we need to change more. Preprocessing we can edit more depending on where we want to move it.  I still didn't test python wrappers.


gmao-rreichle:
Closing for now.  Replaced by #707.

PR #707, opened 2023-02-21
cleanup and reorganization of files in ./Raster (make_bcs)

gmao-rreichle:
Intended to replace #706.

- new directory "./makebcs/" contains scripts and programs to create bcs
- new directory "./preproc/" captures **some** scripts and programs to create **inputs** to make_bcs (INCOMPLETE!)
- new directory "./misc/" contains potentially still useful debugging and plotting scripts for make_bcs

UPDATED 4 March 2023:
- trivially 0-diff for GCM (PR touches only CatchCN and Raster/make_bcs files)
- make_bcs successfully 0-diff tested by @biljanaorescanin 
- left for another time:  remove duplication of functionality in  ./makebcs/CubedSphere_GridMod.F90 (should be using ESMF)
- I'm happy with the PR as it stands right now but since it's my own I cannot approve it.  @biljanaorescanin, @weiyuan-jiang, @mathomp4, please double-check and approve for the respective teams (Land, Surface Preproc, etc) and move it on to @sdrabenh's desk, or let me know if you have any questions/concerns.  Thanks!


gmao-rreichle:
@biljanaorescanin, @weiyuan-jiang, 

I started over with #706 here for two reasons.  One was that #706 somehow ended up adding "execute" permission to F90 files, presumably because the files were first deleted and then added, rather than moved via "git mv ...".  The branch associated with this new PR (#707) used "git mv..." to move the files.  The other reason was that I wanted to add a new ./misc/ directory that captures some potentially still useful debugging and plotting tools. 

There are a few outstanding issues:

- I did not (yet) add CMakeLists.txt files into the ./preproc/ subdirectories.  While it may be good to compile any F90 code in these directories during the build, I'm not convinced that we want to clutter up the build's ./bin directory with the corresponding ("preproc") executables.  Something to discuss.
- Do we need the file ./makebcs/compile_mk_runoff ?? Shouldn't this be done through CMakeLists.txt?
- ./makebcs/date_time_util.F90 and ./makebcs/leap_year.F90 are predecessors of LDAS_DateTimeMod.F90 in the GEOSldas GridComp.   I'm not sure how best to resolve the duplication. 
- What does ./makebcs/CubedSphere_GridMod.F90 do that isn't in MAPL?  Can't we use MAPL tools instead? 
- In the CMakeLists.txt file(s), what does `esma_set_this (OVERRIDE [dirname?])` do?  This might need fixing. 
- Do we need (all of?) `Raster.h`?   It defines its own pi, among other things, which unfortunately differs from MAPL_PI_R8.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/albedo/snow/get_lat_lon4tils.pro
```diff
@@ -0,0 +1,81 @@
...
+      for ihgrid=0,2400-1 do begin
+
+         ; get lats and lons from UV coordinates (requires the map projection loaded above)
+         xOrigin=-20015109.354d +((ihtil)*2400l+ihgrid)*463.31271653d
+         yOrigin= 10007554.677d -((ivtil)*2400l+ivgrid)*463.31271653d
+         lonlat = MAP_PROJ_INVERSE([xOrigin], [yOrigin], MAP_STRUCTURE=sinusMap)      
+         lat    =lonlat[1]
+         lon    =lonlat[0]
+
+         if lat eq lat then begin
```
gmao-rreichle, line 48: What does this if condition mean?  Isn't "lat eq lat" always true?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/albedo/snow/get_snow_alb_mod10a1_30arcsec.pro
```diff
@@ -0,0 +1,568 @@
...
+
+dim_1d=2400l*2400l
+dim_2d=[2400l,2400l]
+
+; ********* STEPS ***********
+; [un]comment out the following goto's depending on what part of the code is to be executed
+;goto, skip_extracting_snow_cvr   ; *** STEP 2
+;goto, skip_calculating_snow_cvr  ; *** STEP 3
+;goto, skip_reading_snow_alb      ; *** STEP 4
+;goto, skip2stitching             ; *** STEP 5
```
gmao-rreichle, line 89: I don't see "skip2stitching" below.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/albedo/snow/grid.pro
```diff
@@ -0,0 +1,71 @@
+function grid,data,lat,lon,region=rgn,nlat=nlat,nlon=nlon, $
```
gmao-rreichle, line 1: It's not at all obvious what this function does.  I suspect it regrids "data"?  But how?   Could use some high-level introductory comments.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/albedo/snow/get_snow_alb_mod10a1_30arcsec.pro
```diff
@@ -0,0 +1,568 @@
...
+
+      endfor ; iih
+    endfor ; iiv
+
+    ; if there are any valid values, grid them to populate the tile
+    if n_elements(alb_lim_tail_mean_all) gt 0 then begin
+
+      print, 'Start gridding'
+
+      ; grid to 1200x1200 (30 arc sec)
```
gmao-rreichle, line 475: Is this the step where data are mapped from the original MODIS sinusoidal 500-m grid to a regular 30 arcsec (~1 km) lat/lon grid?  How are the data regridded?  Nearest neighbor?  Avg of all 500-m grid cells "within" 30 arcsec grid cell?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/albedo/snow/get_snow_alb_mod10a1_30arcsec.pro
```diff
@@ -0,0 +1,568 @@
+pro get_snow_alb_mod10a1_30arcsec, year=year, h_s=h_s, h_e=h_e
+
+; Code to stitch MODIS files, such as: MOD10A1.A2022002.h35v08.006.2022004050029.hdf
```
gmao-rreichle, line 3: I think the original MODIS inputs are on a sinusoidal 500-m grid, and the final snow albedo product is on a 30 arcsec regular lat/lon grid.   It would help to add this kind of information here.  
Is the output 30 arcsec lat/lon grid DC or DE (date-line-on-center/edge)?  How about PC and PE (pole)?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/albedo/snow/get_lat_lon4tils.pro
biljanaorescanin: It's a safety. A quick way to check if a number (in this case 'lat') is not a NaN. In case xOrigin and/or yOrigin exceed the bounds, the MAP_PROJ_INVERSE can return an NaN.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/albedo/snow/grid.pro
biljanaorescanin: Yes, this function [re]grids data. True, it is not obvious what it does. Will add comments to it.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/albedo/snow/get_snow_alb_mod10a1_30arcsec.pro
biljanaorescanin: The latter. The subroutine accumulates values of the input data (albedo) within each 30 arcsec gridbox, keeps track of the number of the accumulated elements and then calculated the mean.  Now with the comments added to the 'grid.pro' function, this should be easier to understand.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/albedo/snow/get_snow_alb_mod10a1_30arcsec.pro
biljanaorescanin: Correct. This information is now added, as well as the info on DE/PE.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/albedo/snow/get_lat_lon4tils.pro
gmao-rreichle: Thanks.  That strikes me as an insider practice.  I don't use IDL, but a quick search turned up the intrinsic function "FINITE".  Something like the following might be less confusing:
`if finite(lat) then begin`
along with a comment that lat may be NaN, in which case there's nothing to do.
No need to edit now, just a suggestion for future development.

biljanaorescanin:
PR was tested for C720 NLv3 O2, C180 ICA O2, NLv5 M09, C12 NLv3 T4MOM6 all were zero diff.

biljanaorescanin:
@gmao-rreichle should I resolve conversations or will you address that?

weiyuan-jiang (approved):

mathomp4 (approved):
Super approve per @gmao-rreichle

sdrabenh (approved):

PR #710, opened 2023-02-28
IGNI: VPD over land

adarmenov:
IGNI and SURFACE are extended to provide VPD over land.

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOSigni_GridComp/GEOS_IgniGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOSigni_GridComp/GEOS_IgniGridComp.F90
@@ -1474,16 +1474,16 @@ contains
         end if
 
         ! calculate ISI, BUI, FWI and DSR
-        isi(i)  = initial_spread_index(ffmc(i), wind(i))
+        isi(i) = initial_spread_index(ffmc(i), wind(i))
 
-        if (snow_depth(i) > 1e-3) then  
-            isi(i) = max(0.0, (1 - f_snow(i))) * isi(i)
+        if (snow_depth(i) > 1e-3) then
+            isi(i) = (1 - f_snow(i)) * isi(i)
         end if
 
-        bui(i)  = buildup_index(dmc(i), dc(i))
+        bui(i) = buildup_index(dmc(i), dc(i))
 
-        fwi(i)  = fire_weather_index(isi(i), bui(i))
-        dsr(i)  = daily_severity_rating(fwi(i))
+        fwi(i) = fire_weather_index(isi(i), bui(i))
+        dsr(i) = daily_severity_rating(fwi(i))
     end do
 
   end subroutine cffwi_daily_driver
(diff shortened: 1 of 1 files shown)


lcandre2 (commented):
These changes do not impact the LandIce grid comp nor exports. I am fine with them.

sanAkel (approved):
Approving on behalf of @GEOS-ESM/ocean-team.

(Not sure when this [surface GC](https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/develop/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90) will see the ACG so nobody has to go through this insanity of adding exports this way. @tclune any hope?)

lcandre2 (approved):
see previous note.

tclune:
For AGC, just ask @darianboggs with a list of components to update.      He's deep enough in that he has some big tasks on his plate, but adding this to components should be easy. (Or if it is not easy, then it will make us aware of needed extensions in the ACG.)

gmao-rreichle (commented):
Hi @adarmenov, I have a couple of questions:
- Since VPD is a simple function of T2m, Q2m, and PS, do we really need another export?  I suppose VPD could readily be computed from existing output.  But I can also see the convenience, so it's ok with me to have a VPD export variable.  FWIW, the SMAP L4C product uses VPD as an input.
- Some of the lines for VPD and other ignition variables are enclosed in `if (DO_FIRE_DANGER) then` blocks.  Is this needed?  I thought that export variables only generate work and require memory if they appear in HISTORY.rc.   By having the if blocks, VPD isn't available as an export unless DO_FIRE_DANGER is .true., but as mentioned above, VPD may also be useful for ecohydrology.

adarmenov:
I typically create new exports if they are useful diagnostics and/or are used in multiple GCs. Regarding VPD, I was a bit surprised that it is not provided by CATCH(CN), so I took the more safe approach and implemented it in IGNI. IMHO, a better place for VPD over land is CATCH(CN). This should address @gmao-rreichle questions about the current implementation of VPD. For symmetry, it would make sense to calculate VPD in the other surface components.

gmao-rreichle:
@adarmenov, many thanks for the insights, and sorry for throwing a wrench into your PR.  It looks like we agree that it would be better to avoid exporting VPD from the IGNI GridComp.   It does get complicated quickly, though:

- Your VPD calc uses T2m, Q2m, and PS.  To drive the land model, however, we use TLML and QLML (i.e., lowest-model-level variables, a.k.a. TA, QA).  Not sure if VPD should be diagnosed from 2-m or LML variables.
- The Surface GridComp works at the grid cell scale and includes all surface types.  If you're after VPD specifically over land and if 2-m inputs can/should used, then we could (perhaps) calculate VPD in the Catch[CN] GridComp using diagnosed T2m, q2m over land tiles only.  Whether these are available seems to depend on whether Louis or Helfand surface turbulence is used.  (Louis is better for offline land-only simulations and used in SMAP L4SM ops.  All other GEOS systems use Helfand as far as I know.) 
- I didn't see VPD at the GridComp levels of CatchCN.  I imagine VPD is calculated somewhere within the CatchCN/CLM science routines but didn't check. 

Given the above, it's unfortunately not a quick task to calculate VPD in Catch[CN] and/or using 2-m inputs, if that's what we want.  (It shouldn't be too difficult to use LML inputs to calc VPD in Surface.)   This probably needs the attention of the land modeling group (i.e., @rdkoster, @gmao-jkolassa), and we'll probably want to get @amolod's take.  

I guess the foremost question now is how urgently you need VPD output.  If it's not urgent, it may make sense to do take some time and do this right.  If you need VPD quickly and are happy with what you have, I'm ok with merging this PR, but this means that someone will have to clean up later. 

cc: @sdrabenh

adarmenov:
I am content with 'my' land only VPD(2m) - this is by construction :) I will gladly use VPD(2m) from CATCH/CATCH_CN should it be made available from them. 

@gmao-rreichle, I am not sure that I follow your concern about 'cleaning up later'. The current VPD implementation is zero diff and it should not have any side effects. Where VPD is calculated should not matter to the components that import it. Unless I am overlooking something, the 'clean up' steps should be 0) remove calculation of VPD in IGNI, 1) add VPD to the CATCH/CATCHCN->IGNI connectivity, and 2) (in SURFACE) change the CHILD_ID to the GC that provides VPD.

gmao-rreichle:
@adarmenov:  Yes, this is what I meant by "cleanup":

>the 'clean up' steps should be 0) remove calculation of VPD in IGNI, 1) add VPD to the CATCH/CATCHCN->IGNI connectivity, and 2) (in SURFACE) change the CHILD_ID to the GC that provides VPD.

When the VPD calc is eventually added to Catch[CN], the challenge is to remember that VPD is calculated in IGNI and to actually implement the above steps.  Also, switching to calculating VPD in Catch[CN] as opposed to IGNI will probably a a non-zero-diff change for IGNI, which will require science verification (and would thus go beyond the usual scope of SI Team work).

I'll go ahead and add my approval.  With any luck, we'll remember all this and can tackle it sooner rather than later.

gmao-rreichle (approved):
approving for @GEOS-ESM/land-team

sdrabenh (approved):

PR #718, opened 2023-03-09
update legacy bcs path

biljanaorescanin:
This PR updates the path to the legacy bcs data by pointing to the new "bcs_shared" directory in the GMAO project space on Discover (replacing `/discover/nobackup/ltakacs/bcs/` )

This is the first step in using (legacy) bcs from the new "bcs_shared" dir.


gmao-rreichle (approved):
- Approving for @GEOS-ESM/surface-preproc-team   
- @weiyuan-jiang: I noticed that `make_bcs.py` has fallen behind `make_bcs`.  It may be good to catch up the python version before there are too many differences to keep straight.

gmao-rreichle (approved):
re-approving

sdrabenh (approved):

PR #729, opened 2023-03-20
make_bcs output directory reorg

biljanaorescanin:
Work towards new output structure for boundary conditions.
Removing "SMAP" from EASE grid cases. Changes were minimal since Weiyuan already prepared all the codes. 

**Requires first merging of #718.**

cc: @gmao-rreichle @weiyuan-jiang


biljanaorescanin:
I've tested few cases and all were zero diff. ( NLv3 C720, NLv5 M09 and ICA C180 and one LatLon case for ICA. )
@gmao-rreichle let me know if you think I should run for more cases.

gmao-rreichle:
Thanks, @biljanaorescanin.  The changes look ok to me, but I think it would be better to put this branch on top of the branch associated with #718, and point the PR to that branch until #718 has been merged.  Currently, there's some overlap between the PRs.
The present PR is meant to have make_bcs create the bcs files in the new dir tree structure, right?  
If that's the case, then I think we should change the title of the PR to something like "make_bcs **output** directory reorg", and maybe change the title of #718 to something like "update make_bcs legacy **input** path.

biljanaorescanin:
@weiyuan-jiang we still have conflicts. 

We should probably also change this in make_bcs and python codes, after email we all got: 
#SBATCH --ntasks=28
#SBATCH --constraint=sky
to
#SBATCH --constraint="[cas|sky]"  and use $SLURM_CPUS_ON_NODE

weiyuan-jiang:
@biljanaorescanin , Have you tried to submitted many jobs? I have concern over this line 
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/4b33979cd9833e3ded52fafa027e30c50586b951/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs#L945

It appears at lat-lon, cubed-sphere and EASE cases.  If some jobs are submitted simultaneously, this directory may be deleted by the other job that finishes earlier.

weiyuan-jiang:
Now make_bcs.py should be in sync with make_bcs.   The three program make_latlon_bcs.py, make_cube_bcs.py and make_ease_bcs.py can be tested independently

biljanaorescanin:
> @biljanaorescanin , Have you tried to submitted many jobs? I have concern over this line
> 
> https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/4b33979cd9833e3ded52fafa027e30c50586b951/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs#L945
> 
> It appears at lat-lon, cubed-sphere and EASE cases. If some jobs are submitted simultaneously, this directory may be deleted by the other job that finishes earlier.

I would have to try again. I did run before and there was no issues. But it has been a while so I don't know for sure.

biljanaorescanin:
Looks like running CF and LatLon cases is OK in all combinations, for many resolutions at the time, but any experiment with EASE has to be run separately just EASE. 
This part we added for version: 
                                                     Options for EASE grid version (land-only):
                                                     EASEv1
                                                     EASEv2 (default)

Gets appended anytime we have EASE in combination... so I get for example this mess if I do:
` Land BCs version:       ICA`
 `Atmospheric resolution: b c12 M36`
 `EASE grid version:      EASEv2`
 `Ocean resolution:       O1`
SMAP_EASEv2_M36.j
SMAP_EASEv2_c12.j
SMAP_EASEv2_b.j

I see during testing I didn't run many combinations at the time but individual and if I did test many cases they were in same case type CF, EASE or LatLon.  So that OUTDIR you asked about is not the problem but EASE grid version being appended.

gmao-rreichle:
> Looks like running CF and LatLon cases is OK in all combinations, for many resolutions at the time, but any experiment with EASE has to be run separately just EASE. 

@biljanaorescanin, @weiyuan-jiang: For `make_bcs` **only**, maybe we can just add augment the following comment:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/4b33979cd9833e3ded52fafa027e30c50586b951/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs#L182

with something like "Do NOT specify a mix EASE grid and cs/LatLon grid resolutions."

This should be enough of a fix for `make_bcs` until it is deprecated.  I assume the same problem does not occur in the python version of make_bcs.

Does this make sense?

biljanaorescanin:
> > Looks like running CF and LatLon cases is OK in all combinations, for many resolutions at the time, but any experiment with EASE has to be run separately just EASE.
> 
> I still think we should remove that line. The OUTDIR is shared across all the jobs. You got lucky if the jobs are executed sequentially.

We should then remove it for python. Let me know when it is ready so I can start testing.

weiyuan-jiang:
> 

I think it is ready for the tests now.  At the same time, I want to change some names like BCNAME --> GRIDNAMES,  but it will not affect the testing.

biljanaorescanin:
@gmao-rreichle this is ready for you to take a look. I've run tests over all options, with multi jobs and single jobs and all are zero diff to make_bcs output.

gmao-rreichle:
@biljanaorescanin, @weiyuan-jiang, I added a couple of commits with minor fixes.  The first commit fixes the spelling of "questionnaire".  If you prefer, you can change to "questionary" (which happens to be the term used by python).   The second commit is minor cleanup.
But I'm not convinced that this PR is ready for prime time as far as the python utilities are concerned.  Specifically, I'm surprised that the various make_bcs_[GRID].py scripts look so different.  That is, I would have expected a cleaner set of differences when doing something like:
`xxdiff make_bcs_cube.py make_bcs_latlon.py`
E.g., the scripts all seem to use different variable names for, say, the SBATCH job name.  Another example is that the "source g5_modules" is done at somewhat different places.  More symmetry across these helper scripts would make future maintenance easier. 
Also, is it possible to better abstract common elements such as the moving of the output files into the new dir tree structure?  I recognize that there will probably always be some differences, but the current chunks of shell script commands ("mv ...", "cd ...") is still as messy as it was in make_bcs.

weiyuan-jiang:
> 

I will make it cleaner

weiyuan-jiang:
The draft PR #740 tries to make it cleaner.

gmao-rreichle:
@biljanaorescanin, please go ahead and start the tests.  I will look at the most recent edits one more time while testing is going on.  Chances are I would only have minimal changes that wouldn't require full retesting.

biljanaorescanin:
@weiyuan-jiang @gmao-rreichle @sdrabenh same error like in other PR's unrelated to our changes: 

`Character Resource Parameter: DU_INTERNAL_RESTART_FILE:du_internal_rst
 Bootstrapping du_internal_rst
forrtl: severe (194): Run-Time Check Failure. The variable 'ss2g_gridcompmod_mp_initialize_$FILE_EXISTS' is being used in '/root/project/GEOSgcm/src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/@GEOSchem_GridComp/@GOCART/ESMF/GOCART2G_GridComp/SS2G_GridComp/SS2G_GridCompMod.F90(378,9)' without being defined
GEOSgcm.x          0000000000FE9461  ss2g_gridcompmod_         378  SS2G_GridCompMod.F90`

biljanaorescanin:
@gmao-rreichle  and @weiyuan-jiang I've tested most often used cases for NLv3 [C180 C720] O2, NL3 MOM6 C180, ICA C180 O2, NLv4 M36 and NLv5 M09 and v06 C180 O2 all were zero diff when compared to what is on develop.

gmao-rreichle (approved):
approving for surface-preproc-team

tclune (dismissed):
CMake changes ok.

Glob be dangerous ...

gmao-rreichle (approved):

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/CMakeLists.txt
```diff
@@ -47,3 +47,5 @@ ecbuild_add_executable (TARGET mk_runofftbl.x SOURCES mk_runofftbl.F90 LIBS MAPL
 ecbuild_add_executable (TARGET mkEASETilesParam.x SOURCES mkEASETilesParam.F90 LIBS MAPL ${this})
 
 install(PROGRAMS make_bcs clsm_plots.pro create_README.csh  DESTINATION bin)
+file(GLOB MAKE_BCS_PYTHON "./make_bcs*.py")
```
mathomp4, line 50:
```suggestion
file(GLOB MAKE_BCS_PYTHON CONFIGURE_DEPENDS "./make_bcs*.py")
```

Ooh. @tclune is right. We want to add `CONFIGURE_DEPENDS` here. It's not a perfect solution, but without this option, if someone edited a python script, CMake would not know that the file changed. With the option, CMake will take the time to look and see if a file has changed.

gmao-rreichle (approved):
I accepted & merged Matt's suggestion and re-approved the PR

gmao-rreichle (approved):

sdrabenh (approved):

PR #637, opened 2022-09-11
R21C HISTORY output specs (precip, land, landice, ...)

gmao-rreichle:
This PR proposes updated R21C HISTORY specs on top of @sdrabenh's R21C branch.
See also [matching PR #356 in GEOSgcm_App](https://github.com/GEOS-ESM/GEOSgcm_App/pull/356)
These updates will eventually need to find their way into the default and/or FP file specs in the 'develop' branch, which will require another PR.

----
**Details:**

Updates are focused on (but not entirely limited to) land and landice outputs.

For land outputs in the "lnd" collection, appended/inserted "LAND" in LONG_NAME wherever there's any chance at all that users could be confused that the output is only over land tiles.  

Clarified LONG_NAME for volumetric soil moisture (m3/m3) vs dimensionless soil wetness.

Made variables that are exported in R21C consistent across land and landice (incl. runoff).

Fixed sign issue for GHTSKIN in landice.

Still to do:

1. **Figure out how to deal with LONG_NAME of model vs. corrected precipitation.**  In MERRA-2, the words "bias corrected" were appended somehow to LONG_NAME for PRECTOTLAND and PRECSNOLAND. 
a. **"Bias corrected" is not a good term**; going forward, we should use "observation corrected" because the corrections are applied at the daily scale.
b. Adding "observation corrected" to the LONG_NAME in *GridComp.F90 files requires thought because **currently the F90 files supports both uncorrected and corrected precip over land in the same export variable.**

2. Add runoff (and other variables?) averaged across land and landice.  [**DONE** @lcandre2, @gmao-rreichle 16Nov22]

3. Resolve conflicting signs for GHTSKIN (GHLAND?) in landice and land collections.  [**DONE** @lcandre2 16Nov22]

4. Resolve assorted in-line comments (to be added). [**DONE** @gmao-rreichle 17Nov22]

I'll keep working on this, but please feel free take a look and add your 2c.  

For precip outputs, figure out how to clarify model vs. corrected in the LONG_NAME. 

cc: @elakkraoui @lcandre2 @gmao-qliu @lltakacs @sdrabenh @rcullath @rdkoster @rlucches

PR #735, opened 2023-04-05
Pull in changes from feature/sdrabenh/r21c_G2G

sdrabenh:


sanAkel (dismissed):

zhaobin74 (approved):

gmao-rreichle:
@sdrabenh, I'm afraid there's an issue with CatchCN in your branch.  After a very quick look, here's what I think happened.  I might be wrong, but what I'm seeing is concerning enough to throw a flag.  
In a nutshell, #637 was on a rather old branch of the GCM GC that still had only one GC F90 file for CatchCNCLM40 and CatchCNCLM45.  At some time between then and now, CatchCN was restructured, and lot of the code in the CatchCN GC file moved into subdirs and GC F90 files specific to CatchCNCLM40 and CatchCNCLM45.  The problem is that when you went from your branch `feature/rreichle/r21c_updates` (associated with #637) to `feature/sdrabenh/r21c_G2G` (associated with the present PR), the CN-related changes from #637 got lost. 
We'll need to take a closer look and, if I'm not mistaken, figure out the best way to resurrect the changes of #637 that fell through the cracks.  For now, I'm holding off on hitting "approve" until I'm more confident of what I'm seeing and how to address the problem. 
cc: @biljanaorescanin @gmao-jkolassa

sdrabenh:
> @sdrabenh, I'm afraid there's an issue with CatchCN in your branch. After a very quick look, here's what I think happened. I might be wrong, but what I'm seeing is concerning enough to throw a flag. In a nutshell, #637 was on a rather old branch of the GCM GC that still had only one GC F90 file for CatchCNCLM40 and CatchCNCLM45. At some time between then and now, CatchCN was restructured, and lot of the code in the CatchCN GC file moved into subdirs and GC F90 files specific to CatchCNCLM40 and CatchCNCLM45. The problem is that when you went from your branch `feature/rreichle/r21c_updates` (associated with #637) to `feature/sdrabenh/r21c_G2G` (associated with the present PR), the CN-related changes from #637 got lost. We'll need to take a closer look and, if I'm not mistaken, figure out the best way to resurrect the changes of #637 that fell through the cracks. For now, I'm holding off on hitting "approve" until I'm more confident of what I'm seeing and how to address the problem. cc: @biljanaorescanin @gmao-jkolassa

I believe that is exactly what happened. The joys of applying commits after refactoring code! I'm afraid ensuring all the changes in #637 make it into this branch will need to be done by hand. I think all automated tools will fail at this.

gmao-rreichle:
> I believe that is exactly what happened. The joys of applying commits after refactoring code! I'm afraid ensuring all the changes in #637 make it into this branch will need to be done by hand. I think all automated tools will fail at this.

Thanks,  @sdrabenh, for confirming my suspicion.  Is the target branch `release/v1` meant for R21C?  If so, we could conceivably forget about the CatchCN updates because presumably R21C won't need CatchCN.  The real question is how the R21C History updates will find their way into the `develop` branch.  That's where we need to make sure we got everything, incl the missing CatchCN updates.  Are there plans to capture the R21C History changes in develop? 

cc: @biljanaorescanin @gmao-jkolassa

sdrabenh:
@gmao-rreichle My latest commit attempts to incorporate all the LONG_NAME changes into the CatchCN components so the changes do not get left behind, even if they are not used in R21C. Please review if these look ok now

sanAkel (approved):
We should really have an automatic code generator for such changes.

gmao-rreichle (approved):
@sdrabenh: Many thanks again for fixing up the CatchCN40 and CatchCN45 GridComp files.  I added one commit https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/735/commits/01799223f811c99b40c90a4ced0336791bf775b2 that updates the runoff longname, which had somehow fallen through the cracks.  Looks good to me know.

mathomp4 (approved):

PR #740, opened 2023-04-19
Feature/borescan restructure bcs script

weiyuan-jiang:
This PR would simply the creation of job scripts and also make scripts cleaner


weiyuan-jiang:
It has been tested. The files are moved to the right directory.

PR #751, opened 2023-05-16
add CATCHMENT_SPINUP mode (remove snow every Aug 1 in NH or Feb 1 in SH)

gmao-rreichle:
Adds a resource variable (`CATCHMENT_SPINUP`) to control removal of snow on land during offline Catch[CN] spinup simulations.

If `CATCHMENT_SPINUP/=0`, snow mass, depth, and heat content on land tiles are zeroed out every Aug 1 at 0z in the Northern Hemisphere and every Feb 1 at 0z in the Southern Hemisphere.

Also introduces new types T_CATCH_STATE and T_CATCHCN_STATE to bundle Catch[CN]-related rc variables.

Additional minor edits:
- clarified comments about `CATCHMENT_OFFLINE` resource variable
- fixed typos in comments

Note related [**GEOSldas PR#647**](https://github.com/GEOS-ESM/GEOSldas/pull/647).

UPDATE by @gmao-rreichle on 12 July 2023:
Tested successfully:
- 0-diff tests for AGCM
- 0-diff tests for GEOSldas
- new functionality (GEOSldas) 

See also [GEOSldas PR#647](https://github.com/GEOS-ESM/GEOSldas/pull/647).


GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
```diff
@@ -207,6 +209,12 @@ subroutine SetServices ( GC, RC )
     VERIFY_(STATUS)
     wrap%ptr%CATCH_OFFLINE = OFFLINE_MODE
 
+    ! CATCHMENT_SPINUP mode (use integer to leave room for additional spinup modes in future)
+    !    0: DEFAULT 
+    !    1: remove snow every Aug 1 (Northern Hemisphere) or Feb 1 (Southern Hemisphere)
+    call MAPL_GetResource ( MAPL, CATCH_SPINUP, Label="CATCHMENT_SPINUP:",  DEFAULT=0, RC=STATUS)
```
gmao-rreichle, line 215: @weiyuan-jiang:  I'm not sure if SetServices() is the right place to read this resource variable, based on Purnendu's comment a few lines earlier.  But then again, there are several calls to MAPL_GetResource() just below.

Also, should the source object be "MAPL" or "SCF"?

weiyuan-jiang: Some variables should be in setService because it is used in SetService , like offline_mode. Some variable can be moved to Initialization section

gmao-rreichle: Thanks, @weiyuan-jiang.  Do I understand you correctly that all of the following should be moved to Initialize() because they are not used in SetServices()? 

CATCH_SPINUP
NUM_LDAS_ENSEMBLE
SURFRC, SCF
SURFLAY 
Z0_FORMULATION
USE_ASCATZ0, 
CHOOSEMOSFC
USE_FWET_FOR_RUNOFF
FWETC
FWETL
AEROSOL_DEPOSITION

That is, only following variables are needed in SetServices():
OFFLINE_MODE
SNOW_ALBEDO_INFO
N_CONST_LAND4SNWALB

I suppose we can add this cleanup to the PR.  It'll make the PR look a lot more complicated than what we would have if we only added the Catch spinup mode, but cleaning up in this way is progress.  And you already committed another bit of cleanup. 
If the above makes sense to you, can you please implement it?  Thanks!

weiyuan-jiang:
The last commit addresses issue https://github.com/GEOS-ESM/GEOSldas/issues/605#issuecomment-1561554899

gmao-rreichle:
> The last commit addresses issue [GEOS-ESM/GEOSldas#605 (comment)](https://github.com/GEOS-ESM/GEOSldas/issues/605#issuecomment-1561554899)

Thanks, @weiyuan-jiang.  Maybe let's not add more cleanup that is unrelated to the main purpose of this PR (Catch spinup mode).  The SetServices() / Initialize() list of variables cleanup is related to the PR because the PR necessarily touches these subroutines.  But the EASE grid fix is in make_bcs and unrelated to the main topic of the PR.  We should start a new branch & PR for additional make_bcs cleanup.

weiyuan-jiang:
> 

The last 2 commits were moved to a new PR #763

biljanaorescanin:
@gmao-rreichle let us know if this is now good to be implemented to CatchCN?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
```diff
@@ -2820,19 +2814,18 @@ subroutine Initialize ( GC, IMPORT, EXPORT, CLOCK, RC )
...
 
-    call ESMF_UserCompSetInternalState ( GC, 'CATCH_STATE',wrap2,status )
+    call ESMF_UserCompGetInternalState ( GC, 'CatchInternal',wrap,status )
     VERIFY_(STATUS)
+    CATCH_INTERNAL_STATE => wrap%ptr
 
+    ! CATCHMENT_SPINUP mode (use integer to leave room for additional spinup modes in future)
+    !    0: DEFAULT 
+    !    1: remove snow every Aug 1 (Northern Hemisphere) or Feb 1 (Southern Hemisphere)
+    call MAPL_GetResource ( MAPL, CATCH_INTERNAL_STATE%CATCH_SPINUP, Label="CATCHMENT_SPINUP:",  DEFAULT=0, RC=STATUS)
```
gmao-rreichle, line 2827: @weiyuan-jiang, why is the MAPL_GetResource() call for CATCH_SPINUP in Initialize() when all the other catch_internal_state variables are done in SetServices()?   Should this be moved to SetServices() for consistency?

gmao-rreichle:
Thanks for unifying the Catch and CatchCN internal/wrap resource variables, @weiyuan-jiang.  This is neat!  
My latest commit adds the new CATCH_SPINUP code to CatchCN.  This was still missing.  In the process, I fixed a couple of things in the new `catch_wrap_state.F90` and removed all comments explaining the resource variables.  They were present for only a subset of the variables and not consistent with the documentation in `GEOS_SurfaceGridComp.rc` (which I also updated).  
We still need to reconcile the variables listed in `catch_wrap_state.F90` and `GEOS_SurfaceGridComp.rc`.  The latter contains LAND_PARAMS, RRM_DT, and PRINTRC.  Not sure which, if any, of these should be added to T_CATCH_STATE.
Also, I'm not sure I got the MAPL_GetResource() calls right.  The latest commit might not compile or run.  When you get a chance, please take a look and let me know if you see anything obviously wrong.  I recommend ignoring white-space differences in your favorite comparison tool.

gmao-rreichle:
Thanks for fixing the uninitialized variables, @weiyuan-jiang.  Re. the LAND_PARAMS rc variable, it is read here:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/e88a9b88fa129a278650c6df931afca87289de85/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90#L238
The same code block reads several other rc variables (ATM_CO2, CHOOSEMOSFC, etc), which are also in `catch_wrap_state.F90`.  Do we need more cleanup here?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
```diff
@@ -194,56 +163,34 @@ subroutine SetServices ( GC, RC )
 ! pchakrab: Read CATCHMENT_OFFLINE from resource file and save
 ! it in the private internal state of the GridComp. It is a little
 ! unusual to read resource file in SetServices, but we need to know
-! at this stage where we are running Catch in the offline mode or not
+! at this stage whether we are running Catch in the offline mode or not
 
-    allocate(internal, stat=status)
+    allocate(CATCH_INTERNAL_STATE, stat=status)
```
gmao-rreichle, line 168: @weiyuan: The pointer CATCH_INTERNAL_STATE introduced here has the same name as the pointer used for the land analysis increments, e.g., here: https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/2c503402fd6b00d4dd26f3116119f1016f84e6a0/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90#L2855  
Is this intentional or potentially a conflict?  It is very important to get this right because none of our usual tests exercises this land increments functionality.

weiyuan-jiang: I did it intentionally, but obviously, it is a bug here. I will fix it

weiyuan-jiang:
> 
I am wondering move this reading up to the SurfcaeGridComp level

gmao-rreichle:
> [...[. Re. the LAND_PARAMS rc variable, it is read here:
> [..]
> The same code block reads several other rc variables (ATM_CO2, CHOOSEMOSFC, etc), which are also in `catch_wrap_state.F90`. Do we need more cleanup here?

@weiyuan-jiang:   I confirmed that `LAND_PARAMS` is at the Surface GC level because it is needed to set `WEMIN` and a couple of other constants, which are snow model variables (exchanged via the SurfParams module), and the snow model lives in Surface GC ./Shared because it's shared between land and landice.
Like `LAND_PARAMS`, the rc variable `CHOOSEMOSFC` is at the level of the Surface GC because it applies to all surface types (incl. land, landice, etc).  
Unlike `LAND_PARAMS`, however, `CHOOSEMOSFC` is in `T_CATCH_STATE` and read from SCF inside `catch_wrap_state.F90`.
I think the reason for this difference between `LAND_PARAMS` and `CHOOSEMOSFC` is as follows: 
The rc variable `LAND_PARAMS` is **only** needed at the level of the Surface GC, where it's converted into a bunch of Catch[CN] parameter values (such as `WEMIN`), which are then passed down via the SurfParams module.
In contrast, the rc variable `CHOOSEMOSFC` is used at the Catch[CN] GC level. 
I suppose for completeness we could treat `LAND_PARAMS` just like `CHOOSEMOSFC` (i.e., add to `T_CATCH_STATE` and read from SCF in `catch_wrap_state.F90`.  But it doesn't really add anything.  
I don't think we need to worry about `RRM_DT` and `PRINTRC`.
So I'm ok with the branch as is. 
Thoughts?

weiyuan-jiang:
> 

I am ok too.

biljanaorescanin:
**Experiment summary for GEOSldas offline runs:**

Eight experiments were run offline for one day each to test snow removal during offline catch and catchCN runs. 
1. 2 tests done for catch are zero diffs to what @gmao-rreichle tested for August (May 26, 2023 email conversation). 
    The same conclusions were obtained for February with this branch. One difference is in the max SNOMAS value, it is lower in February.
2. 4 tests for CatchCN CLM4.5 were run and analyzed in the same way by looking at SNOMAS values for February and 
     August, with and without a flag on. 
     For the first 1-hr average (0-1z), SNOMAS values are non-zero and non-negligible in the simulation with `spinup_mode=0`
      and in case on `spinup_mode  =1`, values are still non-zero but much smaller and in a similar range as they were in the 
      catch model comparisons. 
     SNOMAS max goes up to **~9** for `spinup_mode =0` and up to **~3** for `spinup_mode =1` for February, South  
      Hemisphere only. And for August, North Hemisphere, SNOMAS max goes up to **~2000** for `spinup_mode  =0` and 
      up to  **~2**  with flag turned on.

The number of tiles with minimum threshold SWE value (0 mm) for identifying snow drastically lowers if the flag is turned on, for both land model versions, a feature intended with this PR. 

**Experiment summary for GCM:**

Just 1-day free run AMIP and it is zero diff.

biljanaorescanin (approved):

gmao-rreichle:
@mathomp4, @sdrabenh: This is ready and @biljanaorescanin approved it, but her approval did not count towards approval for the @GEOS-ESM/land-team.  I can't approve anything because I created the PR.  Oh well.  Please super-approve, thanks.

sdrabenh (approved):

PR #757, opened 2023-05-22
Bugfix for vegetation type (ity) 

weiyuan-jiang:
Ignore variable ity when reading from restarts file. 

More description: During run in debug mode remap_restarts.py gives error while working on catch restart.


weiyuan-jiang:
> 

It reads off a record at the position of ity.

mathomp4:
> It reads off a record at the position of ity.

Indeed, it took me a while to figure out what this commit:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/757/commits/3da9f93f0f1ce253ca75d1cee92254e85fe3a8ed

was doing. But I realized, it is still doing a record read, but it's not going into anything.

@weiyuan-jiang Maybe it would be good to make a big fat comment in the code explaining that you are doing an "empty" read to move past that record. (I'm sure @tclune knows the right Fortran words for this. 😄 )

gmao-rreichle:
Thanks, @weiyuan-jiang and @mathomp4, for the clarification.  When I looked at this briefly yesterday, somehow I saw the "!" at the front of the line.  Glad to know it's all good.  So this concern has been addressed.
@weiyuan-jiang, @biljanaorescanin, please add a comment if this was tested for use with `remap_restart.py` and/or `regrid.pl`.  I think the PR was needed because you ran into a problem with remapping restarts in the first place, so I assume it's ok, but it's still better to say so explicitly.  Thanks!

biljanaorescanin:
@gmao-rreichle , we will have to resolve few more vars in similar fashion since now we have this error. 

"forrtl: severe (408): fort: (8): Attempt to fetch from allocatable variable BF1 when it is not allocated" 

@weiyuan-jiang is working on it.

biljanaorescanin:
Testing summary:
PR now runs in debug mode. PR functionality was tested and it is 0-diff between remapped restarts before and after the PR if both done with optimization.  If we compare catch restart from debug mode to one from none debug run difference is roundoff. (5/58 have roundoff differences)

mathomp4:
> Approving for @GEOS-ESM/surface-preproc-team. My final commit [fbc9b80](https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/fbc9b80aa6d3eba95e17d0dc361c298e6eb61cc6) is just white-space changes to improve readability; tests do **not** need to be repeated.

Well that's odd. It still says it needs preproc team approval even though you are a member of that team.

Oh well, it has enough approvals to be merged in when it's time.

sdrabenh (approved):

PR #763, opened 2023-05-30
Cleanup of make_bcs tools that generate raster and tile files

weiyuan-jiang:
Cleanup of make_bcs tools that generate raster and tile files.

- Addresses https://github.com/GEOS-ESM/GEOSldas/issues/605.  

- **Major** cleanup of `mkEASETilesParam.F90`, which was necessary to understand what the program is doing and how to possibly make it work for the high-resolution M01 (1-km) EASE grid.
During cleanup, a couple of concerning issues were discovered:
  - The min/max lat/lon values for each tile in the catchment.def file seem to be associated with the EASE grid cell and **not** the desired lat/lon boundaries of the tile.
  - The Pfafstetter (catchment) ID assigned to a tile appears to be that of the raster grid cell that is the last to contribute to the tile; it does **not** appear to be that of the dominant catchment within the EASE grid cell. 

- Some cleanup of `mkLandRaster.F90`.

Make sure to hide white-space changes when viewing differences.


GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
```diff
@@ -136,12 +136,20 @@ PROGRAM mkEASETilesParam
       gfile = trim(EASElabel)//'_'//trim(nc_string)//'x'//trim(nr_string)
 
       if (index(EASELabel,'M03') /=0) then ! EASE  3 km grid
+        if (index(EASELabel, 'v1') /=0) then
+           print*, " EASEv1_M03 is not implemented"
```
biljanaorescanin, line 140: Why is M03 not implemented? We can run this one, in past actually I know one person who used it.  @gmao-rreichle did we want to remove this as an option? 
For M01 I agree we should not use until we figure out how to resolve time out and memory issues.

weiyuan-jiang: It was me who combined the v1 and v2 codes. I found  M03 and M01 were not implemented in the original codes. So I removed it.  I need someone to tell me how to set the three numbers in that block if it is v1.

gmao-rreichle: Per https://github.com/GEOS-ESM/GEOSldas/issues/605, I wasn't suggesting that EASEv1 and EASEv3 necessarily be removed.  Rather, I noticed the inconsistency between the if condition and the comment in the following line:
`if (index(EASELabel,'M03') /=0) then ! EASE  1 km grid`
and was asking if there may be a bug.  The contents of the if block suggest that the comment ("EASE 1 km grid") is correct and that the if-condition should read 'M01' and not 'M03'.  But I'm not sure and had left this for closer inspection at a later time.  
The fact that we cannot run M01 bcs owing to resource limitations has nothing to do with my original comment.

gmao-rreichle: Sorry, @weiyuan-jiang, our comments crossed paths. You can find the Nx (cols?) and Ny (rows?) numbers for EASEv1 in the following subroutine:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/934e977fa31b048ae4d00277fa0eb4adb707ee9b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/EASE_conv.F90#L472

weiyuan-jiang: I think NR, NC and NT are different numbers. It cannot be obtained from that call.  I removed it just to be on the safe side. But they may be some arbitrary numbers and the conditions can be removed.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
```diff
@@ -178,7 +186,8 @@ PROGRAM mkEASETilesParam
       !else
       !   if((trim(MGRID) == 'M09').or.(trim(MGRID) == 'M36'))call write_tilfile 
       !endif
-      
+
+     ! Only MASK file we use with EASE grid in bcs package is GEOS5_10arcsec_mask, all other are not supported.  
       if (index(MaskFile,'GEOS5_10arcsec_mask') /= 0) then
```
weiyuan-jiang, line 191: We can move this line to the very beginning , abort if the maskfile is not right

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
```diff
@@ -668,7 +678,9 @@ PROGRAM mkEASETilesParam
          if (index(MaskFile,'GEOS5_10arcsec_mask') /= 0) then
             pfaf = cindex
          else
-            pfaf = catid(all_id(l)-ND_raster*(all_id(l)/ND_raster),all_id(l)/ND_raster)
+            print *,'Only mask file we support with EASE grid is GEOS5_10arcsec_mask'
+            call MPI_Abort(MPI_COMM_WORLD,error_code,status)
```
weiyuan-jiang, line 682: This is not an MPI program.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
```diff
@@ -28,7 +28,7 @@ PROGRAM mkEASETilesParam
       
       implicit none
       integer i,j,ig,jg,i0,iop,n,d1,d2,j1,j2,i1,i2,ix, jx,icount,pcount
-      integer :: NC = i_raster, NR = j_raster, NT = 16330000, ND = 10000, ND_raster = 10000
+      integer ::  NC, NR, NT, ND = 10000, ND_raster = 10000
```
weiyuan-jiang, line 31: ND and ND_raster are redefined later on. Not necessary here.

     ND = 10*10**(nint(log10(1.*nr_ease)))
     ND_raster = 10*10**(nint(log10(1.*NR)))

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
```diff
@@ -135,18 +135,26 @@ PROGRAM mkEASETilesParam
       write(nr_string, '(i0)') nr_ease
       gfile = trim(EASElabel)//'_'//trim(nc_string)//'x'//trim(nr_string)
 
-      if (index(EASELabel,'M03') /=0) then ! EASE  3 km grid
+      if (index(EASELabel,'M01') /=0) then ! EASE  1 km grid
          regrid = .true.
```
weiyuan-jiang, line 139: is there case regrid = .false. ? If no, pull it out

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
```diff
@@ -315,155 +324,156 @@ PROGRAM mkEASETilesParam
...
-                     ice_id  (l) = 1
-                  else
-                     land_id (l) = 1
-                  endif
-               endif
-            end do
-         end do
-
-      endif
+         print *,'Only mask file we support with EASE grid is GEOS5_10arcsec_mask'
```
weiyuan-jiang, line 327: If we check the mask file at the beginning, we can remove all the "else" block here

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
```diff
@@ -668,7 +678,8 @@ PROGRAM mkEASETilesParam
          if (index(MaskFile,'GEOS5_10arcsec_mask') /= 0) then
```
weiyuan-jiang, line 678: if we check maskfile at the beginning, only  pfaf = cindex  is necessary

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
gmao-rreichle: I left it in place here.  If in the future we'll have a new mask file, we'll probably want to have the code that processes it within this program.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
gmao-rreichle: fixed

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
gmao-rreichle: fixed

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
gmao-rreichle: fixed.

biljanaorescanin:
I've moved one set of EASEv2_M36 to the area. I've run test with both make_bcs and make_bcs.py and result is the same between the two and is zero diff to what we have on develop. 

> During testing, it would have been helpful to have the EASEv2_M36 output from the develop branch in the ./bcs_shared/fvInput/ directory. @biljanaorescanin, please add this case. We'll then also need to verify _for the EASE case_ that
> * all files (e.g., catchment.def, job script file, etc) end up in the right places
> * all print statements are captured in the log file(s)

gmao-rreichle:
Update on earlier comments:

> I don't think that the science is quite what it should be and what we all thought it was. [...] 
>   * The min/max lat/lon values for each tile in the catchment.def file seem to be associated with the EASE grid cell and not the desired lat/lon boundaries of the tile.
>   * The Pfafstetter (catchment) ID assigned to a tile appears to be that of the raster grid cell that is the last to contribute to the tile; it does not appear to be that of the dominant catchment within the EASE grid cell.

The latter was confirmed by analyzing existing catchment.def and *.rst files.  Going forward, a fix and new EASE bcs version will be needed under a new PR.  It remains to be seen if the issues apply to the cube-sphere bcs.

> * Based on my now much-improved understanding of the "geometry" files, I think I can confirm that the *.rst file is indeed what is needed to improve the mapping of the MODIS snow albedo into tile space; but I forget where things stand with [use tile_id for snow albedo #687](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/687).

Work is continuing on https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/687

> * [...] have the EASEv2_M36 output from the develop branch in the ./bcs_shared/fvInput/ directory.    
>   * all files (e.g., catchment.def, job script file, etc) end up in the right places
>   * all print statements are captured in the log file(s)

Resolved.

> * The make_bcs *.j script only works when the job is submitted by the `make_bcs` script, because `MAKE_BCS_INPUT_DIR` is not set in the job script.  That is, `sbatch *.j` does not work.  I'm not sure if this is intentional and how this looks in the python version of make_bcs.  (For testing, I used `make_bcs` and not `make_bcs.py`.)

Ok for `make_bcs.py`.  Will not be fixed for `make_bcs` because csh script is becoming obsolete.

> * I'm not sure if the adjustment of the land tile elevation properly accounts for lake and landice surfaces.  This requires further examination.

Still requires further examination.  If there is an issue, it would not be fixed as part of this PR.  Going forward, a fix and new bcs version may be needed under a new PR.

> * I did not clean up the existing cavalier use of single- and double-precision variables.  This will need further work.

Mixed mode arithmetic is unfortunate but (probably) cannot be fixed without non-0-diff changes.  Going forward, a fix and new bcs version may be needed under a new PR.

biljanaorescanin:
This PR only affects boundary conditions creation package;  for AGCM it is zero diff trivial. 
PR was tested for C180 and M36 boundary conditions and its zero diff to what is on develop.

gmao-rreichle (approved):

gmao-rreichle:
@mathomp4:  I approved the PR but for reasons that remain a mystery to me, my approval does not count towards approval by @GEOS-ESM/surface-preproc-team.  The python changes are trivial.  Please super-approve when you get a chance.

biljanaorescanin (approved):

biljanaorescanin:
@gmao-rreichle I can approve for `surface-preproc-team`

mathomp4 (approved):
Approve for @GEOS-ESM/python-transition-team

sdrabenh (approved):

PR #770, opened 2023-06-14
bug fix to correct vegetation fraction assessment in GetIds_carbon (getids.F90)

gmao-jkolassa:
This PR fixes a bug in the subroutine GetIds_carbon in the module mk_restarts_getidsMod. The old code mistakenly used the PFT **type** variable (`sub_ityp1`) to determine whether the vegetation fraction in a tile exceeds the minimum vegetation fraction (`fmin`). The fix changes this so that the PFT **fraction** variable (`sub_fevg1`) is used. 

This bug fix is always zero-diff when using Catchment and is zero-diff when using Catchment-CN with no restart remapping. 

The PR is not zero-diff when using Catchment-CN and restart remapping. The differences are limited to a few locations ( <200 for an EASEv2-M09 test case) and are reasonable in magnitude (but larger than roundoff differences). 

cc: @biljanaorescanin , @gmao-rreichle


gmao-rreichle:
Thanks, @gmao-jkolassa, for discovering and fixing this bug in your commit https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/770/commits/5b16ed2fabbf62ca43930303499def403e8d2b96.  I added a couple of commits:  
- https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/770/commits/fe7ff3ef08f836053226eb3ad5422501ea241d6b should improves the human readability of the if-condition in question.  
- https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/770/commits/3316ff65c4a75fcade5016695dd960036c5a6e59 moves getids.F90 into a more appropriate location.  The repo builds, so chances are the move is ok, but we won't know for sure without testing. 

The PR should be trivially 0-diff for the GCM.  

@biljanaorescanin: When you get a chance, for now please run:
- just one basic GCM test to be safe, 
- the full suite of GEOSldas "nightly" tests, and 
- the standard suite of "remap_restart" tests.  
 
We can add the full suite of GCM tests if @sdrabenh would want us to do so.

biljanaorescanin:
Tests summary:

1. 1-day free run (AMIP) develop vs  branch is zero diff. 
2. Nightly Tests are zero diff.
3. Remap package for catch restart remapping is zero diff ( tested for 2 cases) 
4. Remap package for catchCN CLM4.5 is not zero diff as expected. 
Non zero diff run was to remap from M36 to M09 and 2 variables differ: 
```
              Date     Time   Level Gridsize    Miss    Diff : S Z  Max_Absdiff Max_Reldiff : Parameter ID
    33 : 2001-02-01 21:00:00       0 176896125         0   14530 : T T       13339.           1.0000 : -33   (CNCOL)  
    34 : 2001-02-01 21:00:00       0 1516252500       0   23841 : T T   1.0486e+06      1.0000 : -34  (CNPFT)
```
Similarly for M09 to C180 for CatchCN CLM 4.0, same two variables are not zero diff from the restart file.

gmao-rreichle:
Thanks, @biljanaorescanin, for the test reports.  @gmao-jkolassa - do the differences in the CatchCN remapping tests reported by @biljanaorescanin  above make sense to you? 
Also, the tests that I had originally asked Biljana to do included only remapping of Catch restarts, as part of the remap_restarts test suite. We can take this opportunity to add a test case that remaps a CatchCN restart, with the help of @gmao-jkolassa. But the first priority is to make sure we're 0-diff for remapping catch restarts, and to move forward to the point where remap_restarts.py can be rolled out for the general public (so we can retire regrid.pl). To that end, we don't need to have everything sorted out for remapping CatchCN restarts. The bug fix here is definitely good to have, but if the test case for remapping CatchCN restarts takes longer, we don't have to wait for it.

(The latter part of comment was inadvertently placed on the wrong PR: https://github.com/GEOS-ESM/GEOS_Util/pull/23#issuecomment-1594719588)

gmao-jkolassa:
> Thanks, @biljanaorescanin, for the test reports. @gmao-jkolassa - do the differences in the CatchCN remapping tests reported by @biljanaorescanin above make sense to you? 

Yes, the differences reported are consistent with what I found during my tests.

gmao-rreichle (approved):
Approving for @GEOS-ESM/surface-preproc-team and @GEOS-ESM/land-team

mathomp4 (approved):
I am still mightily confused as to why Github doesn't seem to say @gmao-rreichle can approve for some teams. But I approve for CMake and then as superapprover, should approve for everything else.

sdrabenh (approved):

PR #774, opened 2023-06-26
A cleaner version of my branch feature/atrayano/faster_routing

atrayano:
This PR only changes the communication pattern during fresh water river runoff. It is zero diff in coupled mode, and obviously zero diff in EMIP since  routing is not used in that mode. This was tested with coupled GEOS-MITgcm and resulted in significant speedup.  But the code mods are only in GEOS_SurfaceGridComp.F90 and there is no assumption about the specific ocean model. I am hoping runs with MOM6 will also benefit from this

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
@@ -74,7 +74,8 @@ module GEOS_SurfaceGridCompMod
 
   type( ESMF_VM ) :: VMG
 
-! !PUBLIC MEMBER FUNCTIONS:
+
+  ! !PUBLIC MEMBER FUNCTIONS:
 
   public SetServices
 
@@ -112,9 +113,16 @@ module GEOS_SurfaceGridCompMod
   type T_Routing
      integer :: srcTileID, dstTileID,     &
                 srcIndex=-1, dstIndex=-1, &
-                srcPE=-1, dstPE=-1 
+                srcPE=-1, dstPE=-1, SeqIdx=-1 
      real    :: weight
   end type T_Routing
+
+  type T_RiverRouting
+     type(T_Routing), pointer :: LocalRoutings(:) => NULL()
+     integer, pointer :: karray(:)
+     integer, pointer :: kdx(:)
+     integer, pointer :: BlockSizes(:), displ(:)
+  end type T_RiverRouting
 
 ! Internal state and its wrapper
 ! ------------------------------
@@ -123,7 +131,7 @@ module GEOS_SurfaceGridCompMod
      private
      type (MAPL_LocStreamXFORM)  :: XFORM_IN (NUM_CHILDREN)
      type (MAPL_LocStreamXFORM)  :: XFORM_OUT(NUM_CHILDREN)
-     type (T_Routing), pointer   :: LocalRoutings(:) => NULL()                
+     type (T_RiverRouting), pointer   :: RoutingType => NULL()                
   end type T_SURFACE_STATE
 
   type SURF_WRAP
@@ -3648,7 +3656,7 @@ module GEOS_SurfaceGridCompMod
     VERIFY_(STATUS)
 
     if (RoutingFile /= "") then
-       call InitializeRiverRouting(SURF_INTERNAL_STATE%LocalRoutings, &
+       call InitializeRiverRouting(SURF_INTERNAL_STATE%RoutingType, &
             RoutingFile, LocStream, rc=STATUS)
        VERIFY_(STATUS)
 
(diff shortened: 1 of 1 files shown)


amolod (approved):

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
```diff
@@ -3923,6 +3939,72 @@ subroutine InitializeRiverRouting(LocalRoutings, RoutingFile, Stream, rc)
...
+    end do
+    allocate(kdx(nsdx), blocksizes(nDEs), _STAT)
+    blocksizes=0
+    
+    ! exchange with everybody else
+    call MPI_AllGather(nsdx, 1, MP_Integer, &
+         blocksizes, 1, MP_Integer, comm, status)
+    _VERIFY(status)
+    
+    ! not everybody has blocksizes(nDEs)
```
gmao-rreichle, line 3958: @atrayano: Should this comment read:      "! **_now_** everybody has..." ?         Sorry if I misunderstood.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
```diff
@@ -3923,6 +3939,72 @@ subroutine InitializeRiverRouting(LocalRoutings, RoutingFile, Stream, rc)
...
+
+    ntotal = sum(blocksizes) ! this should be the same as npairs
```
gmao-rreichle, line 3960: @atrayano: Should "npairs" read "numRoutings" in this comment?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
```diff
@@ -3923,6 +3939,72 @@ subroutine InitializeRiverRouting(LocalRoutings, RoutingFile, Stream, rc)
...
+    karray = 0
+    allocate (displ(0:nDEs), _STAT) !declare as target!!!
+    
+    ksum = 0
+    displ(0)=ksum
+    do n=1,nDEs
+       ksum = ksum + blocksizes(n)
+       displ(n)=ksum
+    end do
+    ! as another sanity check: ksum should be the same as npairs
```
gmao-rreichle, line 3973: @atrayano: Should "npairs" read "ntotal" (or "numRoutings") in this comment?

gmao-rreichle:
pinging @zyj8881357

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
atrayano: Good catch! Yes, this was a typo! Thanks! I fixed it

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
atrayano: "ntotal" and "npair" intentionally are different variable to do more sanity checks. I prefer to keep these names

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
atrayano: Same answer as above

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
gmao-rreichle: Thanks, but then what is "npairs"?  It's not obviously defined elsewhere in the F90 file as far as I can understand it.  Maybe it is the number of "mpi send/recv pair[s]" referenced here:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/e82ad4d9a0a9a86954e29764ff3f56a6832cab99/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90#L3901
If so, just a tad more clarification would be helpful. 
Sorry to be nitpicking, but going forward it's critical that the land group understands this code, and we will need all the help that we can get.

atrayano: @gmao-rreichle In your last comment, you are referring to a section of the code written many years ago by Max Suarez. The way I see this, the ASCII version of the routing file (*.trn) has number of "paired" sources and sinks, and each of these sources and sinks is identified by the tile Id.. This is what I refer as "npairs". The old version of this code used to send individual messages (sends and receives) between these pairs, and since their number is quite large, we were paying too much cost for the latency of each these calls

zhaobin74 (dismissed):

mathomp4:
Pinging @yvikhlya as well. I think he hit slow river-routing with MOM6 runs at NAS. Hopefully he can test this change to see if speed returns!

gmao-rreichle (dismissed):
Thanks, @atrayano, for making this more tractable!  Happy to approve. 
One last suggestion: Would it be correct to rephrase the comment line as follows:
`ntotal = sum(blocksizes) ! should be same as # of paired sources and sinks (npairs)`
It would be helpful to clarify the meaning of "npairs".

gmao-rreichle (approved):
re-approving

zhaobin74 (approved):

sdrabenh (approved):

PR #775, opened 2023-06-28
Added -DAQUA_PLANET for compiling Surface GridComp

pchakraborty:
With this preprocessor macro, we completely ignore Lake, LandIce and Land GridComps. Results are zero-diff to the case where codepath includes these components.


tclune (changes requested):
Should this maybe instead be a request to add these definitions to the `${this}` target?  CMake best practices are to avoid directory level settings in favor of targets.

mathomp4:
> Should this maybe instead be a request to add these definitions to the `${this}` target? CMake best practices are to avoid directory level settings in favor of targets.

Indeed, I think [`target_compile_definitions`](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html#command:target_compile_definitions) is the way with the least "generality"

pchakraborty:
I have added the definition to `${this}`.

PR #777, opened 2023-06-30
Update WSUB_ExtData.yaml

dbarahon:
Added correction for L72. A better fix would select the folder depending on vertical resolutionn.


mathomp4:
> Added correction for L72. A better fix would select the folder depending on vertical resolutionn.

@dbarahon I believe @bena-nasa has something in his todo list to 'template' the vertical bits of these files, but for now, we have bits in scripting that substitute in the right level numbers.

sdrabenh (approved):

PR #783, opened 2023-07-17
Feature/narnold1/#560 shocmf updates

narnold1:
Extensive updates to the SHOC-EDMF turbulence parameterizations.  

This PR is non-0-diff due to fixing a bug in the calculation of (non-SHOC) TKE in GEOS_TurbulenceGridComp.F90.  If bugfix is excluded, output fields are 0-diff.

This update also renames several SHOC variables (switching 'HL' to 'SL' for consistency with other model variables), which technically changes restart files.  The affected variables are not used in default model configurations.


GEOSagcm_GridComp/GEOS_AgcmGridComp.F90
```diff
@@ -189,6 +191,9 @@ subroutine SetServices ( GC, RC )
         RST = MAPL_RestartSkip
     end if
 
+    call MAPL_GetResource(MAPL, SCM_SL, Label="SCM_SL:", default=0, RC=STATUS)
+
+
```
tclune, line 196: Missing check on STATUS.  
```suggestion
    call MAPL_GetResource(MAPL, SCM_SL, Label="SCM_SL:", default=0, _RC)


```

tclune (dismissed):
CMake changes ok

narnold1:
Updated the branch to remove the bugfix.  Now 0-diff for all restarts except turb_internal_rst, turb_import_rst, and moist_import_rst, which are superficially affected by the HL to SL variable name changes.

mathomp4 (approved):

PR #786, opened 2023-07-18
Boundary conditions cleanup

biljanaorescanin:
This PR: 
 - cleans up output layout directory structure 
 - adds runs start and end time information to log file
 - fixes exit in case of run already being present 

PR was tested for multiple cases for CS run at same time ( c90, c180, c24), ease case M36, lat-lon case b (144x91). 
It is zero diff to what we have on develop for boundary conditions comparison. 
It is trivially zero diff for AGCM since all changed files are in `Raster/makebcs/` directory.


GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_shared.py
```diff
@@ -138,22 +129,33 @@ def get_script_mv(grid_type):
...
+else
+    /bin/cp -p {MAKE_BCS_INPUT_DIR}/land/CO2/v1/CO2_MonthlyMean_DiurnalCycle.nc4 land/shared/CO2_MonthlyMean_DiurnalCycle.nc4
+    echo "Successfully copied CO2_MonthlyMean_DiurnalCycle.nc4 to bcs dir."
 endif
-  /bin/mv land/shared/CO2_MonthlyMean_DiurnalCycle.nc4 ../../land/shared/.
-cd ..
-/bin/mv logs/{GRIDNAME} ../logs/.
-cd ../
+
+# adjust permissions
```
gmao-rreichle, line 155: chmod 755 assigns "execute" permission to all files, regardless of whether they are executables/scripts or just plain data/text files.  Is there a cleaner way to set the permissions?  Are there even executables in the bcs dir?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_shared.py
```diff
@@ -138,22 +129,33 @@ def get_script_mv(grid_type):
...
+/bin/mv land/{GRIDNAME}     ../../land/.
+
+/bin/rm -r {TMP_DIR}
+
+# if necessary, copy of CO2 file from MAKE_BCS_INPUT_DIR to bcs dir 
+
+cd ../..
+
+if(-f land/shared/CO2_MonthlyMean_DiurnalCycle.nc4) then
+    echo "CO2_MonthlyMean_DiurnalCycle.nc4 already present in bcs dir."
```
gmao-rreichle: @biljanaorescanin: I tried to simplify the logic a bit.  I did not test the revised code, and I don't really expect it to work without additional fixes.  I committed it now so you have a sense of what needs streamlining.
It does not make sense to copy the CO2 file in the middle of dealing with the GRIDNAME (resolution) directory, because the CO2 file is not resolution-specific.  This can be done at the end, after the resolution-specific tasks have been completed.
There should be no need to check of a dir exists before creating the dir.  If the dir exists, mkdir will do nothing. 
Some of the mkdir and chmod commands seemed redundant.

gmao-rreichle:
> ... do we want to change `csh` script as well (`make_bcs`) or do we add message that this script is now outdate when compared to `make_bcs.py`?

I think now's the time to deprecate `make_bcs` (csh), that is, move it into the `obsolete` dir and remove it from the install in CMakeLists.txt

weiyuan-jiang (commented):
I have moved all those echos to the shared script

gmao-rreichle:
> I have moved all those echos to the shared script

Many thanks, @weiyuan-jiang!  I was wondering if the same could be done for the "Abort" messages ("This BCS run is configured to create output that already exists:").   If the "if" condition is available in the shared script, it would be cleaner to check for it there.

weiyuan-jiang:
> 

I moved to a function

gmao-rreichle:
@biljanaorescanin : I don't understand this commit:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/786/commits/ef9e56ec417ff29b63dcadb3b6401e418d2466be
We already remove the tmp_dir a few lines earlier.  There should be no need for adding the rm again at the end.  I may be missing something.

biljanaorescanin:
It didn't remove the directory. Once I returned that line it did. Also with this new function model doesn't exit anymore like it did before in case of same experiment being submitted.

gmao-rreichle:
> It didn't remove the directory. Once I returned that line it did. Also with this new function model doesn't exit anymore like it did before in case of same experiment being submitted. UPDATE: I will have to rerun this so my comment might be wrong. I didn't pull all commits for testing.

@biljanaorescanin : I had overlooked the `cd ../..` command when I moved the command to the earlier, more logical position (i.e., immediately after we're done with the tmp dir).  I switched things around again to have the more logical order of operations.  I hope this will work.  I don't know what might be wrong with the check for existing output ("Abort:...").  @weiyuan-jiang's solution of having a function seems perfectly good to me.   Perhaps the string assembly in the argument of `check_script()` has an error?

biljanaorescanin:
@gmao-rreichle When I added `import glob` now abort is working it wasn't defined in script that was using it

biljanaorescanin:
This PR was re-tested. All PR functionality appears to be working.

gmao-rreichle:
@biljanaorescanin: I added a trivial commit for better consistency of the make_bcs_[GRID].py scripts.  The PR does not need to be re-tested, as far as I'm concerned.

gmao-rreichle (approved):

mathomp4 (approved):

sdrabenh (approved):

PR #772, opened 2023-06-20
add new sea ice components for coupling to CICE6

zhaobin74:
This PR adds two new gridded comps to support coupling `GEOS` to `CICE6` sea ice model from [CICE Consortium](https://github.com/CICE-Consortium).  

`GEOS_SeaiceInterfaceGridComp` and `CICE_GEOSPlug` were introduced to allow `GEOS` atmosphere/surface to thermally couple to CICE6 in a semi-implicit manner while conserving energy and mass.

`USE_CICE_Thermo = 2` switch is used to turn on the new `GEOS_SeaiceInterfaceGridComp` and use `CICE6` sea ice model.
Currently `USE_CICE_Thermo = 0` is for AMIP `Dataseaice` and `USE_CICE_Thermo = 1` for coupled `CICE4` sea ice model.

Testing is done in 3 configurations:
- AMIP style run, 2-days segment, `c12xo1` resolution,  **zero-diff results**
- Coupled run with MOM5, 2 days segment, `c90-360x200` resolution, **zero-diff results**
- Coupled run with MOM6, 2 days segment, `c12-72x36` resolution, **zero-diff results**    

**NOTE: CICE6 coupling is still in EXPERIMENTAL stage** 
 
This PR depends on  https://github.com/GEOS-ESM/GEOSgcm_App/pull/478 and https://github.com/GEOS-ESM/GEOSgcm/pull/611

PR #789, opened 2023-07-28
Fix issue with Ocean TerminateImport in Surf

mathomp4:
Runs of the MAPL3 development code triggered a fault due to changes in #772 see:

* https://github.com/GEOS-ESM/MAPL/issues/2262
* https://github.com/GEOS-ESM/MAPL/issues/2263

Now, for reason as yet unknown (though @bena-nasa is investigating) MAPL2 doesn't seem to mind this, but MAPL3 does. However, I believe this is an issue with develop as well.

To wit, in `GEOS_GcmGridComp.F90` we have:
```fortran
    if (DO_CICE_THERMO == 2) then  
       call MAPL_AddConnectivity ( GC,                              &
            SHORT_NAME  = (/'SURFSTATE'/),                          &
            DST_ID = AGCM,                                          &
            SRC_ID = OGCM,                                          &
            RC=STATUS  )
       VERIFY_(STATUS)
    endif
```
and in `GEOS_OgcmGridComp.F90`

```fortran
  if (DO_CICE_THERMO == 2) then
     call MAPL_AddExportSpec ( GC   ,                          &
          SHORT_NAME = 'SURFSTATE',                            &
          CHILD_ID   = SEAICE ,                                &
                                                           _RC )
  endif
```

In both cases the `SURFSTATE` code is protected with an if-block for `DO_CICE_THERMO==2`. But in `GEOS_SurfaceGridComp.F90` we have:
```fortran
   ! call MAPL_TerminateImport    ( GC, CHILD = OCEAN,   RC=STATUS  )
    call MAPL_TerminateImport    ( GC, SHORT_NAMES=['SURFSTATE'],    &
                                   CHILD_IDS=[OCEAN],  RC=STATUS  )
```
which seems to cause some issues when running in the plain `DO_CICE_THERMO=0` way.

So my proposal is we only do the TerminateImport with `SURFSTATE` mentioned if we are running `DO_CICE_THERMO=2` and if not, we use the "old" TerminateImport code which (I think?) terminates all of Ocean.

Since #772 was @zhaobin74 's code, I'll explicitly add him here. I think this is safe, but I want his blessing.

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
@@ -207,7 +207,7 @@ module GEOS_SurfaceGridCompMod
     type (T_SURFACE_STATE), pointer         :: SURF_INTERNAL_STATE 
     type (SURF_wrap)                        :: WRAP
     type (MAPL_MetaComp    ), pointer       :: MAPL
-    INTEGER                                 :: LSM_CHOICE
+    INTEGER                                 :: LSM_CHOICE, DO_CICE_THERMO
     character(len=ESMF_MAXSTR)              :: SURFRC
     type(ESMF_Config)                       :: SCF        ! info from Surface Config File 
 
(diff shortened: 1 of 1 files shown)


gmao-rreichle (approved):

lcandre2 (approved):

zhaobin74 (approved):
Thanks for catching this. It was a bug of not protecting the MAPL_terminate call. Now this fix worked for both CICE4 and CICE6.

sdrabenh (approved):

PR #792, opened 2023-07-31
Update CircleCI to use Orb Default Baselibs and BCs

mathomp4:
This PR updates the CircleCI config to default to using the Baselibs in the CircleCI orb rather than explicitly specifying the versions. In most cases, this is the "best" way as a change to the orb can fix this for projects without the need for a PR.

Note that we keep commented the explicit anchors in case needed.


mathomp4:
Superpowering in.

PR #793, opened 2023-07-31
Integrate OpenACC port of aerosol activation

pchakraborty:
Parallel to `aer_actv_single_moment.F90`, there is a new file `openacc/aer_actv_single_moment.F90` which is used when compiled with `-DGPU_PHYSICS`.


GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/CMakeLists.txt
```diff
@@ -44,7 +47,8 @@ target_include_directories(${this} PRIVATE
    )
 if (GPU_PHYSICS)
   message(STATUS "Building GPU ports of Moist routines")
-  target_compile_options(${this} PRIVATE -fopenacc -foffload=nvptx-none)
+  set (CMAKE_Fortran_FLAGS_DEBUG "${CMAKE_Fortran_FLAGS_DEBUG} ${GEOS_Fortran_OpenACC_Flags}")
```
mathomp4, line 50: @pchakraborty I'm wondering. Why aren't you using `target_compile_options` here? It seems like that would work...

pchakraborty: For a very silly reason actually - can’t seem to get around the stringification
`-fPIC   -pthread "-fopenacc -foffload=nvptx-none" -pthread`, which throws the error `gfortran: error: unrecognized command-line option '-fopenacc -foffload=nvptx-none’`.

PR #801, opened 2023-08-17
Add Stretched Cube Sphere option to BCS package

biljanaorescanin:
In support of the Hazardous Weather Testbed ( HWT project) @atrayano implemented stretched grid capability. Original implementation he shared was done for old boundary conditions package so it had to be redone for python version.  

This current implementation offers these choices:

`Choose Stretched_Cubed-Sphere grid option:`
     `    Name            Stretch Factor          Lat             Lon              Resolution Choices`
     ` ----------       -------------        ------------   -------------   ---------------------`
     `  SG001               2.5                 39.5                 - 98.35           c270, c540, c1080 and c2160`
     `  SG002               3.0                 39.5                - 98.35           c1536`

Output grid name will have stretched grid identifier for example: `CF1080x6-SG001_CF1080x6C` 

I'm reproducing this this PR set currently staged and used from here: /discover/nobackup/projects/gmao/osse2/stage/BCS_FILES/CF1080x6C_CF1080x6C


biljanaorescanin:
@gmao-rreichle  and @weiyuan-jiang few questions: 
1. I wasn't sure do we want to add new identifier with the tile file name or just the output name so now its just at output level. Let me know if I need to change this. 
Example tile file is still `CF1080x6C_CF1080x6C-Pfafstetter.til` and not "`CF1080x6C-SG001_CF1080x6C-Pfafstetter.til`
2. I don't know why but file file itself has different number of tiles that one originally @atrayano produced?  I'm running few more tests after that I will update with more information.

mathomp4:
I suppose one question I've had is should we be more specific with the naming scheme? That is instead of `SG001` something like `lat_x_lon_x_stretch`?

Also, I think these are for longitude of -98.35 as they are over CONUS.

biljanaorescanin:
> I suppose one question I've had is should we be more specific with the naming scheme? That is instead of `SG001` something like `lat_x_lon_x_stretch`?
> 
> Also, I think these are for longitude of -98.35 as they are over CONUS.

I think @gmao-rreichle can comment more but his idea is to leave flexibility if this changes we can easily add other lats /lons.  And not have super long name for boundary conditions.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_questionary.py
```diff
@@ -203,6 +208,24 @@ def ask_questions(default_grid="Cubed-Sphere"):
...
+       {
+            "type": "checkbox",
+            "name": "Stretched_Cubed-Sphere",
+            "message": f'''Choose Stretched_Cubed-Sphere grid option:
+               Name   Stretch_Factor  Focus_Lat  Focus_Lon  Resolution_Choices
+               -----  --------------  ---------  ---------  ------------------------
+               SG001      2.5            39.5      -98.35   c270, c540, c1080, c2160
+               SG002      3.0            39.5      -98.35   c1536                    \n ''',
+
+            "choices": [ \
```
gmao-rreichle, line 220: I'm probably misinterpreting this, but when the task is to select a particular stretched CS grid, shouldn't the `"choices":` be "SG001", "SG002", etc?

gmao-rreichle:
> I suppose one question I've had is should we be more specific with the naming scheme? That is instead of `SG001` something like `lat_x_lon_x_stretch`?
> 
> Also, I think these are for longitude of -98.35 as they are over CONUS.

@mathomp4: As @biljanaorescanin said, the string that identifies the particular Stretched CS Grid would have to be rather long, and we only have a single underscore that separates the atm grid resolution from the ocean grid resolution in the bcs naming convention.   Using something like `lat_x_lon_x_stretch` would introduce additional underscores that would mess up the parsing of the resolution string, and without the underscores the string would be even harder to interpret.  But I'm open to suggestions.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_questionary.py
gmao-rreichle: @biljanaorescanin, @weiyuan-jiang: I'm still confused about the logic here.  How can the "choices" below be for resolution only when there are different stretch parameters listed above?  Don't we have to ask for SG[xxx] first and then for resolution?  How would the choices offered below (i.e., "cXXX") identify the stretch parameters?

biljanaorescanin: For example, all these resolutions (c270, c540, c1080, c2160)  have the same stretched parameter and identifier `SG001`. The user picks the grid option and gets a table where they know that if they choose `c270`, they will get this `SG001`  factor. Currently, only two possibilities exist, and it is an easy way to add more with any stretched factor.  

Can we run `c1536`  with `SG001`? The answer is no.  
But why would we want to do that if it's never going to be used? 

We can change this if you wish. It will be flexible, but that would make it prone to mistakes. What will stop users from running `c270` with `SG002` in that case and complaining they are not getting what they are supposed to?  
This branch is zero diff for all resolutions listed above when compared to Bill's area. I didn't see any runs in same resolution for different factors.

gmao-rreichle: @biljanaorescanin: I see, but I think this is unlikely to hold up going forward.  It's not clear to me why SG001 and SG002 will never share a given cXXXX resolution.  Or maybe someone will need SG003 with the same stretch factor and resolution as SG001 but a different focal point.  The only way to make the logic hold up going forward is to first ask for the SG parameter set (ie, SGxxx) and then ask for the resolution (ie, cXXXX).  We can make the choice of resolution depend on the SGxxx choice, if you're worried that users will inadvertently pick the wrong resolution.  That is, for SG001 we only offer the resolutions listed in the table, and for SG002 we only offer c1536 (for now).

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_questionary.py
```diff
@@ -203,6 +214,39 @@ def ask_questions(default_grid="Cubed-Sphere"):
             "when": lambda x: "Lat-Lon" == x['grid_type'],
         },
 
+       {
+            "type": "checkbox",
```
gmao-rreichle, line 218: @biljanaorescanin: Shouldn't this be type "select"?  My uneducated guess is that "checkbox" let's the user select more than one option.  But selecting both SG001 and SG002 probably does not work when we then select the resolution because the resolution questions both return "Stretched_CS", so only the answer for SG002 would be remembered.  I think it's easiest to treat SG001 and SG002 as two grids (like, say, CS and EASE).  That is, the user would need to run make_bcs.py separately for SG001 and SG002.  Does this make sense?

biljanaorescanin:
This PR was tested after last commit on Friday Sept 15th.  
It is zero diff for CS grid and zero diff to what Bill was using for stretched grid. I've tested tripolar grid options as well and those are zero diff. 
Only boundary conditions set I can't confirm is zero diff is the new option stretched grid with Tripolar (we don't have reference for it) for example: `CF0270x6C-SG001_M6TP1440x1080` but we are able to run it and all looks like its working. 

Since all code changes are in boundary conditions this is trivially zero diff for GCM.

gmao-rreichle (approved):

gmao-rreichle:
@sdrabenh : This PR is ready for merging.  I approved it, and my approval should have counted towards the @GEOS-ESM/surface-preproc-team, but that failed again.  I guess @mathomp4 can super-approve, assuming the two of you -- as part of the @GEOS-ESM/python-transition-team -- are ok with the python changes.

gmao-rreichle (approved):

mathomp4:
Ahh. @gmao-rreichle I just removed you and readded you to @GEOS-ESM/surface-preproc-team . And Github Seems to have liked that.

gmao-rreichle:
@mathomp4:  On the PR landing page, I had just noticed a request for my review, despite my earlier approval and the page showing a green checkmark next to my name.  I went ahead and approved again, which I thought then took care of the surface-preproc approval.  Not sure whether your action of  removing and re-adding me prompted the 2nd review button, or whether it was something else. Also, I do remember that you did the removing/re-adding a few weeks ago when we were in a similar situation.  So whatever the removing/re-adding might fix, it doesn't seem to last, sadly.  In any case, I think we're good to go - provided there aren't any objections to py changes.

mathomp4 (approved):
Python looks good. I mean, I assume it was tested.

sdrabenh (approved):

PR #813, opened 2023-08-31
Additional cleanup of Stieglitz snow code

gmao-rreichle:
Miscellaneous, 0-diff cleanup of Stieglitz snow code:
- Print WARNING when encountering positive snow temperatures (PosSnowHeat) [see item 4 below].
- Pass tile lat/lon into StieglitzSnow_snowrt().
- Clean up logic of if-block for (tzero, ice1) conditions.
- Added comment about value of CPW (spec heat of ice) [see item 2 below].

This PR started out as an attempt to address occasionally large garbage fluxes during snow compaction, which would be a non-0-diff change [see item 1 below].  This particular change should be included in a future, non-0-diff PR.

Additional non-0-diff changes were attempted but found to be problematic or undesirable.

Below is the summary of all changes that were implemented or considered and ultimately.

--------------------------------------------------------------

### 1. Maintain SWE after snow compaction (avoid large garbage fluxes)

**For LAND tiles only**: When snow compaction reaches the max allowed snow density, keep the snow mass (or snow water equivalent; SWE) unchanged and adjust the snow depth to limit the density to the max allowed (as proposed by @rdkoster). 
Previously, after compaction reached the max allowed density, snow mass was removed to cut the density back to the max allowed value.  This excess snow mass was added into a garbage flux term ("excs").  This mechanism resulted in large excs residual terms because of repeated densification beyond the max density and subsequent removal of snow mass excs.  (This mechanism is retained for **LANDICE** tiles.)

A global, land-only simulation on the 36-km EASEv2 grid with hourly output for one winter season (~Oct-May) confirmed that the standard snow-related output (snow mass, snow depth, snow surface temperature, sublimation) all remain reasonable with the currently implemented bug fix.  

The downside of the bug fix is that the snow mass and snow depth can grow indefinitely for land tiles that are "faux landice" (where snow does not melt off completely during the warm season).  It is not clear if among the landice tiles there are "faux land" tiles (where snow/ice melt off completely during the warm season) and what, if anything, could be done to minimize the total number of "faux" tiles.

For reference, after snow densification and the aforementioned check on the max density, there is another check on the max snow depth here:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/3d764c6c45d6cdb26e66131a8e6b188923f9bd7d/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L890
For land tiles, however, this snow depth check is essentially disabled (CATCH_MAXSNDEPTH=1.e20 [m]).
For landice tiles, the threshold is MAXSNDZ=15. [m].

**Summary:** Non-0-diff change implemented was originally implemented as discussed above.  Rolled back in [this commit](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/813/commits/4633293db1faa7f5617d0d2e6343f5ef10d0a4f6) to avoid non-0-diff changes.  This particular change should be considered in a future non-0-diff change. 

### 2. Custom value of CPW (spec heat of ice)

StieglitzSnow_CPW=2065.22 J/kg/K is hard-coded here:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/65465f6dbe9a524876225a10fe9b21b8294be362/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L62C31-L62C43

Brief research suggests that this value is good near 0 deg C whereas MAPL_CAPICE=2000. J/kg/K is more appropriate near -10 deg C.  

**Summary:** Added comment to source code, but did not change numerical values. 

### 3. Make calc_tpsnow calculations consistent for scalar and vector inputs.

The legacy code uses hardwired 1.0000001 and 0.99999999 factors in the `if` condition when diagnosing snow temperature in the diagnostic routine for scalars [originally get_tf0d()].  These factors are not used in the diagnostic routine for vectors [originally get_tf_nd()].  

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/65465f6dbe9a524876225a10fe9b21b8294be362/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L1363

Legacy threshold factors were [disabled in this commit](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/813/commits/496b6513fcb89e5d06a52e7d577b9bb1eb361765)

**Summary:** Non-0-diff change was originally implemented as discussed above.  Rolled back in [this commit](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/813/commits/4633293db1faa7f5617d0d2e6343f5ef10d0a4f6) because the change resulted in bad snow simulations.

### 4. Fix bug in assigned snow temperature of partially frozen pack at 0 deg

The calc_tpsnow subroutine could produce a positive snow temperature if the input snow heat content is positive.

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/65465f6dbe9a524876225a10fe9b21b8294be362/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L1409

**Summary:** Check for positive snow heat content in StieglitzSnow_calc_tpsnow() [added in this set of commits](https://github.com/GEOS-ESM/GEOSgcm_GridComp/compare/496b651..b888602?w=1).  Note that one call to StieglitzSnow_calc_tpsnow() in StieglitzSnow_snowrt() uses a tentative snow heat content and may produce positive snow temperatures.  The check is turned off for this call. 

### 5. Snow heat content adjustments after relayer()

After snow redistribution in relayer(), the thermal conditions of each new layer `i` are checked against those of the corresponding old layer `i`.  Partially frozen conditions at 0 deg are not allowed to turn into fully frozen conditions below 0 deg (and vice versa), and the snow heat content and temperature are adjusted accordingly.   It is unclear, however, if this makes sense when the redistribution is such that the old and new layers have minimal overlap.  This is a particular concern when relayer is used in Landice (N_snow=15) or after adding snow mass increments from a land analysis.

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/65465f6dbe9a524876225a10fe9b21b8294be362/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L1279

**Summary:** After initially abandoning the (ice10,tzero) check and snow heat content adjustment [in this commit](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/813/commits/9d562a955d5937099804180390c624b7b0bfd580?w=1), the check was found to be important for maintaining a correct energy balance and reinstated [in this commit](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/813/commits/802765e603b65eba518b0aabdca7a98bbf9670a4) (plus the subsequent 2 fixes).

### 6. Snow density check in relayer()?

After (or near the end of) relayer(), do we need a check of the snow density against the min/max allowed?  

**Summary:** Snow density will be checked again at next time step.  Leave source code unchanged. 

### 7. Minimum value for snow cover area fraction

In a couple of places, the snow cover area fraction is set to a minimum of small=1.e6.  Is there an inconsistency with the minimum of 0. implied elsewhere? 

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/65465f6dbe9a524876225a10fe9b21b8294be362/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L428

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/65465f6dbe9a524876225a10fe9b21b8294be362/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L894

**Summary:** Slight inconsistency is acceptable.  Leave source code unchange. 

cc: @rdkoster @lcandre2


GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
```diff
@@ -1664,6 +1664,51 @@ subroutine SetServices ( GC, RC )
...
+  VERIFY_(STATUS)
+
+  call MAPL_AddExportSpec(GC,                               &    ! lca/rhr not sure if right
+    LONG_NAME          = 'total_meltwater_production_flux' ,&
+    UNITS              = 'kg m-2 s-1'                      ,&
+    SHORT_NAME         = 'MLTWTR'                          ,&
+    DIMS               = MAPL_DimsTileOnly                 ,&
+    VLOCATION          = MAPL_VLocationNone                ,&
+                                           RC=STATUS  )
+  VERIFY_(STATUS)
```
gmao-rreichle, line 1710: need to verify what is and isn't included in PREOUT and MLTWTR and adjust LONG_NAME accordingly;  also need better SHORT_NAME for PREOUT

gmao-rreichle:
@lcandre2, @zhaobin74:  I'm wondering why the "tiletype" isn't passed into StieglitzSnow_snowrt() by using the simple constant MAPL_LANDICE, as in `call SNOWRT( 1, NUM_SNOW_LAYERS, MAPL_LANDICE, ...`

Instead, GEOS_LandIceGridComp.F90 uses `TileTypes`:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/9e8ad9a292fdb33e3e3f4abb55596e2660d64470/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSlandice_GridComp/GEOS_LandIceGridComp.F90#L3089
which is obtained from MAPL here:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/9e8ad9a292fdb33e3e3f4abb55596e2660d64470/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSlandice_GridComp/GEOS_LandIceGridComp.F90#L2183

Is it possible that `TileTypes` is a vector for **all** tiles (incl. land, lake, landice, and ocean)?  I assume `TileTypes` is somehow zoomed into just the landice tiles, but then it would be equivalent to using the scalar constant `MAPL_LANDICE` as suggested above, which is much simpler.  

If `TileTypes` is **not** zoomed into just the landice tiles, could StieglitzSnow_snowrt() be getting the wrong tile type?  Again, I can't imagine this is the case, but I thought I should flag it just in case.    Sorry I'm so ignorant about MAPL's inner workings.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
lcandre2: Will look into this and come up with a better name for PREOUT.

zhaobin74:
> @lcandre2, @zhaobin74: I'm wondering why the "tiletype" isn't passed into StieglitzSnow_snowrt() by using the simple constant MAPL_LANDICE, as in `call SNOWRT( 1, NUM_SNOW_LAYERS, MAPL_LANDICE, ...`
> 
> Instead, GEOS_LandIceGridComp.F90 uses `TileTypes`:
> 
> https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/9e8ad9a292fdb33e3e3f4abb55596e2660d64470/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSlandice_GridComp/GEOS_LandIceGridComp.F90#L3089
> 
> which is obtained from MAPL here:
> 
> https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/9e8ad9a292fdb33e3e3f4abb55596e2660d64470/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSlandice_GridComp/GEOS_LandIceGridComp.F90#L2183
> 
> Is it possible that `TileTypes` is a vector for **all** tiles (incl. land, lake, landice, and ocean)? I assume `TileTypes` is somehow zoomed into just the landice tiles, but then it would be equivalent to using the scalar constant `MAPL_LANDICE` as suggested above, which is much simpler.
> 
> If `TileTypes` is **not** zoomed into just the landice tiles, could StieglitzSnow_snowrt() be getting the wrong tile type? Again, I can't imagine this is the case, but I thought I should flag it just in case. Sorry I'm so ignorant about MAPL's inner workings.

@gmao-rreichle, you are right about  ```TileTypes``` being zoomed into just the landice tiles. Yes, it is equivalent to using  ```MAPL_LANDICE```. I don't recall the details as to why the call to ```snow_rt``` used ```TileTypes```.  I think using ```MAPL_LANDICE``` in the call could simply it.

gmao-rreichle:
@rdkoster: A bit more info on the positive snow temp issue.  Biljana set up an offline run on the 36-km EASE grid starting in Sep 2022.  When the run gets to 9 March 2023, 23z, it encounters a positive snow heat content and temperature in the call to StieglitzSnow_calc_tpsnow() here:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/a87349a279564a23fb8fd16bed3a884dd742fb38/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L639

I don't know exactly what the snow model does in this block.  In any case, the offending values within StieglitzSnow_calc_tpsnow() are: 
htsn = 49.60547     
wesn = 3.1575907E-02  
tpsn  = 0.7606894    
fice   = 0.0000000E+00 
ice1  = F 
tzero = T

That is, the snow mass is small (~0.03 mm) and only barely above the MINSWE value of 0.013 mm used in the snow routine, but the temperature is ~0.8 deg C. 

I don't know the tile ID (or EASE grid cell) nor the snow layer index associated with the offending snow values.  Because StieglitzSnow_snowrt() is called for just one tile at a time and doesn't know the ID or location of the tile it is working on, this requires more debugging work.  But finding out the tile ID and layer index by itself won't provide new insights.

Not sure about what to do next.  I wonder if the "snow heat content adjustment" (item 5 in the introductory comment above, abandoned in this PR) or the inconsistent "minimum value for snow cover area fraction" (item 7 in the introductory comment) might need revisiting.  Or do we need to get all of the inputs to StieglitzSnow_snowrt() so we can run just the snow routine for just the tile and time step where we get the offending values?  Do we have a way to run just StieglitzSnow_snowrt() without having to do lots of work?

gmao-rreichle:
@rdkoster: Since my last comment, I completed @biljanaorescanin's simulation through 1 June 2023 and printing all instances where the snow temperature is positive (without stopping).  It turns out that the only times when the snow temperature is positive includes six 7.5-minute time steps in March 2023, for a single tile at 124.17W, 39.95N.  The positive snow temperature occurs in the middle of StieglitzSnow_snowrt() here:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/5ebbb7b4d28a53dd1e7686e8312ee8086a2daffd/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L615
and here:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/5ebbb7b4d28a53dd1e7686e8312ee8086a2daffd/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L639
After these instances, the snow pack is further updated, and the snow temperature diagnosed at the end of the subroutine is **not** positive.  
I suppose we can address this by allowing the snow temperature to be positive in the offending calls above, but we'll need to take a closer look before committing to this approach.  We may also want to revise the structure of the sequence of if block in question.  It looks to me like these could be changed into an if/elseif/elseif... block, which would make it easier for humans to read the code.

rdkoster:
Rolf, I'm soon to leave for AMS, but I took a look at the places in the code you indicate.  Were the positive temperatures *inside* StieglitzSnow_calc_tpnow?  I ask because the temperatures in the parameter statements at the indicated locations are dummy variables and are never meant to be used.  If it's only inside StieglitzSnow_calc_tpsnow, it should be okay, because these calls have the sole purpose of dealing with the difficult transition from one energy balance regime (involving changes in temperature) to another (involving the melting of snow at constant temperature).  That is, these calls are only to figure out how much snow is frozen, for use in later calculations.  A positive temperature would only mean that we crossed the critical point, so that things would have to be adjusted.
I can take a better look later this week. (I won't have email today.)  Text me if you need to talk sooner.  -- Randy

From: Rolf Reichle ***@***.***>
Sent: Sunday, January 28, 2024 1:43 PM
Cc: Koster, Randal D. (GSFC-6101) ***@***.***>; Mention ***@***.***>
Subject: [EXTERNAL] [BULK] Re: [GEOS-ESM/GEOSgcm_GridComp] Non-0-diff fixes for Stieglitz snow code (PR #813)

CAUTION: This email originated from outside of NASA.  Please take care when clicking links or opening attachments.  Use the "Report Message" button to report suspicious messages to the NASA SOC.



@rdkoster<https://github.com/rdkoster>: Since my last comment, I completed @biljanaorescanin<https://github.com/biljanaorescanin>'s simulation through 1 June 2023 and printing all instances where the snow temperature is positive (without stopping). It turns out that the only times when the snow temperature is positive includes six 7.5-minute time steps in March 2023, for a single tile at 124.17W, 39.95N. The positive snow temperature occurs in the middle of StieglitzSnow_snowrt() here:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/5ebbb7b4d28a53dd1e7686e8312ee8086a2daffd/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L615
and here:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/5ebbb7b4d28a53dd1e7686e8312ee8086a2daffd/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90#L639
After these instances, the snow pack is further updated, and the snow temperature diagnosed at the end of the subroutine is not positive.
I suppose we can address this by allowing the snow temperature to be positive in the offending calls above, but we'll need to take a closer look before committing to this approach. We may also want to revise the structure of the sequence of if block in question. It looks to me like these could be changed into an if/elseif/elseif... block, which would make it easier for humans to read the code.

-

biljanaorescanin:
@gmao-rreichle  After testing with new changes this PR is, as expected ,zero diff to branch `bugfix/rreichle/snow_excs_thresholds`.

lcandre2 (approved):
Looks fine to me.

lcandre2:
After rereading the documentation on this pull request I have some thoughts on the continuously snow covered land tiles and the possibility of no glacier landice tiles. 

The latter option should not occur. The landice gridcomp uses a fixed 15m ice depth. No matter the melt and runoff, the ice remains 15m thick.

In the former case, e.g., 'faux-landice tiles,' this does occur naturally, in landice, in addition to the density  threshold for excs, we also use a thickness threshold of 15m, though it is rarely/never triggered due to the density threshold. An implementation like this may be relevant to prevent runaway snow accumulation in catchment.

gmao-rreichle:
@sdrabenh : This PR is finally ready for you.  Thanks for your patience! 

@biljanaorescanin ran the GCM tests.  As expected for a non-0-diff change, the tests runs finished ok but failed the comparisons against the Baseline results.  

@lcandre2 approved the PR for the landice team, but after her approval I made a very minor and trivially 0-diff commit (removed a few comment lines), which may have formally invalidated her approval.  

When you merge the PR, you may want to include the 5 bullet points of the PR "Summary" above in the commit message. 

@weiyuan-jiang @rdkoster

biljanaorescanin:
@sdrabenh  PR was tested for 1day AMIP; REPLAY and Inc REPLAY and all checkpoint/restart files were zero diff for both exp and baseline.
+ Regress job returned **success** status.

biljanaorescanin (approved):

sdrabenh (approved):

sdrabenh:
@gmao-rreichle please approve this PR if it is ready

gmao-rreichle:
> @gmao-rreichle please approve this PR if it is ready

@sdrabenh : I initiated the PR and cannot formally approve it.  @biljanaorescanin approved it for the land group.  It is ready for merging.

biljanaorescanin (approved):

gmao-rreichle:
I guess we'll still need approval from @GEOS-ESM/landice-team.   @lcandre2, @rcullath, can one of you please take a look and approve, or let me know if you have any questions or concerns?  Thanks!

sdrabenh:
> > @gmao-rreichle please approve this PR if it is ready
> 
> @sdrabenh : I initiated the PR and cannot formally approve it. @biljanaorescanin approved it for the land group. It is ready for merging.

@gmao-rreichle you are right of course. My oversight. I'll wait for @lcandre2 or @rcullath approval

PR #816, opened 2023-09-01
Fixes #815. Remove OpenMP C CMake

### LINKED ISSUE #815: Unneeded OpenMP C Requirement in makebcs
In moving to Arm Macs recently, we've found that Apple Clang does not do OpenMP (see https://mac.r-project.org/openmp/). In Baselibs, this means we can't compile xgboost (which isn't required *yet* in GEOS) and `makebcs` blocks the build of GEOS because of:

```cmake
esma_add_library(${this} SRCS ${srcs} 
  DEPENDENCIES MAPL GEOS_SurfaceShared GEOS_LandShared 
  esmf NetCDF::NetCDF_Fortran OpenMP::OpenMP_Fortran OpenMP::OpenMP_C)
```

I apparently added that a long time ago (see #183) and I think I was still a CMake naif at that point. 

But, the thing is, there is no OpenMP C code here. I'll ping @biljanaorescanin and @gmao-rreichle to confirm, but `omp.h` does not appear makebcs or in GEOSgcm or GEOSldas as far as I know.

I'm going to test and see if I remove this if things work. I'm 99.99% sure the model will be happy, but I want to make sure makebcs is as well, so I'll coordinate with @biljanaorescanin.

mathomp4:
Closes #815 

This just removes an unneeded requirement for OpenMP C. 

Keeping draft until I can work with @biljanaorescanin or @weiyuan-jiang or @gmao-rreichle to try the makebcs code (one of them must know how to run it! 😄 )


gmao-rreichle:
Thanks for looking into this, @mathomp4.
@biljanaorescanin can run makebcs tests.
Having said that, I'm not convinced that OpenMP is obsolete for makebcs.
Go to: `GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs`
Then grep for OMP (excluding some - but not all - unwanted matches):
`grep -i omp * |grep -v compute | grep -v README`
This returns a long list of OMP directives across several of the F90 utilities underpinning makebcs.  It's possible the OMP directives are all somehow disabled, but my gut feeling is that they are still in use.  
If it turns out that OpenMP is not use, we should clean up the source code. 
If OpenMP is still used, I have no clue how to address the issue that prompted this inquiry.

mathomp4:
@gmao-rreichle This would still allow OpenMP in Fortran. This change is just to remove OpenMP C. I mean, there are only two C codes in makebcs, `util.c` and `zip.c` and there is no OpenMP in them at all!

So, we have no reason to link against the OpenMP C libraries.

gmao-rreichle:
Sorry, I misunderstood.  The subtle diff between OpenMP Fortran and OpenMP C was lost on me.  Glad to hear.  Should be a non-issue then.

mathomp4:
> Its zero diff on runs I did. We can remove it.

Woo! I'm undrafting it so all can review.

gmao-rreichle (approved):

PR #818, opened 2023-09-05
Correct LONG_NAME of CDCR2

gmao-rreichle:
Correct LONG_NAME of Catchment model parameter CDCR2

PR prompted by a MERRA-2 user inquiry into wilting point and field capacity in MERRA-2.  It turned out that the long_name of CDCR2 is incorrect (or incomplete) in the MERRA-2 file specs and also in the latest (develop) version of Catchment model.

@rdkoster: Is this change ok, or do you recommend edits to the proposed LONG_NAME? 

@sdrabenh:  The same changes will need to be put on top of the R21C branch.

cc: @qliu


rdkoster:
I agree with the recommended long_name for cdcr2.

gmao-rreichle:
Reopening pull request after inadvertent closure

biljanaorescanin:
@gmao-rreichle I found one more instance in CatchmentRst.F90 so I changed that one as well.

biljanaorescanin (approved):

lcandre2 (approved):

PR #826, opened 2023-09-25
Add in missing ESMF_ConfigSetAttribute for SNOW_ALBEDO_INFO

mathomp4:
This is a bugfix on #687. My nightly tests for CatchCN40 and CatchCN45 were failing. Each was dying when we try to get the `SNOW_ALBEDO_INFO` from `MAPL`:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/499bf6f53f78fe0746e47933ca5de2ed7f37afa7/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM40_GridComp/GEOS_CatchCNCLM40GridComp.F90#L221

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/499bf6f53f78fe0746e47933ca5de2ed7f37afa7/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM45_GridComp/GEOS_CatchCNCLM45GridComp.F90#L226

The value was being read successfully from `GEOS_SurfaceGridComp.rc`, but didn't seem to be in the `MAPL` state. It looks like the concomitant change to `GEOS_CatchCNGridComp.F90` where we set the resource in `MAPL` from the CatchCN Internal state never got into the PR. It must have been there at some point but just mistakenly wasn't committed.

With this change, everything works and is zero-diff to my baselines (well, modulo the `CDCR2` name change in #818 ... which I am surprised did not trigger my comparison script).

Also, apologies for the whitespace changes. I still cannot figure out why vim on discover does this but not vim on my laptop. Same vimrc! If you hide whitespace, it looks sane:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/826/files?diff=split&w=1


biljanaorescanin:
@gmao-rreichle is bug I mentioned on chat.

gmao-rreichle (approved):
@mathomp4, @biljanaorescanin: Looks ok to me.  Having said that, I'm still a bit unclear as to what's going on here.  I understand the bug only impacted CatchCN simulations, and not the usual GCM tests required by @sdrabenh, so merging the original PR #687 made sense to Scott.  @biljanaorescanin, was the original PR #687 tested  with the LDAS tests?  Is it possible that we never had the missing line and never noticed because we didn't test with CatchCN?  Then again, how could the LDAS tests of last night all pass with this bug in place?

biljanaorescanin:
@gmao-rreichle we did run nightly tests on branch and all passed before merge. 
I also have output from CatchCN with flag set to zero from July and it was working.  Somehow that commit got lost or when we did cleanup and moved something around that become needed and wasn't before.  I don't have sandbox copy of model from that run, just output. I can share offline run link if someone wants to see. 
I only focused to flag ON for CatchCN testing before merge so I didn't repeat flag OFF before merge. 
Its really nice we have @mathomp4 running these test for CatchCN or this would have been lost.

gmao-rreichle:
@biljanaorescanin, @mathomp4 : I just invalidated my approval.  I'm not convinced the line is needed in GEOS_CatchCNGridComp.F90.  This file does not use SNOW_ALBEDO_INFO, and the line is present in the version-specific GridComp F90 files (GEOS_CatchCNCLM40GridComp.F90 and GEOS_CatchCNCLM45GridComp.F90).  Something else is going on.  Can you please enlighten me when/how Matt's test of CatchCN40 and CatchCN45 might have failed?  The automated LDAS tests of last night succeeded for all cases.

mathomp4:
@gmao-rreichle My CatchCN tests passed on 2023-09-20 and started failing on 2023-09-21. I guess I missed them because of other failures happening around the same time that were fixed.

gmao-rreichle:
Thanks, @mathomp4, for clarifying the difference between the MAPL object and the CATCH[CN]_INTERNAL_STATE.  I honestly don't quite understand what is and what should be available when and where.  My sense is that this requires a proper cleanup, which means waiting until @weiyuan-jiang is back in the office.  
For the record:
- The tests that failed are the GCM CatchCN  tests run by Matt.  The LDAS CatchCN tests all pass.   It is not clear why the LDAS tests are ok but the GCM CatchCN tests fail.
- The problem was missed prior merging #687 because Scott's GCM tests don't include CatchCN and the LDAS tests are all ok.  Going forward, we'll need to include Matt's GCM/CatchCN tests in our routine testing of GCM GC PRs.
- I'm not convinced that the fix proposed in the present PR is the right one.  My understanding is that SNOW_ALBEDO_INFO needs to be present in CATCH_INTERNAL_STATE because it is common to Catch and CatchCN.  That is, CatchCN simulations should query SNOW_ALBEDO_INFO from CATCH_INTERNAL_STATE (or whatever MAPL object that includes it), and not from CATCHCN_INTERNAL_STATE.  Note that the proposed fix does not have an equivalent line in GEOS_CatchGridComp.F90 (unless I missed it).  I might be wrong.

weiyuan-jiang:
@gmao-rreichle I think the fix @mathomp4 proposed is a correct one, especially for GEOSgcm run which is different from GEOSldas run (SURFRC: LDAS.rc).  In GEOSldas run, those params eventually go into MAPL object without extra effort.   For GEOSgcm run, in GEOScatcnclmxx setService, we either open GEOS_SurfaceGridCom.rc  to read them again or they are set in their parent's MAPL object.  In GEOScatch_GridComp, it indeeds reads GEOS_SurfaceGridCom.rc ( again, in GEOSldas run, LDAS.rc is used as GEOS_SurfaceGridCom.rc)

gmao-rreichle (approved):

sdrabenh (approved):

Issue #681, opened 2022-12-03
Fix routing/runoff for coupled model applications


sanAkel:
# Statement:
Routing table file (e.g., `CF0180x6C_TM1440xTM1080-Pfafstetter.TRN`) that gets generated by [`make_bcs script(s)`](https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/develop/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/make_bcs) puts run off in _wrong_ locations:
- In the middle of ocean (tile) points (some could be islands?).
- On land (tile) points.

See following figure, locations are marked in panels (c) and (d). Attached is a terrain map of the Indonesian seas (W Pac) region that can be compared to panel (d).
These have major consequences for the coupled model: causing it to crash and _leaks_ in the hydrological budget, to name a few.
Comments from @atrayano, @rdkoster and @gmao-rreichle will help.

@yvikhlya has been running [this utility](https://github.com/GEOS-ESM/GMAO_Shared/blob/main/GEOS_Util/coupled_diagnostics/g5lib/src/reroute_runoff/reroute.py) to _fix_ above problem to some extent.

# History:
- @amolod has asked for a fix since _many_ years. 
- More recently effort to _centralize_ boundary conditions (files): @biljanaorescanin and generation (source code) by  @weiyuan-jiang et al., requires us to take a fresh look at it.

# List of steps to generate a routing file that is currently used:
For e.g., `/discover/nobackup/projects/gmao/ssd/aogcm/atmosphere_bcs/Icarus-NLv3/MOM6/CF0180x6C_TM1440xTM1080_newtopo/CF0180x6C_TM1440xTM1080-Pfafstetter.TRN`

1. Run make_bcs **package**. This will produce a TRN file, but it directs runoff to the land, that is incorrect (see below figures).

2. Take the TRN file from above step 1 and run @yvikhlya's `reroute.py` utility (see code at `https://github.com/GEOS-ESM/GMAO_Shared/blob/main/GEOS_Util/coupled_diagnostics/g5lib/src/reroute_runoff/reroute.py`) which fixes the routing so it ends up being in the ocean.

[...]

PR #827, opened 2023-09-26
Routing Outlets in Boundary Conditions

biljanaorescanin:
Boundary conditions produce river routing file with coupled grid runs. ( *.TRN)
Current problem: This file right from the package on develop now would have to be manually adjusted if we wanted to use it with GCM coupled runs.  All routing outlets should end up in oceans but that is not the case with develop version.

This PR brings preprocessing tools used to create new “Outlet” file used during boundary conditions package runs to produce more correct river outlet information for coupled runs. Everyting will now end up in the ocean.
 
Once we conform this information we will add code changes to point and use “v2” of this Outlet file.

This PR is zero diff for AGCM but not zero diff for boundary conditions package. 

This PR is related to issue: #681 

@gmao-rreichle @sanAkel


zyj8881357:
> Boundary conditions produce river routing file with coupled grid runs. ( *.TRN) Current problem: This file right from the package on develop now would have to be manually adjusted if we wanted to use it with GCM coupled runs. All routing outlets should end up in oceans but that is not the case with develop version.
> 
> This PR brings preprocessing tools used to create new “Outlet” file used during boundary conditions package runs to produce more correct river outlet information for coupled runs. Everyting will now end up in the ocean.
> 
> Once we conform this information we will add code changes to point and use “v2” of this Outlet file.
> 
> This PR is zero diff for AGCM but not zero diff for boundary conditions package.
> 
> This PR is related to issue: #681
> 
> @gmao-rreichle @sanAkel

This version is only for 5 degree ocean grid.

gmao-rreichle:
@zyj8881357, @biljanaorescanin, thanks for putting this together.  Here are a couple of initial thoughts after a very quick glance:

1. We'll need a README file or some other form of documentation.  With a couple of exceptions, all f90 files are programs, and the only documentation is in the file names, with no comments or instructions at all inside the files.  This isn't sufficient for someone else to understand how to use the package to generate the outlet files. 
2. Re. your comment that this only works for the 5-deg ocean grid:  I might not understand this correctly, but if the package is specific to a given ocean model grid and resolution, and presumably a very coarse resolution that is for testing purposes only, there aren't yet outlet files that would work with S2Sv3 or what's in development.  Is this correct?  
3. Re. the previous item: My understanding is that the package first obtains the definition of the catchments on the raster grid and then determines the connection from each catchment to the ocean.  This step should be independent of the ocean model resolution.  The second step would then be to determine lat/lon values for the outlets that work with a particular ocean model grid and resolution.  I suspect the existing f90 programs and ncl tools can be easily assigned to one of these two steps, and this separation is just a matter of documentation.  If the package doesn't separate easily into sub-packages for the (ocean) resolution-independent step and the resolution-dependent steps, it should be modified so the package becomes modular in this way. 
4. Do we really want to produce (ocean) resolution-specific outlet files here, or would it be easier to stick with the current approach and have this package only produce a single outlet file with outlet lat/lon values that match the catchment raster file.  The moving of the outlet lat/lon points into suitable ocean grid points (ie, the resolution-dependent step) could then be coded within the make_bcs package, where the *.TRN file is produced.  The latter approach has the advantage of having the ocean model resolution information at hand as part of make_bcs processing.  If the ocean model resolution is addressed within the present preproc package, it'd be harder to maintain consistency between the preproc package and make_bcs. 

Not sure if this really makes sense, but it's a start for discussing how to move the PR forward.

zyj8881357:
> @zyj8881357, @biljanaorescanin, thanks for putting this together. Here are a couple of initial thoughts after a very quick glance:
> 
>     1. We'll need a README file or some other form of documentation.  With a couple of exceptions, all f90 files are programs, and the only documentation is in the file names, with no comments or instructions at all inside the files.  This isn't sufficient for someone else to understand how to use the package to generate the outlet files.
> 
>     2. Re. your comment that this only works for the 5-deg ocean grid:  I might not understand this correctly, but if the package is specific to a given ocean model grid and resolution, and presumably a very coarse resolution that is for testing purposes only, there aren't yet outlet files that would work with S2Sv3 or what's in development.  Is this correct?
> 
>     3. Re. the previous item: My understanding is that the package first obtains the definition of the catchments on the raster grid and then determines the connection from each catchment to the ocean.  This step should be independent of the ocean model resolution.  The second step would then be to determine lat/lon values for the outlets that work with a particular ocean model grid and resolution.  I suspect the existing f90 programs and ncl tools can be easily assigned to one of these two steps, and this separation is just a matter of documentation.  If the package doesn't separate easily into sub-packages for the (ocean) resolution-independent step and the resolution-dependent steps, it should be modified so the package becomes modular in this way.
> 
>     4. Do we really want to produce (ocean) resolution-specific outlet files here, or would it be easier to stick with the current approach and have this package only produce a single outlet file with outlet lat/lon values that match the catchment raster file.  The moving of the outlet lat/lon points into suitable ocean grid points (ie, the resolution-dependent step) could then be coded within the make_bcs package, where the *.TRN file is produced.  The latter approach has the advantage of having the ocean model resolution information at hand as part of make_bcs processing.  If the ocean model resolution is addressed within the present preproc package, it'd be harder to maintain consistency between the preproc package and make_bcs.
> 
> 
> Not sure if this really makes sense, but it's a start for discussing how to move the PR forward.

@gmao-rreichle @biljanaorescanin @sanAkel 
All the comments from Rolf have been addressed.
1. A readme.txt has been added.
2. Now the pre-processing package (GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/routing) only contains code to create the outlet locations in land. The moving from land to ocean is done by the mk_runofftbl.f90 in the mk_bcs package. The new mk_runofftbl.f90 (GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.f90) is also included in this branch.
3. The ocean resolution is not fixed or hard coded in the script. The new mk_runofftbl.f90 reads the ocean resolution that inputs to mk_bcs. It should work for any ocean resolution.

But the outlet locations have not been checked by Biljana or Santha. It should be tested before merging to the develop branch.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/routing/run.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+set -e
+
+INPUT=/discover/nobackup/yzeng3/work/outlets/inputs
```
gmao-rreichle, line 4: We can't point to a user directory here.  Please work with @biljanaorescanin to place the required input data into the appropriate sub-directory of the bcs_shared project directory

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90
```diff
@@ -383,17 +309,598 @@ subroutine write_route_file( fileB, NumTrans, SrcTile, DstTile, SrcFraction)
...
+integer,intent(in) :: nx,ny,nl,nt
+integer,intent(inout) :: lons(nx,ny),lats(nx,ny)
+integer,intent(in) :: rst_ocn_lnd(nx,ny)
+integer,intent(out) :: ns
+
+integer,allocatable,dimension(:) :: lonp,latp
+integer,allocatable,dimension(:,:) :: acc,np_map
+
+integer :: i,j,k,l,lonc,latc,flag,maxbak,status,num
+
```
gmao-rreichle, line 530: @zyj8881357, please fix the indentation across all files in this package

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/routing/Pfaf_to_2d_30s_land.f90
```diff
@@ -0,0 +1,59 @@
...
+use omp_lib
+use rwncfile
+implicit none
+
+character(len=100) :: var1="outlet_sinky_allcat"
+character(len=100) :: var2="outlet_sinkx_allcat"
+character(len=100) :: map="Pfafstetter_Greenland_real.nc"
+integer,parameter :: nc=291809
+integer,parameter :: nlon=43200
+integer,parameter :: nlat=21600
```
gmao-rreichle, line 12: @zyj8881357, please collect the parameter constants that are common across your f90 program files into a single module and annotate each constant with a brief comment about its meaning

gmao-rreichle:
@zyj8881357, @biljanaorescanin:  The software infrastructure of this package needs fixing up as follows:

1. Use Cmake build instead of "build" bash script.
2. Can we avoid adding the custom netcdf modules (rwncMod.F90, ncdioMod.F90) and use whatever is available in the GEOS environment? 
3. The "run.sh" script for Discover should be converted to python.

The above list is just a start, and I may be wrong on some or all of it.  I'll defer to the SI team on what needs changing and how. 
 The SI team should be able to assist with these fixes. 

@rdkoster @tclune @weiyuan-jiang

sanAkel:
> @zyj8881357, @biljanaorescanin:  The software infrastructure of this package needs fixing up as follows:
> 
> 
> 
> 1. Use Cmake build instead of "build" bash script.
> 
> 2. Can we avoid adding the custom netcdf modules (rwncMod.F90, ncdioMod.F90) and use whatever is available in the GEOS environment? 
> 
> 3. The "run.sh" script for Discover should be converted to python.
> 
> 
> 
> The above list is just a start, and I may be wrong on some or all of it.  I'll defer to the SI team on what needs changing and how. 
> 
>  The SI team should be able to assist with these fixes. 
> 
> 
> 
> @rdkoster @tclune @weiyuan-jiang 

If you guys are stuck, let me know. Above seem straightforward to me.

biljanaorescanin:
@weiyuan-jiang take a look at build maybe create CMakeList file

biljanaorescanin:
@gmao-rreichle  this PR was tested for:
1. Preprocessing functionality - do we create proper outlet file. (Compile `GEOSldas` and run in `preproc` dir `./run.py)` 
2. Do we create new sets of boundary conditions with fixed routing code. This PR fixes outlets on current (old) outlet file.
3. Same as 2 but using new outlet file created with preprocessing code. ( Boundary conditions created for bullet 2 and 3 differ just in `TRN` file) 
4. Coupled runs using new boundary conditions ( both sets were tested). To do this we had to remap restart files for `c12` and `c180` for `MOM6` runs.  Since current BCS package produces different number of tiles. 

@sanAkel @yvikhlya let me know if you want paths to my 1-day coupled runs. Or if I need to run something more?  Our plan is  after merge of this PR to add these boundary conditions with fixed `TRN` file to `v11` boundary conditions version.  And new outlet file will be added as `v2` to official path for input files.  

This PR is `NOT zero diff for Boundary Conditions Package` but is `zero diff for AGCM` since all changes are part of boundary conditions package.  With this PR there will be no need for manual editing of ocean tiles after bcs package is done.

zyj8881357:
> @zyj8881357, @biljanaorescanin: The software infrastructure of this package needs fixing up as follows:
> 
>     1. Use Cmake build instead of "build" bash script.
> 
>     2. Can we avoid adding the custom netcdf modules (rwncMod.F90, ncdioMod.F90) and use whatever is available in the GEOS environment?
> 
>     3. The "run.sh" script for Discover should be converted to python.
> 
> 
> The above list is just a start, and I may be wrong on some or all of it. I'll defer to the SI team on what needs changing and how. The SI team should be able to assist with these fixes.
> 
> @rdkoster @tclune @weiyuan-jiang
@biljanaorescanin @gmao-rreichle @sanAkel @yvikhlya 
In addition to the @biljanaorescanin's updates, all the comments from Rolf have been addressed. That means:
1. The run.sh was re-written with Python as run.py.
2. The code uses netcdf modules in a public dir of Discover (/usr/local/other/netcdf4/4.1.2/).
3. The "build" bash script is removed as it is no longer needed with the python script run.py. 
4. All other minor comments were addressed too.

zyj8881357:
> @weiyuan-jiang take a look at build maybe create CMakeList file

@biljanaorescanin @weiyuan-jiang I think this (using CmakeList) may be no longer needed. The preprocessing is a simple program. There is no sequence of compiling for different modules. Compiling those code in the python script run.py is enough.

gmao-rreichle:
> > @weiyuan-jiang take a look at build maybe create CMakeList file
> 
> @biljanaorescanin @weiyuan-jiang I think this (using CmakeList) may be no longer needed. The preprocessing is a simple program. There is no sequence of compiling for different modules. Compiling those code in the python script run.py is enough. Even the "build" bash script is no longer needed and was removed in the most updated branch.

@zyj8881357 :   The commit by @weiyuan-jiang earlier today took care of this comment.  We do need the CMakeLists.txt file to build the pre-proc code.  It's important to do this so we know when the pre-proc code might have issues with future compiler and hardware changes

sanAkel:
Guys, I will be on AL until Jan 2, 2024.

gmao-rreichle:
> Compiling those code in the python script run.py is enough.

@zyj8881357 : I hadn't noticed yet that run.py **builds** the pre-proc scripts!  This is different from what we usually do and should be changed.  The build of the pre-proc programs should happen with the build of the GCM or GEOSldas (ie., through CMakeLists.txt as committed by @weiyuan-jiang earlier today).  The run.py script should only execute the preproc scripts and programs as needed to create the outlet file.  It shouldn't build the preproc programs.

zyj8881357:
> > > @weiyuan-jiang take a look at build maybe create CMakeList file
> > 
> > 
> > @biljanaorescanin @weiyuan-jiang I think this (using CmakeList) may be no longer needed. The preprocessing is a simple program. There is no sequence of compiling for different modules. Compiling those code in the python script run.py is enough. Even the "build" bash script is no longer needed and was removed in the most updated branch.
> 
> @zyj8881357 : The commit by @weiyuan-jiang earlier today took care of this comment. We do need the CMakeLists.txt file to build the pre-proc code. It's important to do this so we know when the pre-proc code might have issues with future compiler and hardware changes

Got it! I just saw the new commit from @weiyuan-jiang. Weiyuan, thanks so much for the help!

zyj8881357:
> > Compiling those code in the python script run.py is enough.
> 
> @zyj8881357 : I hadn't noticed yet that run.py **builds** the pre-proc scripts! This is different from what we usually do and should be changed. The build of the pre-proc programs should happen with the build of the GCM or GEOSldas (ie., through CMakeLists.txt as committed by @weiyuan-jiang earlier today). The run.py script should only execute the preproc scripts and programs as needed to create the outlet file. It shouldn't build the preproc programs.


@gmao-rreichle @biljanaorescanin 
Thanks for this information! I have updated the run.py. Now it will not build the code, but only run the executes built with the GCM or GEOSldas.

weiyuan-jiang:
@zyj8881357 I have simplified the running scripts.  After it is built, you will find all the files needed for run_routing.py ( I changed some file name too) in the build directory. Before the run in that directory, you need to source g5_modules to have proper environment.

zyj8881357:
> @zyj8881357 I have simplified the running scripts. After it is built, you will find all the files needed for run_routing.py ( I changed some file name too) in the build directory. Before the run in that directory, you need to source g5_modules to have proper environment.

@weiyuan-jiang @gmao-rreichle @biljanaorescanin 
Thanks Weiyuan! I tested, the updated branch works well with the run_routing.py.

gmao-rreichle:
Thanks, @weiyuan-jiang, for fixing the build/install issue.  @zyj8881357: I noticed that the readme.txt file still refers to the old name of the run script ("run.sh").  This needs updating.  While we're at it, maybe rename "run_routing.py" to "run_routing_raster.py" to clarify the meaning of the run script a bit more.  Also, the "readme.txt" file refers to 7 programs, but the run script only uses 6 programs.  Is "read_riveroutlet_land.f90" obsolete?  If so, delete from "readme.txt".
There may be other updates needed in "readme.txt"

zyj8881357:
> Thanks, @weiyuan-jiang, for fixing the build/install issue. @zyj8881357: I noticed that the readme.txt file still refers to the old name of the run script ("run.sh"). This needs updating. While we're at it, maybe rename "run_routing.py" to "run_routing_raster.py" to clarify the meaning of the run script a bit more. Also, the "readme.txt" file refers to 7 programs, but the run script only uses 6 programs. Is "read_riveroutlet_land.f90" obsolete? If so, delete from "readme.txt". There may be other updates needed in "readme.txt"

@gmao-rreichle Thanks for this! I have updated the readme.txt and renamed the run script.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/routing/run_routing_raster.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
+
+#source g5_modules before run to get the necessary env
+
+import os
+import subprocess
+
+input_path = "/discover/nobackup/projects/gmao/bcs_shared/preprocessing_bcs_inputs/land/routing/"
```
gmao-rreichle, line 8: @zyj8881357 @biljanaorescanin : Don't we have the corresponding path for NAS now?  If so, please add it here with comments that clarify which path works where.  We don't need to auto-detect which machine we're on.  It's ok to have the user [un]comment the appropriate path when they want to run the python script.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/routing/get_outlets_catchindex.f90
```diff
@@ -0,0 +1,47 @@
...
+  use routing_constant,only : nc,ns,ng
+  implicit none
+  
+  integer,allocatable,dimension(:) :: msk,outid,mskall,final,finalall
+  
+  integer :: k,i,ntot
+  
+  ntot=nc+ng
+  allocate(msk(nc),outid(ns),mskall(ntot),final(nc),finalall(ntot))
+  open(77,file="inputs/Pfaf_msk.txt")
```
gmao-rreichle, line 12: The file "Pfaf_msk.txt" comes from the directory `[..]/preprocessing_bcs_inputs/land/routing`, which also contains several other txt and nc files.  None of these files is documented, and it is unclear what each of them contains exactly.  We need to add documentation to this directory.  Otherwise this preprocessing code is impossible to understand.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/routing/routing_constant.f90
```diff
@@ -0,0 +1,25 @@
...
+  integer,parameter :: nlat1m     =  10800  ! number of lon of the world grid with  1 min resolution
+  integer,parameter :: nlon_G     =   8400  ! number of lat of the Greenland grid (30 sec resolution)
+  integer,parameter :: nlat_G     =   4800  ! number of lon of the Greenland grid (30 sec resolution)
+  integer,parameter :: loni_min   =  12001  ! index of the lon start of the Greenland grid in the world grid (30 sec resolution)
+  integer,parameter :: loni_max   =  20400  ! index of the lon end of the Greenland in the world grid (30 sec resolution)
+  integer,parameter :: lati_min   =  16801  ! index of the lat start of the Greenland grid in the world grid (30 sec resolution)
+  integer,parameter :: lati_max   =  21600  ! index of the lat end of the Greenland in the world grid (30 sec resolution)
+  integer,parameter :: id_glac    = 290191  ! index of glacier tiles in the Pfafstetter.rst
+  integer,parameter :: id_lake    = 290190  ! index of lake tiles in the Pfafstetter.rst
+  integer,parameter :: id_landend = 290188  ! index of the last land tile in the Pfafstetter.rst
```
gmao-rreichle, line 18: I am very confused about this constant.  The name "**id**_landend" implies that it is a Pfafstetter **ID** (or name), not an **index** within a 1-d array.  

If "id_landend" is indeed an ID, then I'm surprised that "landend" ID is less than 291284, which is the Pfaf ID with the largest integer value.  My understanding was always that there are 291284 **unique** Pfafstetter watersheds.  If that's the case, something is wrong with these constants, and we're almost certainly not capturing all of the Pfafstetter catchments in the preprocessing/routing package.

It is possible that there are only 290188 **unique** Pfafstetter watersheds, in which case the distinction between "ID" and "index" becomes very important.   If this is the case, we need to clean up the variable names and make sure to document the distinction between "ID" and "index" very clearly.  The lack of documentation in the preprocessing_bcs_inputs/land/routing input directory does not help here.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/routing/get_outlets_catchindex.f90
```diff
@@ -0,0 +1,47 @@
+program main
+  
+  use routing_constant,only : nc,ns,ng
+  implicit none
+  
+  integer,allocatable,dimension(:) :: msk,outid,mskall,final,finalall
+  
+  integer :: k,i,ntot
+  
+  ntot=nc+ng
```
gmao-rreichle, line 10: Isn't "ntot" the same as "nall" from the "routing_constants" module?  Might need more cleanup.

gmao-rreichle:
@zyj8881357 @biljanaorescanin : I added a handful of in-line comments (see above) and commits.  Here's a link to the cumulative changes (excl white space changes):  https://github.com/GEOS-ESM/GEOSgcm_GridComp/compare/7c10381..ba2577f?w=1

zyj8881357:
> @zyj8881357 @biljanaorescanin : I added a handful of in-line comments (see above) and commits. Here's a link to the cumulative changes (excl white space changes): https://github.com/GEOS-ESM/GEOSgcm_GridComp/compare/7c10381..ba2577f?w=1

@gmao-rreichle Thanks for all your edits and comments! They really make sense. I will find some time to address them.

zyj8881357:
> @zyj8881357 @biljanaorescanin : I added a handful of in-line comments (see above) and commits. Here's a link to the cumulative changes (excl white space changes): https://github.com/GEOS-ESM/GEOSgcm_GridComp/compare/7c10381..ba2577f?w=1

@gmao-rreichle @biljanaorescanin 
I have changed the code and document based on Rolf's comments. Major changes are:

1. A new F90 script get_finalID_msk.f90 was added. It reads a well-documented Pfaf data in the Discover (/discover/nobackup/projects/gmao/bcs_shared/make_bcs_inputs/land/topo/v1/SRTM-TopoData/Pfafcatch-routing.dat, documented in the section 8.2.1 of https://gmao.gsfc.nasa.gov/pubs/docs/Mahanama804.pdf) and derives the needed routing files for the program. Now there are only four input data needed for the program. They are:
I. HydroSHEDS_drainage_area.nc:
The drainage_area dataset provided by the HydroSHEDS (https://www.hydrosheds.org/).
II. GreenlandID_30s.nc:
The Greenland ID map provided by @lcandre2 .
III. Greenland_outlets_lat.txt:
The latitude of the routing outlet for each Greenland catchment provided by @lcandre2 .
IV. Greenland_outlets_lon.txt:
The longitude of the routing outlet for each Greenland catchment provided by @lcandre2 .

2. The readme.txt was updated to include the input data description.

As I tested, for the updated version, there are 29 outlets (in ~22,000) are different with the previous version. That is because for the previous version, I did some manual corrections for the 29 outlets at a very early stage (I even forgot it until I checked this). Now I found the manual corrections are too subjective and not necessary, so I did not perform them in the get_finalID_msk.f90.

@biljanaorescanin At your convenience, could you please remove the files from input data /discover/nobackup/projects/gmao/bcs_shared/preprocessing_bcs_inputs/land/routing except the four listed above? Thank you!

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/routing/routing_constant.f90
zyj8881357: The id_landend as well as id_glac and id_lake were removed from the constant list. The code will read the til file Pfafstetter.til (and rst file Pfafstetter.rst) from bcs that is already provided in the Discover [/discover/nobackup/projects/gmao/bcs_shared] and determine the id_landend by the til file.
However, there are still 290188 land tiles in the Pfafstetter.til  and Pfafstetter.rst files. That means some tiles are removed when make_bcs is making the Pfafstetter.til and Pfafstetter.rst. Maybe a better way is to avoid using Pfafstetter.til and Pfafstetter.rst, and directly using SRTM_PfafData.nc to get the outlets map on 1-m resolution and then convert to 30s resolution to be consistent with the format of Outlet_latlon.43200x21600. I'll think about this...

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/routing/get_landocean_Greenland_real.f90
```diff
@@ -1,33 +1,38 @@
...
-  ret=nf_inq_varid(ncid,"lat",varid)
-  ret=nf_get_var_double(ncid,varid,lat)
-  ret=nf_close(ncid)
-  ret=nf_open("inputs/Pfafstetter.nc",0,ncid)
-  ret=nf_inq_varid(ncid,"data",varid)
-  ret=nf_get_var_int(ncid,varid,landocean)
-  ret=nf_close(ncid)
-  
-  
+  open(77,file="/discover/nobackup/projects/gmao/bcs_shared/legacy_bcs/Icarus-NLv5/Icarus-NLv5_EASE/SMAP_EASEv2_M09/rst/Pfafstetter.rst",form="unformatted",status="old")
```
biljanaorescanin, line 16: @zyj8881357 why do we use here specific Pfafstetter to read? What happens with CF grid where header is different? Can we have a more generic path with any set or it has to be EASE M09 from NLv5 set? @lcandre2 was EASE grid used to preprocess Greenland tiles?

zyj8881357: @biljanaorescanin Thanks for the comment! Please see my reply to Rolf above, the Pfafstetter.rst and Pfafstetter.til are not necessary. I am planing to use GEOS5_10arcsec_mask.nc or SRTM_PfafData.nc as the Pfaf ID map to create the Outlet_latlon.43200x21600.

gmao-rreichle: @zyj8881357 : If the `Pfafstetter.rst` is not necessary, then why does `get_landocean_Greenland_real.f90` read it? And why does it read the M09-specific file?  Is this all still work in progress?  After you long comment yesterday, I had assumed that you were done with your changes, and that we should take a look again.

zyj8881357:
@biljanaorescanin @gmao-rreichle I have updated the branch to remove its dependence on any Pfafstetter.rst. You can look into and test the code when you have time. Thank you!

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90
```diff
@@ -295,105 +243,682 @@ program Runoff
...
+    write(10) NumTrans
+    write(10) SrcTile
+    write(10) DstTile
+    write(10) SrcFraction
+    close(10)
+    
   end subroutine write_route_file
-! ----------------------
+  
+  ! ------------------------------------------------------------------------
```
gmao-rreichle, line 307: The documentation in the next 600+ lines is limited to the subroutine names.   For each subroutine that isn't very short/obvious, please add a statement with the objective and some explanatory comments about the variable names in the subroutine interface.  E.g., what does subroutine outlets_to_ocean() do exactly, and what are the input arguments?  What is the "file" that is passed in?  What exactly are "lons" and "lats"?   Longitudes/lats of what?  For more complex subroutines, some comments inside the subroutine are also needed.  E.g., what is "plats" as opposed to "plate"?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/routing/readme.txt
```diff
@@ -0,0 +1,48 @@
+v1 11/08/2023, Yujin Zeng
+
+The "preproc/routing" package is used for creating a 30-arcsec raster file with the locations of river outlets to the ocean.
+
+The output from this package is a binary file "Outlet_latlon.43200x21600".  
+
+The river outlets are located in land or landice tiles as defined in the raster file "Pfafstetter.rst" from the makebcs package.
```
gmao-rreichle, line 7: Is the raster file "Pfafstetter.rst" mentioned here still the right reference?  See the following comment from an earlier discussion:
"[zyj8881357]: However, I just realized using Pfafstetter.rst in a specific bcs is not a good idea last night, after my long comment (sorry about this...). Instead, more generic map such as SRTM_PfafData.nc or GEOS5_10arcsec_mask.nc should be used" 
Please make sure the readme.txt file is up-to-date

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/routing/get_finalID_msk.f90
```diff
@@ -0,0 +1,152 @@
...
+  integer,allocatable,dimension(:,:) :: pfaf_digit
+  integer*8,allocatable,dimension(:) :: res
+  integer,allocatable,dimension(:)   :: pfaf_last,pfaf_msk,code,behind
+  integer,allocatable,dimension(:)   :: first,last
+  
+  integer :: i,j,jj,k,p,down,cur,idx,num,ok,samed
+  integer :: fulli(12),fullj(12)
+  real    :: val(9)
+  
+  open(77,file="/discover/nobackup/projects/gmao/bcs_shared/make_bcs_inputs/land/topo/v1/SRTM-TopoData/Pfafcatch-routing.dat", form="formatted", status="old")
```
gmao-rreichle, line 17: The path here should be passed into the program via a command line argument.  It should be defined only in the python run script.  This applies to several of the Fortran programs in this directory.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/routing/get_outlets_land.f90
```diff
@@ -0,0 +1,90 @@
...
+  ret=nf_inq_varid(ncid,"latitude",varid)
+  ret=nf_get_var_double(ncid,varid,lat)
+  ret=nf_close(ncid)
+  
+  ret=nf_open("/discover/nobackup/projects/gmao/bcs_shared/make_bcs_inputs/land/topo/v1/SRTM-TopoData/SRTM_PfafData.nc",0,ncid)
+  ret=nf_inq_varid(ncid,"CatchIndex",varid)
+  ret=nf_get_var_int(ncid,varid,catchind)
+  ret=nf_close(ncid)
+  
+  ret=nf_open("inputs/HydroSHEDS_drainage_area.nc",0,ncid)
```
gmao-rreichle, line 36: Why is the relative path used here for the HydroSHEDS file but the full path for the files above? 
In any case, the path to the input files should be defined in the py run script and passed into each of the programs via a command line argument (or some other dynamic way).  It's not good practice to have the same path hardwired in many places.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/routing/run_routing_raster.py
gmao-rreichle: This comment has not yet been addressed.  Please add the path at NAS in a commented-out line, so users know where to find the data at NAS.   
Also, this should be the only place where the path is defined.  It should then be passed into the F90 programs.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/routing/readme.txt
```diff
@@ -0,0 +1,48 @@
...
+5. get_outlets_land_allcat.f90: 
+Assign outlet locations to all upstream catchments to create a 1d list showing the final x and y indexes for each catchment.
+
+6. get_landocean_Greenland_real.f90: 
+Insert the Greenland index map into the catchment index map.
+
+7. Pfaf_to_2d_30s_land.f90: 
+Transform the 1d list above to the unformatted Fortran binary file "Outlet_latlon.43200x21600" that can be read directly by "mk_runofftbl.F90" of makebcs.
+
+The input data are briefly described as follows:
```
gmao-rreichle, line 36: There should also be a readme file in the data directory.  Please work with @biljanaorescanin to add the info here into the data directory  (preprocessing_bcs_inputs/land/routing)

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90
```diff
@@ -295,105 +243,682 @@ program Runoff
...
+    real,                intent(out) :: wetMask(nx,ny)
+    
+    integer                          :: ncid, varid, ret
+    character(len=4)                 :: subname="read"
+        
+    ! try MOM6 first
+    
+    ret=nf90_open( trim(MAKE_BCS_INPUT_DIR) // "/ocean/MOM6/" // trim(res_MAPL) // "/MAPL_Tripolar.nc", 0, ncid )
+    
+    ! if MOM6 did not work, try MOM5,
```
gmao-rreichle, line 653: @biljanaorescanin : Do we need to add the MIT model here?

gmao-rreichle:
@zyj8881357 : I made a couple of minor commits and added a few new inline comments, see above.  There's also [this comment about the NAS path](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/827/files#r1541561386) that still needs addressing.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90
biljanaorescanin: I don't know. Maybe we should wait since we only have one NL3 resolution, `CF0090x6C_MITLLC5400x0015` .  + I'm not sure did Fahad and Atanas resolve all issues with that grid. I also rember from talking to Andrea there is a lot of manual adjustments to this grid, she wasn't sure it will ever really be part of boundary conditions package. I can fallow up with them and let you know.

zyj8881357:
> @zyj8881357 : I made a couple of minor commits and added a few new inline comments, see above. There's also [this comment about the NAS path](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/827/files#r1541561386) that still needs addressing.

Thank you very much for the editing and comments! I will go to address them.

zyj8881357:
> @zyj8881357 : I made a couple of minor commits and added a few new inline comments, see above. There's also [this comment about the NAS path](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/827/files#r1541561386) that still needs addressing.

@gmao-rreichle @biljanaorescanin I have addressed the comments above. You can review and test the branch when having time. Thanks!

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90
```diff
@@ -1,235 +1,183 @@
 program Runoff
-
+
```
gmao-rreichle, line 2: The program mk_runofftbl.F90 needs documentation.  What exactly does it do?  What are the limitations?  My understanding is that it is only used in make_bcs for the tripolar ocean grid because it only works for the tripolar ocean grid.  But I did not see this documented anywhere.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90
```diff
@@ -1,235 +1,183 @@
...
 
-  logical                :: adjust_oceanLandSea_mask = .false. ! default is .false.
   integer                :: nxt, command_argument_count
-  character*(128)        :: arg, &
-                            Usage = "mk_runofftbl.x CF0012x6C_TM0072xTM0036-Pfafstetter", &
-                            mapl_tp_file
-
-
+  character*(128)        :: arg
+  character*(128)        :: Usage = "mk_runofftbl.x CF0012x6C_TM0072xTM0036-Pfafstetter"
```
gmao-rreichle, line 28: The example geometry configuration "CF0012x6C_TM0072xTM0036-Pfafstetter" seems outdated.  Didn't we rename the ocean grid to something like "M6TP0072x0036"?  This would need to be updated throughout the file.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90
```diff
@@ -1,235 +1,183 @@
...
   integer, pointer       :: rst(:,:), SortArr(:,:), key(:)
   integer, pointer       :: srctile(:),  srcweight(:), dstweight(:), dsttile(:)
-  real, allocatable      :: SrcFraction(:), area(:), in(:), out(:)
+  real,    allocatable   :: SrcFraction(:), area(:), in(:), out(:)
   integer                :: i,j,k,l, Hash, HashC, ii,jj,kk
   integer                :: type, np,lnd, is,ie,ww
   integer                :: numtrans,  numclosed
   integer                :: status
-  character*100          :: file, fileT, fileR, fileO, fileB, fileBB
+  character*100          :: file, fileT, fileR, fileO, fileB
```
gmao-rreichle, line 19: The variable name "file" is confusing.  It's really a string that describes the grids associated with the atmosphere and ocean model configuration (e.g., "CF0180x6C_M5TP0720x0410").   The "-Pfafstetter" part of the string can be appended as needed when the string is used to create an actual file name, such as `fileT`.)
I know this confusing var name has been part of the make_bcs package, but in other parts o the package it's been clarified to make the code more accessible.   E.g., in mkCatchParam.F90 we use `Gridname`.
I suggest making a similar change here.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90
```diff
@@ -295,105 +243,759 @@ program Runoff
...
-  print *, trim(nf90_strerror(status))
-  print *, "Error in reading ocean mask file."
-  stop
-  end if
-  end subroutine check
-! -----------------------------------------------------------------
+!-------------------------------------------------------------------------
+! This subroutine gets the name of 'file_ocn' and 'file_ocn_lnd' from the input name 'file'.
+! It also gets resolution of the ocean domain.
+  subroutine get_domain_name(file,file_ocn,file_ocn_lnd,res_MAPL,nx_MAPL,ny_MAPL)
```
gmao-rreichle, line 393: This subroutine makes me nervous.  It parses the string "file" (or rather, the "Gridname") to extract substrings for the atmosphere and ocean grids and the associated ocean model resolution.  This can easily break if new grid definitions are added in the future. 
But the information that is needed here is in the 8 header lines of the *.til file and could presumably be read from there.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90
```diff
@@ -295,105 +243,759 @@ program Runoff
...
+    do i=1,nx
+       do j=1,ny
+          if(lons(i,j)/=-999.and.lats(i,j)/=-999)then
+             lonc=lons(i,j)
+             latc=lats(i,j)
+             if(acc(lonc,latc)==0)then
+                k=k+1
+                acc(lonc,latc)=1
+             else
+                acc(lonc,latc)=acc(lonc,latc)+1
```
gmao-rreichle, line 600: Is this "else" block needed?  It looks like the only information used here is whether acc(lonc,latc)==0.  The exact count for each element of acc does not seem to be needed, so why calculate it?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90
gmao-rreichle: This subroutine makes me think that mk_runofftbl.F90 works only the MOM5 and MOM6 Tripolar grids.  It may not be possible or desired to add the functionality for the MIT grid at this time, which is ok.  But we do need to document that mk_runofftbl.F90 only works for the MOM5 and MOM6 Tripolar grids.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90
```diff
@@ -295,105 +243,712 @@ program Runoff
...
+    do i=1,5
+      read(10,*)      
+    enddo
+    read(10,*)file_ocn_lnd
+    do i=100,1,-1
+      if(file_ocn_lnd(i:i).eq."-")then
+        file_ocn(1:i-1)=file_ocn_lnd(1:i-1)
+        exit
+      endif
+    enddo
```
gmao-rreichle, line 424: Is this do-loop effectively an identity operation?  It looks like file_ocn_lnd is the same before and after the do loop.

zyj8881357: The do-loop is used to get the file_ocn, not file_ocn_lnd.  file_ocn_lnd is got directly from *.til file, eg, TM1440xTM1080-Pfafstetter. This do-loop is used to extract the "TM1440xTM1080" from "TM1440xTM1080-Pfafstetter". It works as I run some test.

zyj8881357:
@gmao-rreichle I added the documentation in the mk_runofftbl.f90 and also addressed other comments. Thanks for checking!

gmao-rreichle:
@zyj8881357, I added two more commits:

1) comments/documentation update: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/827/commits/349249f61254b8a81f7a145f15ee0f0b11cd9944

2) cleanup of input path https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/827/commits/0d8b396b9c5571066b2da749430b8ff3216feff6

Between the two commits, I also merged "develop" into the branch.

Please double-check the commits to make sure I didn't mess up anything.   

cc: @biljanaorescanin @rdkoster

gmao-rreichle (approved):
This good to go from my perspective -- provided the GNU build error is fixed:

```
/root/project/GEOSgcm/src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90:437:12:

  437 |     nx_MAPL=""
      |            1
Error: Cannot convert CHARACTER(0) to INTEGER(4) at (1)
/root/project/GEOSgcm/src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90:438:12:

  438 |     ny_MAPL=""
      |            1
Error: Cannot convert CHARACTER(0) to INTEGER(4) at (1)
```

Should probably be something like:

nx_MAPL = -9999
ny_MAPL = -9999

zyj8881357:
> This good to go from my perspective -- provided the GNU build error is fixed:
> 
> ```
> /root/project/GEOSgcm/src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90:437:12:
> 
>   437 |     nx_MAPL=""
>       |            1
> Error: Cannot convert CHARACTER(0) to INTEGER(4) at (1)
> /root/project/GEOSgcm/src/Components/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90:438:12:
> 
>   438 |     ny_MAPL=""
>       |            1
> Error: Cannot convert CHARACTER(0) to INTEGER(4) at (1)
> ```
> 
> Should probably be something like:
> 
> nx_MAPL = -9999 ny_MAPL = -9999

@gmao-rreichle  Changed. Thank you!

gmao-rreichle (dismissed):
@sanAkel, @sdrabenh:  This PR is good to proceed as far as I'm concerned.  I'm not sure what level of testing is needed.  It's trivially 0-diff for running the GCM, and it's nearly trivially 0-diff for creating bcs.   (We only changed the generation of the routing trn and rst files.)  I assume that @biljanaorescanin has already tested the bcs generation.

Aside from the regression testing, what science validation would be needed, and who will do this?  This involves running the coupled model and understanding the output from an ocean sciences perspective, which isn't in the wheelhouse of the land group. 

I suppose we could merge the PR and retrofit the existing latest ("v12") bcs datasets with the new routing files.  That would make it easier to do the science testing, because the coupled model runs would then point to a stable bcs dir.  Alternatively, we could hold off on merging the PR and augmenting the existing bcs with the routing files until the science validation has caught up.  My concern with the latter approach is that during the process we need to keep the PR current with "develop", which is a distraction. 

Either way, we might have to wait for @biljanaorescanin to be back in the office, but that shouldn't keep us from planning ahead.

biljanaorescanin:
I've added to centralized location in `bcs_shared` coupled `MOM6` options so @sanAkel can run from gcm setup:
`CF0012x6C_M6TP0072x0036`
`CF0090x6C_M6TP0540x0458`
`CF0180x6C_M6TP1440x1080`

I will reach out with some questions about pre processing, creating `Outlet` part to @zyj8881357 .

biljanaorescanin:
Quick update after talking to Yujin I can confirm these bcs I created using this branch for v12 option still use old Outlet information. If we want to use new produced Outlet with preprocessing from this PR I need to add versioning to outlet file and rerun bcs package for Santha to use with both updates. 
Basically this part needs to be versioned: /discover/nobackup/projects/gmao/bcs_shared/make_bcs_inputs/land/route/

gmao-rreichle:
> Quick update after talking to Yujin I can confirm these bcs I created using this branch for v12 option still use old Outlet information. If we want to use new produced Outlet with preprocessing from this PR I need to add versioning to outlet file 

Thanks, but I'm a bit confused here.  Did the bcs package ever create the "correct" routing trn and rst files for existing bcs?  I thought that the routing trn and rst files that were used in the coupled model were custom-made (with some manual intervention).  

Having said that, versioning is definitely a good idea.  I suppose we can create a look-up table that assigns the version of the outlet-lat-lon file based on the bcs version?

I might have misunderstood

biljanaorescanin:
Yes, so existing PR fixes outlets on old Outlet file. This is what Santha tested with v11, it works for coupled runs. 
Now this next step is really what we want to do. We can use existing setup or use new Outlet info Yujin creates with preprocessing. 

Yujin with this PR adds 2 things to happen:
1. we can change what is our base Outlet info ( new one done with preprocessing seen here in PR; old is Randy's on develop)
2. we fix where outlets end up ( running this on NL3 bcs will also fix the problem and we don't need manual editing anymore) 

BCS were run with PR for v11 coupled and Santha tested them. Yesterday I created v12 but with old Outlet info. I just wasn't sure do we want to move to new Outlet info now that we tested moving of tiles. We were not sure that part works until Santha confirmed.

gmao-rreichle:
I'm not sure I understand...

> 2. we fix where outlets end up ( running this on NL3 bcs will also fix the problem and we don't need manual editing anymore)

a) When we run make_bcs from the present PR with the old outlet file, we get a routing trn file.  I understand this is what Santha was using (at least temporarily) along with v11 bcs.  But is this routing trn file also identical to what was produced with manual intervention?  If the two routing trn files are different, we cannot reproduce the old (pre-v11) routing trn file even with the old outlet lat/lon file as input.  @biljanaorescanin, can you please clarify? 

b) Is there a need for versioning of what Santha was using with v11?  This would be up to @sanAkel.

c) Either way, it's not wrong to associate a given version of bcs with a specific version of the outlet lat/lon file.  I suppose we have the following cases:

```
bcs version    -->  Outlet lat/lon file version
---------------------------------------------------
v12            -->  (new)  v2  (produced with Yujin's pre-processing routines)
v11            -->  (old)  v1  (Randy) 
otherwise      -->  n/a  (do not run routing module of make_bcs)
```

This versioning needs to be mk_runofftbl.F90, in the preprocessing code, and in the bcs_shared directory.

> Yesterday I created v12 but with old Outlet info

I don't think that's what we want.  @sanAkel will have to chime in on this.

biljanaorescanin:
> But is this routing trn file also identical to what was produced with manual intervention? 

No, this is different from what was done manually before. Files will differ. I did try this out of curiosity. We can run any set, even the old one, and create coupled boundary conditions that will run with GCM. I shouldn't have confused people with NL3; why I mentioned it was to say now that it fixes any coupled set we run from the package, so output doesn't need manual adjustment. This is just for all future sets.


I will remove MOM6 v12 from centralized location until we decide. I will keep them in my area for comparison if needed.

gmao-rreichle:
> No, this is different from what was done manually before. 

Ok, thanks.  This means that the old routing trn files with manual intervention cannot be reproduced by make_bcs and can be considered "legacy" data.  It is still a good idea to implement the versioning for the outlet lat/lon files.  The only remaining question then is which outlet lat/lon file we should associate with v11 and v12 bcs (see draft table in [comment above](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/827#issuecomment-2181589453)) @sanAkel, please let us know.

sanAkel:
👋 All.

I'll comment on questions that have been (explicitly) addressed to me in a few days. Appreciate your patience!

zyj8881357:
@biljanaorescanin @gmao-rreichle I have added the version information in mk_runofftble.f90

I slightly changed the version links as follows:
=======================================================
bcs version    -->  Outlet lat/lon file version
---------------------------------------------------
v12                 -->  (new)  v2  (produced with Yujin's pre-processing routines)
v11                  -->  (old)  v1  (produced with Randy's old file) 
otherwise       -->  n/a  (produced with Randy's old file, but do not move outlet locations to ocean)
=========================================================

gmao-rreichle:
@zyj8881357:

> otherwise --> n/a (produced with Randy's old file, but do not move outlet locations to ocean)

We should **not** produce routing files  for any bcs other than v11 and v12.  Some of the existing older bcs have custom-made routing files (manual intervention) that we cannot and do not want to replicate in make_bcs.  Routing files generated with make_bcs will be in conflict with the existing custom-made routing files.  That is, they should not be generated by make_bcs in the first place.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90
```diff
@@ -1,235 +1,251 @@
...
-  fileT = "til/"//trim(file)//".til" ! input
-  fileR = "rst/"//trim(file)//".rst" ! input
-  fileO = "til/"//trim(file)//".trn" ! output
-  fileB = "til/"//trim(file)//".TRN" ! output
+  if(trim(LBCSV)=="v11")then
+    OUTLETV="v1"
+  else if(trim(LBCSV)=="v12")then
+    OUTLETV="v2"
+  else
+    OUTLETV="others"
```
gmao-rreichle, line 111: We have a config section in subroutine init_bcs_config() in rmTinyCatchParaMod.F90.  I think it would be better to include the string "OUTLETV" there.  Then it's much more obvious what bcs "v11" and "v12" etc use for the outlets.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_cube.py
```diff
@@ -53,9 +53,11 @@
       /bin/mv rst/Pfafstetter-M.rst rst/Pfafstetter.rst
       bin/CombineRasters.x -f 0 -t {NT} {OCEAN_VERSION}{DATENAME}{IMO}x{POLENAME}{JMO} Pfafstetter >/dev/null
       bin/CombineRasters.x -t {NT} CF{NC}x6C{SGNAME} {OCEAN_VERSION}{DATENAME}{IMO}x{POLENAME}{JMO}-Pfafstetter
-      bin/mk_runofftbl.x CF{NC}x6C{SGNAME}_{OCEAN_VERSION}{DATENAME}{IMO}x{POLENAME}{JMO}-Pfafstetter
       setenv OMP_NUM_THREADS 1
-      if ({SKIPLAND} != True) bin/mkCatchParam.x -x {NX} -y {NY} -g CF{NC}x6C{SGNAME}_{OCEAN_VERSION}{DATENAME}{IMO}x{POLENAME}{JMO}-Pfafstetter -v {lbcsv}
+      if ({SKIPLAND} != True) then
+         bin/mkCatchParam.x -x {NX} -y {NY} -g CF{NC}x6C{SGNAME}_{OCEAN_VERSION}{DATENAME}{IMO}x{POLENAME}{JMO}-Pfafstetter -v {lbcsv}
+         bin/mk_runofftbl.x -g CF{NC}x6C{SGNAME}_{OCEAN_VERSION}{DATENAME}{IMO}x{POLENAME}{JMO}-Pfafstetter
```
gmao-rreichle, line 59: @zyj8881357 : Thanks for moving the runoff config info into the land bcs version case statement within rmTinyCatchParaMod.F90.  Have you tried running the code?  I wonder if it can work.  I would have expected that we need to pass {lbcsv} into mk_runofftbl.x via a command line argument and then use
`call init_bcs_config (trim(LBCSV))       ! get bcs details from version string`
from within the mk_runofftbl program.  I might be wrong, but at a quick glance I'm not sure the version info makes it to where it's needed.

zyj8881357:
Hi Rolf,

I have realized the code does not work and know the solution same as your
suggestion. I push the code just for testing purposes. Please do the check
it after I make a comment in the github. Thanks so much!

zyj8881357:
@gmao-rreichle I have fixed the problem. The program should work now.

gmao-rreichle:
@zyj8881357 , @biljanaorescanin :
The code now looks like it should work.  The documentation needed cleaning up.  Please double-check my latest commit. 

@biljanaorescanin, please generate select v12 bcs to see if the results match those in the bcs_shared dir.  If that works out, I think the PR is ready for Scott. 

@biljanaorescanin:  The `[..]/bcs_shared/make_bcs_inputs/land/route/` directory may still need some cleanup.  I don't understand why the there are "v1" files in the `[..]/bcs_shared/make_bcs_inputs/land/route/` directory and then links to these files in the `./v1` subdirectory.  Ideally, the files should all be inside the v1 subdirectory only.  In this case, I don't think backward compatibility with older make_bcs versions matters.  The older make_bcs versions don't produce sensible routing TRN files anyway.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mk_runofftbl.F90
```diff
@@ -319,81 +323,705 @@ program Runoff
...
+    
+    if (present(msg)) then
+       write(6,*)'ENDRUN: ', msg
+    else
+       write(6,*) 'ENDRUN: called without a message string'
+    end if
+    
+    stop 
+    
+  end subroutine endrun
```
gmao-rreichle, line 766: As a note for later: This subroutine is used selectively.  If we introduce it here, why is it not applied consistently?  It could  be used in place of the various explicit "write & stop" commands throughout the code.  There is probably also be a MAPL function that can do the same.  Flagging this for the SI team (@weiyuan-jiang)

zyj8881357:
> .

@gmao-rreichle The edit looks fine in your latest commit. Thank you!

zhaobin74:
Testing of coupled MOM6 at C180 with V12 BCS(and new outlets) also works.

gmao-rreichle (approved):

sdrabenh (approved):

mathomp4 (approved):

sdrabenh (approved):

PR #854, opened 2023-11-15
added support for CICE6 thermo state rewind for replay

zhaobin74:
This PR adds a customized entry point method to support restoring CICE6 state variables such that they can be reset before ```corrector``` step starts. The reason for saving and restoring CICE6 states is that ```predictor``` step updates these states via callbacks even though the ```seaice_gridcomp``` is not running.

This PR also fixed a bug in  ```Record``` method which doesn't handle dumping MAPL states properly.

This PR is zero diff and it depends on https://github.com/GEOS-ESM/GEOSgcm/pull/702


yvikhlya (approved):

tclune (approved):
cmake change ok.

sanAkel:
@sdrabenh I got a confirmation from @zhaobin74 that he verified in both configurations:
- With replay: it worked for him to _rewind_.
- Without replay: answers preserved.

I'll not be reviewing this PR.

sdrabenh:
> @sdrabenh I got a confirmation from @zhaobin74 that he verified in both configurations:
> 
> * With replay: it worked for him to _rewind_.
> * Without replay: answers preserved.
> 
> I'll not be reviewing this PR.

Sounds good. My AGCM tests passed... but obviously not a test of CICE

PR #695, opened 2022-12-29
Rename coupled grids

biljanaorescanin:
Rename coupled grids to: 
CF0180x6_<XX>_TP1440x1080; XX = M5, M6 and “TP” stands for Tripolar grid (Ref: https://doi.org/10.1006/jcph.1996.0136)

@sanAkel let me know does this work for you?
I've tested it and it works for us and as you said it is more understandable. And most important no more overwriting with same output name for different oceans. Once we resolve issue with c180 runs we can work on PR to use updated info with gcm_setup. 

We are changing in case of T4MOM6 at C180 output from `CF0180x6C_TM1440xTM1080` to `CF0180x6C_M6TP1440x1080` .
Tile name will go from `CF0180x6C_TM1440xTM1080-Pfafstetter.til` to `CF0180x6C_M6TP1440x1080-Pfafstetter.til`

@yvikhlya this is so we don't overwrite outputs of MOM5 and MOM6 with BCS Package for 1440x1080 runs.

PR #860, opened 2023-11-29
add ocean version to tile file header info

biljanaorescanin:
In one of the previous merges, PR #695, we also changed the tile header information to update the ocean grid name. 
To minimize confusion and help people better understand tile files and what they see in boundary conditions, we are adding an ocean version descriptor to coupled tile file header info.  This tile header change is not a bug; users can run experiments and use these new sets of boundary conditions by adjusting the `AGCM.rc` file; this PR is to make it cleaner and easier to understand; with this change, it is easier for `MAPL` and `AGCM.rc` to recognize coupled grid based on tile header information.  

_**In legacy, how it used to be: (MOM6/CF0090x6C_TM0540xTM0458/CF0090x6C_TM0540xTM0458-Pfafstetter.til)**_

  841477  291284  43200  21600
     2
 PE90x540-CF
    90
    540
 **PE540x458-TM**
    540
    458

_**Now on develop (Nov 2023): 
( NL3/geometry/CF0090x6C_M6TP0540x0458/CF0090x6C_M6TP0540x0458-Pfafstetter.til )**_

  841477  291284  43200  21600
     2
 PE90x540-CF
    90
    540
 **PE540x458-**
    540
    458

_**With this PR :**_

  841477  291284  43200  21600
     2
 PE90x540-CF
    90
    540
 **PE540x458-M6TP**
    540
    458

Changes for this PR are zero diff trivial since all changes are within the boundary conditions package (we are just adding info to the header of the coupled tile file as seen in the example above)

Boundary conditions in `bcs_shared` area created with these new ocean names will be updated to match information package will produce in tile file.  @mathomp4 @gmao-rreichle

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_shared.py
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_shared.py
@@ -50,7 +50,7 @@ cd geometry/{GRIDNAME}/
 if( {TRIPOL_OCEAN} == True ) then
 cat > sedfile << EOF
 s/CF{NC}x6C/PE{nc}x{nc6}-CF/g
-s/{OCEAN_VERSION}{DATENAME}{IMO}x{POLENAME}{JMO}-Pfafstetter/PE{imo}x{jmo}-{DATENAME}/g
+s/{OCEAN_VERSION}{DATENAME}{IMO}x{POLENAME}{JMO}-Pfafstetter/PE{imo}x{jmo}-{OCEAN_VERSION}/g
 EOF
 sed -f sedfile       {GRIDNAME}{RS}.til > tile.file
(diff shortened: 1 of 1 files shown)


gmao-rreichle:
@biljanaorescanin:  This looks like a bug fix to me.  It looks like #695 inadvertently omitted the trailing string ("TM") in the ocean grid descriptor, or else the dash would have also been removed.  I suppose it is true that bcs generated with this bug can still be used by adjusting AGCM.rc, but it may be better to recreate (or manually fix) the coupled model bcs once the present PR has been approved and merged.  
Having said that, the ocean grid descriptor (e.g., "PE540x458-M6TP") in the til file header seems inconsistent with the corresponding string in the til file name ("M6TP0540x0458").  Is this something that also needs to be addressed?  Although this inconsistency seems to also be present for atm grid strings, and making this consistent would presumably be much more far-reaching.  Oh well.

biljanaorescanin:
All looks good now in tile file. 
@mathomp4  after we merge this PR I will edit files on bcs_shared.

sdrabenh (approved):

gmao-rreichle (approved):

PR #864, opened 2023-11-30
WIP: Separate ncar_gwd into its own repo

mathomp4:
This PR is in conjunction with 

It moves ncar_gwd out into its own repo per a request of @wmputman 

CI will fail due to chicken-and-egg issue with the main repo. Though maybe I can use this as an opportunity to try and logic this out...


mathomp4:
@atrayano should not care about this

PR #867, opened 2023-12-05
Added connectivity for a few fields from MOIST to CHEM

mmanyin:
Added connectivity, from MOIST to CHEM, for CAPE, BYNCY and INHB, for use in lightning-related work.
This is zero-diff.


mathomp4:
Thank you for making GNU happy. Always fun with Connectivities :)

sdrabenh (approved):

PR #875, opened 2023-12-22
Bugfix/atrayano/uw regression fix

### LINKED ISSUE #874: uwshcu.F90 has a regression bug
The limiter uses a scheme involving cnvtr. But the pointer to this variable is passed in from an export of Moist (CNPCPRATE). 

The regression (start/stop and 0 IAU increments) fails, unless tests are using either LM=72, or BACM_1M, or UW_OPTIONS: NONE

Inside GEOS_UW_Interface.F90 there is logic that if (LM==72) the default for JASON_UW is set to true (false otherwise). And with JASON_UW on, one of the parameters we set there effectively chooses a different limiter. Of course, the regression passes when we choose BACM_1M since we are not using UW shallow water code,

atrayano:
This fixes #874. The choice of the limiter was suggested by @wmputman. We need to coordinate with @narnold1


narnold1 (commented):
The cnvtr dependence was originally added to reduce entrainment in the presence of deep convective cold pools. GF cold pool development was later set aside and we switched to a precip dependence.  Removing this is fine, but we should verify it doesn't negatively impact skill.

atrayano:
@narnold1 thanks for your quick response! I wanted to hear your opinion, and this was the reason I made this PR "draft". Let see how this impacts the skill. If we need to keep your "cntvr" dependence, we could fix the regression issue differently

sdrabenh:
@atrayano @narnold1 are we waiting on something additional, or is this PR ready to go?

narnold1:
I'll run a test this afternoon to verify that the effect is small.

narnold1:
In replay experiments, this results in a small improvement in 850 mb temperature and small degradation in water vapor.  I suspect the overall change is either neutral or slightly beneficial.

PR #876, opened 2024-01-05
Changes needed to allow GNU build of wave GEOS

mathomp4:
Working off of the components in https://github.com/GEOS-ESM/GEOSgcm/pull/724, this PR along with https://github.com/GEOS-ESM/WW3/pull/5 contains changes needed to allow Wave GEOS to build with GNU. 

Most changes here are CMake changes to reflect our current way of CMake such as:
- Less use of `INC_XXX` and using targets instead
- Support for all mepo styles (prefix, postfix and plain)
- Protecting an Intel only flag

There is a slight code change in `GEOS_UMWMGridComp.F90` because GNU does not allow the `I` descriptor to be used without a width, so I used `I0`. (To see this easier, hide whitespace, my editor automatically removes trailing spaces.)

I can say this builds with both GNU and Intel, and a simple Intel AMIP run is zero-diff. I have no idea how to run the wavemodel, so hopefully @adarmenov can test this PR (along with the one in ww3).


adarmenov (approved):
Thanks @mathomp4 !

Issue #878, opened 2024-01-12
Turbulence imports QCTOT as cloud fraction


narnold1:
QCTOT is exported from MoistGridComp as total condensate, but Turbulence imports it expecting cloud fraction.  The Turb cloud fraction import is only used by SHOC, so this bug has no effect for default settings.  Solution is to change the Turb import to FCLD, with connectivity changed appropriately in PhysicsGridComp.

PR #880, opened 2024-01-12
Fix for issue #878, incorrect cloud fraction import to TurbGridComp

narnold1:
This changes the QCTOT import in TurbulenceGridComp to FCLD, to fix issue #878.  The import is only used in SHOC, so other than changing the Turb Import state, this fix has no effect in default model configurations.


sdrabenh:
@narnold1 I'm closing this PR as it looks like this bugfix already came in through @wmputman branches. Nothing to merge.

sdrabenh:
Nothing to merge.

PR #881, opened 2024-01-12
Restore the MTRI bundle

mmanyin:
Added back the Bundle of 3D Tendencies from MOIST, which had been lost when the 2D Tendency Bundle was introduced. This is zero diff.


mathomp4:
Adding @bena-nasa as well just so he knows. Heck, maybe he helped fix this, but he's the MCHEMTRI expert I go to ...

mmanyin:
For the record, an easy demonstration of the MTRI 3D tendency is to include a line like this in HISTORY:
     'MTRI%TR::aoaIM'   ,  'PHYSICS'      ,  'aoa_MOISTTEND',

sdrabenh (approved):

PR #883, opened 2024-01-22
NWP patches

sdrabenh:
Changes introduced by @wmputman . This PR is non-zero-diff for the default configuration and aims to reduce forecast RMSE in Krok for skills closer to Jason v10.25.x and x0048. The also added back the necessary exports for TLM/ADJ.


wmputman (approved):
The key non-zero diff changes are to bring the cloud liquid radii calculations back to as in Jason for BACM_1M, a switch is introduced to use the new schemes for GFDL_1M as desired.  The many changes in GWD are related to new schemes which are off by default.

PR #889, opened 2024-01-25
Update LONG_NAME of DO3DT_ANA

sdrabenh:


mathomp4 (approved):

mathomp4:
Not sure this will ever pass CI due to images, etc.. @sdrabenh I wonder if I should just remove any CI here? Obviously you wouldn't make a change here without testing.

PR #892, opened 2024-01-30
Feature/aoloso/hybrid 012924

aoloso:
Removed dependencies on FMS and MPP for gfdl_cloud_microphys.F90 and replaced with MAPL/Fortran Standard alternatives. These was necessary to avoid FMS conflicts in the mixed hybrid implementation i.e. MPI/OpenMP for Dyncore and MPI only everywhere else.


sdrabenh:
@aoloso looks like there are still some build issues here

sdrabenh:
This PR wasn't pulled in initially since @wmputman wanted to personally review this. However, I'm closing for now since this is outdated.

PR #900, opened 2024-02-13
disable CatchmentCNCLM45 (LSM_CHOICE=3)

gmao-jkolassa:
This PR disables the use of CatchmentCNCLM45 (LSM_CHOICE=3) and supersedes #898.

This particular model version has lots of science issues, and NCAR discourages the use of CLM45, from which the carbon and nitrogen cycle components of CatchmentCNCLM45 are borrowed.

The PR removes the LAND_PARAMS=CN_CLM45 option from the surface resource file and removes the LSM_CHOICE=3 option in select case/if blocks to make sure  the code does not run with this model version.  A more thorough source code cleanup will be needed at a later time. 

Related PRs:

https://github.com/GEOS-ESM/GEOSldas/pull/707 

This PR is zero-diff. 

See also planned cleanup of source code https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/907 

cc: @gmao-rreichle , @biljanaorescanin


gmao-rreichle (approved):
PR only comments out a few if blocks for CatchmentCNCLM45.  I'd call it trivially 0-diff.  CI build and runs work.  @sdrabenh, please let us know if you need more formal testing.

sanAkel:
why don't you just remove this choice?

gmao-rreichle:
> Instead of commenting it, why don't you just remove this choice?

A more thorough code cleanup will be done when we replace CatchCNCLM4.5 with CatchCNCLM5.1.  This is a quick safety to make sure nobody can bypass the corresponding changes in setup.

gmao-rreichle:
> @gmao-rreichle in that case, we'll need an issue that goes with this PR. The issue I suggest would have: `A more thorough code cleanup will be done`. Does that make sense?

@sanAkel : Done.

sanAkel (approved):

lcandre2 (commented):
Fine by me.

sdrabenh (approved):

sdrabenh:
Fine with me too

PR #918, opened 2024-03-28
Helfand Monin-Obukhov derivatives for GEOSldas offline runs

biljanaorescanin:
This PR adds Helfand M-O derivatives **needed for GEOSldas offline runs**. 
Almost all GEOSldas offline runs are run using Louis scheme choice due to the lack of these derivatives for Helfand M-O option.  This is not in sink to what AGCM uses which is Helfand always. 

`Surface layer turbulence scheme`
`0 : Louis                  (MERRA, Fortuna-DAS, SMAP NRv4/4.1/5/7.2/8.1/9.1/10.0)`
`1 : Helfand Monin-Obukhov  (Fortuna-AR5, Ganymed, Heracles, Icarus-3_2, MERRA-2)`

GEOSagcm=>CHOOSEMOSFC: 1

Derivatives added are:
1.  DSHSBT - derivative of sensible heat flux w.r.t. Tc (ground temperature)
2.  DHSDQA - cross derivative of sensible heat flux w.r.t. Qc (ground humidity)
3.  DEVSBT - derivative of latent heat flux w.r.t. Qc (ground humidity)
4.  DEDTC - cross derivative of latent heat flux w.r.t. Tc (ground temperature)

This PR is NOT zero diff for GEOSldas runs with CHOOSEMOSFC: 1 but is zero diff for AGCM runs. 

NOTE: Long test runs are in progress I will update PR once testing is done.


GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90
```diff
@@ -3438,6 +3541,29 @@ subroutine RUN1 ( GC, IMPORT, EXPORT, CLOCK, RC )
         CALL helfsurface( UWINDLMTILE,VWINDLMTILE,TA,TC(:,N),QA,QC(:,N),PSL,PSMB,Z0T(:,N),lai,  &
```
weiyuan-jiang, line 3541: is this intentional to call this function three times? or it should be
if (incl_HelfandMo_extra_derives ==0) then
    call once
else if(incl_HelfandMO_extra_derivs ==1)  then
   call your addition ( twice)
 endif

weiyuan-jiang:
The crash is due to small_TC which is way to small.  TC is around 270. and the smal_tc = 0.0001 is just roundoff error. This eventually leads to NaN of TC and crash.  It should work if small_TC=0.1  . But I think the better way is to use relative perturb factor, so each tile has the same perturbation.   
small_TC(:) = factor * TC(:)
small_QC(:) = factor* QC(:)

When I choose factor = 0.01, the program has run for 1 year.  /gpfsm/dnb34/wjiang/LDAS_exp/model/exp_helfand_sles15_small_factor

gmao-rreichle:
Replaced with new PR #999

PR #920, opened 2024-04-22
fix and update cice6 parameters

zhaobin74:
This PR updated some run time CICE6 params based on recent test runs.

* 540x458
   -  fix update_ocn_f (should be true)
   -  set default shortwave to ```dEdd``` scheme
   -  change default dEdd params
   -  do not output dailies by default
* 1440x1080
   -  fix update_ocn_f (should be true)

This PR is zero diff for AMIP and coupled running with CICE4


sanAkel (approved):
I have no test or comment on these, but approving to move it along.

sdrabenh (approved):

PR #929, opened 2024-05-10
Bugfix/narnold1/turb budget fix

narnold1:
This fixes an issue of double-counting in the Turbulence kinetic energy budget.  The surface drag contribution, KESRF, was being included in the diffusive term, KEINT.  This properly separates the two for the budget calculation.


wmputman (commented):
@narnold1 , just verifying that you really want to add SRFDIS to this export and not subtract?

narnold1:
> @narnold1 , just verifying that you really want to add SRFDIS to this export and not subtract?

Yes, the bug is that the SRF contribution is already included in INTDIS and KEINT above.  So here we need remove it from KEINT (by adding instead of subtracting).  Similarly, no need to include the SRF contribution in KETRB, because it was already included along with 'INT' above.  

This fix leaves the 'INTDIS' export untouched so as to remain 0-diff (it's used in PhysGridComp to compute dissipative heating, but not directly used in M21C HISTORY), but eventually we should fix INTDIS as well.  I have code to do that for a future non-0-diff update to develop.

sdrabenh (approved):

PR #944, opened 2024-05-24
make_bcs fix for bad HWSDv1.21 data in Argentina "peatland" 

gmao-rreichle:
In GEOS bcs versions NL5 and v11 (after the introduction of PEATMAP and PEATCLSM), a small region in Argentina (~38W,60W) is wrongly classified as peatland.  The reason for the wrong classification are unrealistically high values of organic carbon content (OC) in the HWSDv1.21 input data for soil mapping unit (SMU) #12302.

make_bcs uses preprocessed soil properties data that contain merged data from HWSD and STATSGO2 for the US.  The SMU in question is spread across two of the SoilProperties_*.nc files: H12V06 and H13V06.  These two files were patched by replacing the clay, sand, and OC values for the 0-30cm and 30-100cm layers in the SMU in question with values obtained manually from v2 of HWSD.  

The present PR enables make_bcs to use these revised SoilProperties_*.nc data to generate v12 bcs.

PR only touches make_bcs and is thus **trivially 0-diff for GCM and LDAS simulations**. 

PR was successfully 0-diff tested for generation of existing bcs (NL3 for EASE and cube-sphere tile spaces) on 31 May 2024 by @biljanaorescanin 

v12 bcs were verified by @gmao-rreichle through direct comparison of bcs files and the SMAP Nature Run v11.3 simulation.

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/GEOS_SurfaceGridComp.rc
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/GEOS_SurfaceGridComp.rc
@@ -120,7 +120,7 @@
 #           - backfilled with global land average snow albedo where unavailable
 #           - must use compatible bcs version that includes MODIS-based snow albedo (e.g., v06, v08, v09, ...)
 #           - NOTE: bcs v06, v08, and v09 used approximate averaging of MODIS-based snow albedo to tile space;
-#             later bcs versions employ more accurate, raster-based averaging       
+#                   bcs v11 and v12 employ more accurate, raster-based averaging.
 #
 # GEOSagcm=>SNOW_ALBEDO_INFO: 0
 # GEOSldas=>SNOW_ALBEDO_INFO: 0
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/CMakeLists.txt
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/CMakeLists.txt
@@ -51,4 +51,9 @@ ecbuild_add_executable (TARGET mkEASETilesParam.x SOURCES mkEASETilesParam.F90 L
 
 install(PROGRAMS clsm_plots.pro create_README.csh  DESTINATION bin)
 file(GLOB MAKE_BCS_PYTHON CONFIGURE_DEPENDS "./make_bcs*.py")
+list(FILTER MAKE_BCS_PYTHON EXCLUDE REGEX "make_bcs_shared.py")
 install(PROGRAMS  ${MAKE_BCS_PYTHON} DESTINATION bin)
+
+set(file ./make_bcs_shared.py)
+configure_file(${file} ${file} @ONLY)
+install(PROGRAMS ${CMAKE_CURRENT_BINARY_DIR}/${file} DESTINATION bin)
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/create_README.csh
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/create_README.csh
@@ -620,7 +620,7 @@ APPENDIX I - mkCatchParam input options and log ................................
 	 file name: soil_param.dat
 	 do n = 1, ${NTILES}
 _EOI_
-if( $mysoil == HWSD ) then
+if( $mysoil == HWSD || $mysoil == HWSD_b ) then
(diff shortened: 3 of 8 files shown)


gmao-rreichle:
@mathomp4 : 
As a side show in the present PR, we're trying to get "make_bcs.py" to work on SLES15/Milan.  My initial approach was to simply remove the #SBATCH constraint (https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/944/commits/fb06070619e4bb8061df99026102d247eca180fa).
But this only works on SLES12.  On SLES15, the job that is submitted by make_bcs.py either hangs or crashes unexplicably with library errors.  We apparently can get this to work by adding an explicit (non)-constraint 
`#SBATCH --constraint=sky|cas|mil`  (https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/944/commits/8fd5fe311a4451859d49da1e2cfd1e5c22a766cf).  For SLES15, using just `#SBATCH --constraint=mil` also works.  
That is, it looks like the SLES15 job won't start or run without the explicit #SBATCH constraint directive, which doesn't really make sense to me.  Maybe this is something for NCCS to sort out? 
cc: @biljanaorescanin @weiyuan-jiang

mathomp4:
> @mathomp4 : As a side show in the present PR, we're trying to get "make_bcs.py" to work on SLES15/Milan. My initial approach was to simply remove the #SBATCH constraint ([fb06070](https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/fb06070619e4bb8061df99026102d247eca180fa)). But this only works on SLES12. On SLES15, the job that is submitted by make_bcs.py either hangs or crashes unexplicably with library errors. We apparently can get this to work by adding an explicit (non)-constraint `#SBATCH --constraint=sky|cas|mil` ([8fd5fe3](https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/8fd5fe311a4451859d49da1e2cfd1e5c22a766cf)). For SLES15, using just `#SBATCH --constraint=mil` also works. That is, it looks like the SLES15 job won't start or run without the explicit #SBATCH constraint directive, which doesn't really make sense to me. Maybe this is something for NCCS to sort out? cc: @biljanaorescanin @weiyuan-jiang

You might ping NCCS, but now that I think about it, it's possible an sbatch with no constraint currently only targets `sky|cas` and not `mil`. I guess in gcm_setup/gcm_run we always specify the constraint, though that's because we ask the user what they want to run on (makes the math for oserver, etc. easier)

weiyuan-jiang:
> #SBATCH --constraint=sky|cas|mil

I don't think this is correct. The executables built on SLES12 cannot be run on SLES15.

mathomp4:
> > #SBATCH --constraint=sky|cas|mil
> 
> I don't think this is correct. The executables built on SLES12 cannot be run on SLES15.

Oh yeah. That is true. That's the reason I have the `BUILT_ON_SLES15` variable in CMake. I use it in `gcm_setup` and @weiyuan-jiang used it in the LDAS setup as well I think.

This is probably the "right" thing to do.

gmao-rreichle:
Thanks, @weiyuan-jiang  and @mathomp4 for the commit and advice, much appreciated.
@biljanaorescanin : When you get a chance, please run just one set of bcs on both SLES12 and SLES 15 to make sure there isn't a syntax error somewhere.

weiyuan-jiang:
> > > #SBATCH --constraint=sky|cas|mil
> > 
> > 
> > I don't think this is correct. The executables built on SLES12 cannot be run on SLES15.
> 
> Oh yeah. That is true. That's the reason I have the `BUILT_ON_SLES15` variable in CMake. I use it in `gcm_setup` and @weiyuan-jiang used it in the LDAS setup as well I think.
> 
> This is probably the "right" thing to do.

@biljanaorescanin said the string is not replaced. Does that mean the trick does not apply to GEOSgcm_GridComp? @mathomp4

biljanaorescanin:
After last fix everything works now with no errors. I've run few diff sets of bcs on both sles12 and sles15

mathomp4:
> @biljanaorescanin said the string is not replaced. Does that mean the trick does not apply to GEOSgcm_GridComp? @mathomp4

Ah. You figured it out before I could reply. Yeah, it's `configure_file` :)

gmao-rreichle:
@wmputman @sdrabenh @biljanaorescanin : This PR is now ready for merging:
1. The PR only touches make_bcs and is trivially 0-diff for AGCM and LDAS simulations.  
2. The PR was successfully 0-diff tested for the generation of existing bcs.
3. The new "v12" bcs have been verified in the offline (land-only) SMAP Nature Run system.  In the offline system, the changes are strictly limited to the small region in Argentina where the erroneous soil parameters were replaced with more sensible values.  I'm happy to share a few slides if you're interested.  In the GCM, a change in one location will obviously impact the rest of the world, but it's fair to assume that the difference between the "v12" and "v11" bcs is not going to change the GCM's climate noticeably. 

cc: @rdkoster

biljanaorescanin (approved):

bena-nasa (approved):

PR #946, opened 2024-05-31
bug fix: remove double-counting of heat corr (StieglitzSnow.F90)

gmao-rreichle:
In the rare circumstance when more snow would need to be sublimated than is available in the top layer, the residual energy accounting term was added twice, once here and again ~30 lines further down.

This appears to be a bug.  It was discovered during the review of the land energy budget for M21C.  

Fixing this bug is (probably) a non-0-diff change for the restarts.  It should be non-0-diff for at least some model diagnostics. 

Further examination is on-going. 

cc: @rdkoster

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Shared/StieglitzSnow.F90
@@ -671,7 +671,7 @@ contains
           if(-dw > wesn(1) ) then
              dw = -wesn(1)
              evap = -dw/(dts*areasc)
-             hcorr=hcorr+(lhflux-evap*alhv)*areasc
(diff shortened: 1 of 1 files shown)


gmao-rreichle:
The change is 0-diff for restarts and diagnostics in the full suite of GEOSldas nightly tests run by @gmao-rreichle on 17 Jun 2024.
This means that the condition indeed happens very rarely, if ever. 

The next step is to run the GEOSgcm tests once @biljanaorescanin  is back in the office.

cc: @elakkraoui @rdkoster @lcandre2

biljanaorescanin:
I am getting zero diff for all GCM testing. 1 day AMIP, REPLAY and INC REPLAY are all zero diff branch vs stock ( develop GEOSgcm_GridComp). The regression test is a success as well.

biljanaorescanin (approved):

lcandre2 (approved):

sdrabenh (approved):

PR #951, opened 2024-06-12
Update FMS CMake target

mathomp4:
As we move to FMS in Baselibs, we shouldn't use the old `fms_r4` and `fms_r8` targets anymore as they are non-standard. Instead we move to `FMS::fms_r4` and `FMS::fms_r8`.

This in combination with other PRs in GEOSgcm v12 testing.

NOTE: CI is failing since it isn't picking up the v12 testing branch.


mathomp4:
Superseded by #987

PR #952, opened 2024-06-12
Even cleaner version of MGB.

dbarahon:
Cleaned up version of MGB


mathomp4:
I wonder if this is worth putting into `develop`? Or should we concentrate on getting it into v12?

dbarahon:
> I wonder if this is worth putting into `develop`? Or should we concentrate on getting it into v12?
> 
> ETA: or I guess both?

Needs to be in develop. Current MGB code does not work and it is really outdated. Also it would be easier to merge to v12.

mathomp4:
@dbarahon If you are happy, you can press the "Update branch" button so that it is up-to-date with `develop`. Note: Probably still need your help then getting this into v12

PR #969, opened 2024-07-31
bugfix for M21C: restore MKTILE for RUNOFF (GEOS_SurfaceGridComp.F90)

gmao-rreichle:
The RUNOFF export (runoff averaged across all tile types) has been available in Surface GC for a long time, but RUNOFF was never written out from Surface GC.  (Only land and landice specific exports were written out.)
The proposed M21C file specs include RUNOFF in the "flx" collection.  As of now, the R21C branch only produces NaNs for RUNOFF.  This is a bug that was presumably introduced with the  integration of the runoff routing module.
The bug fix proposed in the present PR fixes this problem.
**HOWEVER, this fix is only suitable if the runoff routing is not exercised, which should be ok for M21C.** 
For "develop", a more elaborate fix will be needed.

The fix was tested with an AGCM run by @bena-nasa.   Zero-diff testing etc will still be needed.

@bena-nasa @sdrabenh @elakkraoui @lcandre2 @atrayano


lcandre2 (approved):
Looks good.

elakkraoui (approved):

sdrabenh (approved):

PR #974, opened 2024-08-07
remove filling of WET1 (a.k.a. GWETTOP) with 1s over non-land

gmao-rreichle:
Filling WET1 (a.k.a. GWETTOP) seems to be a bug.  This PR removes the filling. 

PR should be 0-diff except for GWETTOP diagnostic.  Leaving decision about need for testing to GCM gatekeepers.

@biljanaorescanin @sdrabenh @elakkraoui @mathomp4 @atrayano @bena-nasa @rdkoster 

PS: History of this chunk of code courtesy of @mathomp4:

From the blame, it's been there since the initial commit in Git:

[https://github.com/GEOS-ESM/GEOSgcm_GridComp/blame/95b31b7dc478be622d1593daba8aaf2f971416de/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90#L8465-L8470](https://gcc02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FGEOS-ESM%2FGEOSgcm_GridComp%2Fblame%2F95b31b7dc478be622d1593daba8aaf2f971416de%2FGEOSagcm_GridComp%2FGEOSphysics_GridComp%2FGEOSsurface_GridComp%2FGEOS_SurfaceGridComp.F90%23L8465-L8470&data=05%7C02%7Crolf.h.reichle%40nasa.gov%7C4eba46ea711b4b9d2ddd08dcb6d9e565%7C7005d45845be48ae8140d43da96dd17b%7C0%7C0%7C638586294201308041%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=DpOf43C%2BWDi12UONK8oKLWkMyVyQP1kpOfZxuFKEgP4%3D&reserved=0)

and looking in CVS, it's been like that since rev 1.130 from 2007:

```
revision 1.130

date: 2007/09/24 02:21:08;  author: trayanov;  state: Exp;  lines: +883 -884

Conversion of GEOS-5 tag GEOSagcm-Eros-beta7p13 to MAPL
```

I went back further in time and found the code used to be:

 ```
      if( associated(WET1) ) then

          call GET_POINTER( EXPORT, FRLAND,  'FRLAND' , RC=STATUS )

          VERIFY_(STATUS)

          WET1 = FRLAND*WET1 + 1-FRLAND

      endif
```

but then Max changed it to something like present day in rev 1.109:

 
```
revision 1.109

date: 2006/05/25 18:37:50;  author: f4mjs;  state: Exp;  lines: +222 -38

mjs: correction to u10m, etc, added u50m, etc, added land diagnostics, modified meaning of snow depths.
```

So not much help with the CVS message for the why.

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOS_PhysicsGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOS_PhysicsGridComp.F90
@@ -1262,6 +1262,10 @@ contains
                                                        RC=STATUS  )
      VERIFY_(STATUS)
 
+     ! NOTE: GOCART's dust code expects WET1 to have all the cells with MAPL_UNDEF
+     !       (aka not land) to be replaced with 1.0. We want WET1 to have
+     !       MAPL_UNDEF over non-land points, so we need a separate export to pass
+     !       to GOCART which is WET1 with all non-land points set to 1.0.
      call MAPL_AddConnectivity ( GC,                              &
         SRC_NAME  = [ 'WET1_FOR_CHEM' ],                          &
         SRC_ID      = SURF,                                       &
@@ -1269,6 +1273,7 @@ contains
         DST_ID      = CHEM,                                       &
                                                        RC=STATUS  )
      VERIFY_(STATUS)
+
      if (DO_CO2CNNEE == 1) then
         call MAPL_AddConnectivity ( GC,                           &
              SHORT_NAME  = (/'CNNEE'/),                           &
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
@@ -8462,14 +8462,6 @@ module GEOS_SurfaceGridCompMod
       endif
 
 
-! Fill WET1 over non-land points
-!-------------------------------
-
-      if( associated(WET1) ) then
-          where(WET1 == MAPL_UNDEF) WET1 = 1.0
-      endif
-
-
 ! Fill imports/exports for OBIO
 !-------------------------------
       if((DO_OBIO/=0) .OR. (ATM_CO2 == ATM_CO2_FOUR)) then

### CODE REVIEW

**Comment by mathomp4** [comment]
I'll do a test now to make sure it is zero-diff for the state. I mean, it should be but still, best to check!

**Comment by mathomp4** [comment]
I am blocking this as for now as it changes the state. The issue seems to be Dust in GOCART relies on the behavior. I see two places. First, [in the Ginoux emission code](https://github.com/GEOS-ESM/GOCART/blob/cf91268b2188e5166ce66e3794b1746ef3f04dcb/Process_Library/GOCART2G_Process.F90#L517-L537):

```fortran
!     Spatially dependent part of calculation
!     ---------------------------------------
      do j = j1, j2
         do i = i1, i2
            if ( oro(i,j) /= LAND ) cycle ! only over LAND gridpoints


            w10m = sqrt(u10m(i,j)**2.+v10m(i,j)**2.)
!           Modify the threshold depending on soil moisture as in Ginoux et al. [2001]
            if(gwettop(i,j) .lt. 0.5) then
               u_thresh = amax1(0.,u_thresh0* &
               (1.2+0.2*alog10(max(1.e-3,gwettop(i,j)))))


               if(w10m .gt. u_thresh) then
!                 Emission of dust [kg m-2 s-1]
                  emissions(i,j,n) = (1.-fraclake(i,j)) * w10m**2. * (w10m-u_thresh)
               endif
            endif !(gwettop(i,j) .lt. 0.5)
         end do ! i
      end do ! j
      emissions(:,:,n) = Ch_DU * du_src * emissions(:,:,n)
    end do ! n
```
This seems "safe" as `MAPL_UNDEF` is quite a bit larger than 0.5

But then we have in [DryDeposition](https://github.com/GEOS-ESM/GOCART/blob/cf91268b2188e5166ce66e3794b1746ef3f04dcb/Process_Library/GOCART2G_Process.F90#L2634-L2651):

```fortran
      if(present(fraclake) .and. present(u10m) .and. present(v10m) .and. &
         present(radius) .and. present(rhop) .and. present(gwettop)) then


!      Calculate the threshold velocity for dust emissions
       u_thresh0 = 0.13 * sqrt(rhop*grav*2.*radius/rhoa(i,j,km)) &
                        * sqrt(1.+6.e-7/(rhop*grav*(2.*radius)**2.5)) &
              / sqrt(1.928*(1331.*(100.*2.*radius)**1.56+0.38)**0.092 - 1.)
       w10m = sqrt(u10m(i,j)**2. + v10m(i,j)**2.)


!      Calculate the coefficient for resuspension
       if(oro(i,j) .eq. OCEAN) then
        R2 = 1.
       else
        R2 = fraclake(i,j)+(1.-fraclake(i,j)) &
                           *( gwettop(i,j)+(1.-gwettop(i,j)) &
                             *exp(-max(zero,(w10m-u_thresh0))))
       endif
      endif
```

Here moving from 1.0 to MAPL_UNDEF is a change to reckon with.

lcandre2 (dismissed):

mathomp4:
Perhaps we could make a "new" diagnostic from `WET1` that is for history purposes that has the undefs...but then I suppose that doesn't reflect what `WET1` looks like in the model technically.

The other "fix" is changes to GOCART to allow it to handle undef as it handled the 1.0 case? not sure which is "better"

gmao-rreichle:
> Perhaps we could make a "new" diagnostic from WET1 that is for history purposes that has the undefs...but then I suppose that doesn't reflect what WET1 looks like in the model technically.

Thanks, @mathomp4, for finding this out so quickly.  What a mess!  I don't think we should get into changing whatever is done inside the dust code.  But rather than introduce a new EXPORT for the land outputs, I would much prefer fixing the WET1 export and creating a new one, say WET1_FOR_DUST, that is filled with 1s and is connected to the dust code.   This would be much more transparent.  I really dislike the fact that the model abuses WET1 in this way without a speck of in-code documentation.  I'm not confident that I would get the renamed connection to the dust code right.  Can you please take a crack at this?  Thanks!!!

mathomp4:
> > Perhaps we could make a "new" diagnostic from WET1 that is for history purposes that has the undefs...but then I suppose that doesn't reflect what WET1 looks like in the model technically.
> 
> Thanks, @mathomp4, for finding this out so quickly. What a mess! I don't think we should get into changing whatever is done inside the dust code. But rather than introduce a new EXPORT for the land outputs, I would much prefer fixing the WET1 export and creating a new one, say WET1_FOR_DUST, that is filled with 1s and is connected to the dust code. This would be much more transparent. I really dislike the fact that the model abuses WET1 in this way without a speck of in-code documentation. I'm not confident that I would get the renamed connection to the dust code right. Can you please take a crack at this? Thanks!!!

@gmao-rreichle Okay. I looked at it with @bena-nasa and I think it should be fairly simple. Ugly, but simple.

mathomp4:
@gmao-rreichle et al I think what is here now should be good. I made a new `WET1_FOR_CHEM` export and pass that to chemistry. Meanwhile, the `WET1` export is now UNDEF over the ocean.

Note: My guess is the "right" thing is to only have one `WET1` and fix up GOCART, but for now, this works.

mathomp4:
Oh. And my editor on Discover is brutal with end-of-line whitespace, which is why this PR looks a bit messy. 😄

gmao-rreichle:
Thanks, @mathomp4!  Looks good to me now

biljanaorescanin (approved):

PR #983, opened 2024-08-16
add functionality to pre-commit

twicki:
## Purpose

Enhance featuers used in the pre-commit hook.

This PR adds two new hooks to run in pre-commit: codespell and pydocstyle. codespell checks for spelling errors and pydocstyle ensures conventions in documentation are upheld.

## Code changes:
**Add new hooks**
- `.pre-commit-config.yaml`
  - add new hooks

## IMPORTANT

I did not change all the files to match these hooks yet. This works should be done once we've consolidated this list and would be a PR into this branch before merging to `dsl/develop`

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/pyMoist/.pre-commit-config.yaml
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/pyMoist/.pre-commit-config.yaml
@@ -8,18 +8,21 @@ repos:
     rev: 20.8b1
     hooks:
     - id: black
+      description: Format Python code
       additional_dependencies: ["click==8.0.4"]
 
 -   repo: https://github.com/pre-commit/mirrors-isort
     rev: v5.4.2
     hooks:
     - id: isort
+      description: Group and sort Python imports
       args: ["--profile", "black"]
 
 -   repo: https://github.com/pre-commit/mirrors-mypy
     rev: v1.4.1
     hooks:
     - id: mypy
+      description: Perform static type analysis of Python code
       name: mypy-pyMoist
       args: [
         --ignore-missing-imports,
@@ -35,6 +38,8 @@ repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v2.3.0
     hooks:
+    - id: check-ast
+    - id: check-case-conflict
     - id: check-toml
     - id: check-yaml
     - id: end-of-file-fixer
@@ -44,6 +49,7 @@ repos:
     rev: 3.9.2
     hooks:
     - id: flake8
+      description: Check Python code for correctness, consistency and adherence to best practices
       name: flake8
       language_version: python3
       args: [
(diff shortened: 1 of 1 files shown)


romanc (commented):
Love the codespell initiative. Not sure how well this will fly given all the cryptic abbreviations in the current code. It's probably a good thing though. Will force us to increase readability while we port. We probably want to have our own dictionary since the "--ignore-word-list" will be rather big.

If we gonna look at linting & formatting, we might want to look at the new [ruff](https://docs.astral.sh/ruff/) linter/formatter. According to the docs, it's a drop-in replacement for flake8, isort, and black. It comes with [pre-commit integration](https://docs.astral.sh/ruff/integrations/#pre-commit) and is orders of magnitudes faster than the traditional pipeline (its written in rust).

`gt4py`  is also using it and if we decide to move, it would be nicer to work in the NDSL (not here) repo because that pulls in gt4py as a submodule and currently I have to deactivate "format on save" option because vscode is will try to format it with the black formatter (apparently there's no way in vscode to have different formatters for the same language depending on folder structure). `ruff` is mono-repo friendly and will use the [closest ruff configuration](https://docs.astral.sh/ruff/configuration/#config-file-discovery) allowing different config files per (submodule) folder.

FlorianDeconinck:
Good point about codespell. We do want to make code more readable, but we have to be careful about doing that to early and losing people in the port. My heart is split on the subject. I think a more conservative approach would be to keep most variable names the same, change the function names, then go through the port with a few scientists for further explanation and refactor then? Maybe it's too complex of a strategy.

As much as I'd like to burn `flak8` and dace on its grave, I am worried about how "drop-in" `ruff` is going to be. @twicki take an hour and see, this is a small codebase, easier to handle than NDSL

twicki:
ok, tested ruff. Works pretty well. Easy to handle config file. Is added now

twicki:
Additionally I created a codespell ignore file if we want to get rid of the appreviations we have in here.

I still strongly believe that if we want to build a sustanable codebase our variables should not be named `TE` but at least `temp` or `temperature`. At last we no longer live in a world where long variable names kill us because of either punchcards or displays that are limited to 60 characters,

romanc:
> ok, tested ruff. Works pretty well. Easy to handle config file. Is added now

nice, with that you should be able to remove not only `black`, but also `flake8`  and `isort`  from the `pre-commit`  config, or do you just want to use the formatter?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/pyMoist/.pre-commit-config.yaml
```diff
@@ -4,22 +4,18 @@ default_language_version:
...
-    rev: 20.8b1
-    hooks:
-    - id: black
-      additional_dependencies: ["click==8.0.4"]
-
 -   repo: https://github.com/pre-commit/mirrors-isort
     rev: v5.4.2
     hooks:
     - id: isort
+      description: Group and sort Python imports
```
romanc, line 11: we should also be able to get rid of `isort` if we go with `ruff`

PR #992, opened 2024-09-09
v12: Fix issue with gfdl qs_table

mathomp4:
## NOTE:

I'm keeping draft right now just to make sure the CI works with this. If it does, I will undraft.

---

Today, I was trying to fix the v12 CI in GEOSgcm and once I patched up some Docker image issues, it crashed with:

```
forrtl: error (182): floating invalid - possible uninitialized real/complex variable.
libpthread-2.31.s  0000153E0848A910  Unknown               Unknown  Unknown
GEOSgcm.x          00000000072F3DA1  gfdl2_cloud_micro        4105  gfdl_cloud_microphys.F90
GEOSgcm.x          00000000072E1B01  gfdl2_cloud_micro        3473  gfdl_cloud_microphys.F90
```

Prints show that `table` is filled with `NaN` which then dies during:
```fortran
        table (i + 1200) = wice * table (i + 1200) + wh2o * esupc (i)
```
when it tries to use it on the right-hand side.

This is true with even the latest v12 updates from @wmputman and @atrayano. So, I took a look, and yep a bug. To wit in gfdl_cloud_microphysics we have in `qs_table` (truncated to look at `table`):
```fortran
    ! -----------------------------------------------------------------------
    ! compute es over ice between - 160 deg c and -40 deg c.
    ! -----------------------------------------------------------------------

    do i = 1, 1200
        table (i) = e00 * exp (fac2)
    enddo

    ! -----------------------------------------------------------------------
    ! compute es over water between - 40 deg c and 102 deg c.
    ! -----------------------------------------------------------------------

    do i = 1, es_table_length-1200
        if (i <= 400) then
            esupc (i) = esh40
        else
            table (i + 1200) = esh40
        endif
    enddo

    ! -----------------------------------------------------------------------
    ! derive blended es over ice and supercooled water between - 40 deg c and 0 deg c
    ! -----------------------------------------------------------------------

    do i = 1, 400
        table (i + 1200) = wice * table (i + 1200) + wh2o * esupc (i)
    enddo
```
The comments make it seem like we want to do:

1. -160 → -40
2. 0 → 102
3. -40 → 0

This seems to be slightly altered from the ["canonical" GFDL code](https://github.com/NCAR/ccpp-physics/blob/main/physics/MP/GFDL/module_gfdl_cloud_microphys.F90) which does:

1. -160 → 0
2. -20 → 102
3. -20 → 0

But notice the issue. The first loop in our code fills `table(1)` to `table(1200)`. The second one fills `table(1601)` to `table(2621)`. The third tries to fill `table(1201)` to `table(1600)` by blending. 

We *NEVER* set `table(1201)` to `table(1600)` in the first loop to then blend with in the third.

My proposed solution is to extend the first loop to 1600 so that `table` is filled from 1200 to 1600 with something for the blend.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,10 @@ workflows:
           #baselibs_version: *baselibs_version
           repo: GEOSgcm
           checkout_fixture: true
-          mepodevelop: true
+          # V12 code uses a special branch for now.
+          fixture_branch: feature/sdrabenh/gcm_v12
+          # We comment out this as it will "undo" the fixture_branch
+          #mepodevelop: true
           persist_workspace: true # Needs to be true to run fv3/gcm experiment, costs extra
 
       # Run AMIP GCM (1 hour, no ExtData)
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,5 +55,4 @@ workflows:
           #baselibs_version: *baselibs_version
           #bcs_version: *bcs_version
           gcm_ocean_type: MOM6
-          landbcs_type: NL3
           change_layout: false
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ version: 2.1
 #bcs_version: &bcs_version v12.0.0
 
 orbs:
-  ci: geos-esm/circleci-tools@dev:f0c2c8ef70b5b840bbd8e08f6fabaa9f7bb4d5d4
+  ci: geos-esm/circleci-tools@5
 
 workflows:
   build-test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,11 +1,11 @@
 version: 2.1
 
 # Anchors in case we need to override the defaults from the orb
-#baselibs_version: &baselibs_version v8.0.2
-#bcs_version: &bcs_version v11.5.0
+#baselibs_version: &baselibs_version v8.5.0
+#bcs_version: &bcs_version v12.0.0
 
 orbs:
-  ci: geos-esm/circleci-tools@3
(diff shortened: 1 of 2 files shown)


mathomp4:
Ugh. Coupled having issues with my restarts:
```
 Character Resource Parameter: CATCH_INTERNAL_RESTART_FILE:catch_internal_rst
 Using parallel NetCDF to read file: catch_internal_rst
pe=00002 FAIL at line=03377    NCIO.F90                                 <Could not find field SNOWALB in catch_internal_rst  >
```

I will consult with @biljanaorescanin tomorrow. I'm sure I've done something dumb!

wmputman (approved):

wmputman (approved):

biljanaorescanin:
> Ugh. Coupled having issues with my restarts:
> 
> ```
>  Character Resource Parameter: CATCH_INTERNAL_RESTART_FILE:catch_internal_rst
>  Using parallel NetCDF to read file: catch_internal_rst
> pe=00002 FAIL at line=03377    NCIO.F90                                 <Could not find field SNOWALB in catch_internal_rst  >
> ```
> 

This happens if restart file was not remapped to `v12 `boundary conditions... Your starting restart file should have `SNOWALB `variable.  If you started from `NL3 `you first remap to `v12 `and than you also need to turn in `GEOS_SurfaceGridComp.rc` : `SNOW_ALBEDO_INFO: 1 `.  

This is now issue only due to that limbo of waiting Bill's branch to come to develop. GCM v12 already has this RC setup as default. We just need to remap all current testing restarts to v12 to include this info. I'm not sure when is right time to do that.

mathomp4:
@biljanaorescanin It was my fault. I was forcing NL3 landbcs because for some reason I thought I needed that for MOM6. Nope. So I just needed to undo that and make sure I had the mom6 v12 bits in the image. Works now!

PR #994, opened 2024-09-10
[NDSL] QSat port

CharlesKrop:
Functional port of GEOS_QSat function (and related calls). Function currently fails at majority of points (~86%) with old metric. New metric (not currently working on my machine) should show that these errors are small enough to be considered acceptable. 

Errors can be traces back to differences in saturation pressure tables generates at QSat init. It is believed that these errors pop up because these tables are generated in standard python (gt4py is missing adequate extra dimension capabilities) and therefore executed in 64 bit


FlorianDeconinck (changes requested):
- [x] Merge `dsl/develop` and resolve conflicts
- [ ] Run lint with `pre-commit run --all-files`
- [x] Radiation coupling & Aer Activation translates have been removed from `__init__.py` and file deleted. Revert

FlorianDeconinck (approved):
LGTM - working with netcdf based table

PR #1006, opened 2024-09-20
send clear message to log when mk_lakelandicesaltwater failed

weiyuan-jiang:
When restart file does not match the tile file, the program crashes. But without clear message, it is hard to figure out why it crashes. This PR would send the message to the log file.

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/mk_LakeLandiceSaltRestarts.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/mk_LakeLandiceSaltRestarts.F90
@@ -88,6 +88,9 @@ program mk_LakeLandiceSaltRestarts
 
      call InFmt%open(InRestart,pFIO_READ,rc=rc)
      InCfg = InFmt%read(rc=rc)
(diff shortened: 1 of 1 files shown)


gmao-rreichle (approved):

weiyuan-jiang:
If the tile number of restart file is greater than that of the tile file, the program would run successfully. That is even worse.

gmao-rreichle:
> If the tile number of restart file is greater than that of the tile file, the program would run successfully. That is even worse.

@weiyuan-jiang, I guess by "run successfully" you mean that the program runs but produces garbage?  In any case, are you suggesting that we also make this a hotfix?  Merging it into develop is obviously needed but would not fix the issue in tagged versions of the model. 

cc: @biljanaorescanin @sdrabenh

weiyuan-jiang:
> > If the tile number of restart file is greater than that of the tile file, the program would run successfully. That is even worse.
> 
> @weiyuan-jiang, I guess by "run successfully" you mean that the program runs but produces garbage? In any case, are you suggesting that we also make this a hotfix? Merging it into develop is obviously needed but would not fix the issue in tagged versions of the model.
> 
> cc: @biljanaorescanin @sdrabenh

Yes, that is what I mean. I don't think we need a hot fix. If the users provide right restarts and BCs, everything should be right.

sdrabenh (approved):

PR #1007, opened 2024-09-25
Fix `f32` calculation for QSat tables

FlorianDeconinck:
NetCDF table method has been removed, exact calculation only remains.


katrinafandrich (approved):
This all looks good to me. Ready to merge.

CharlesKrop (approved):

PR #1016, opened 2024-10-14
[NDSL] Refactor constants

FlorianDeconinck:
Aggregate all constants in one file
Put in place a data-driven test for precision verification against reference Fortran


katrinafandrich (approved):
All looks good so far. OK to merge

CharlesKrop (approved):

CharlesKrop:
Are we just going to stick with these names for the foreseeable future or is the plan still to move to a new naming scheme at some point soon?

FlorianDeconinck:
I'd say same as the rest of the codebase: move to something that's not two letters, keep the original name in the comments and/or a file that does spells out the correspondence

PR #1025, opened 2024-11-04
Bug fix for optional input arg "conserve_ice10_tzero" (StieglitzSnow.F90)

gmao-rreichle:
An optional input argument ("conserve_ice10_tzero") was used in an if statement inside subroutine StieglitzSnow_relayer(), but the optional input argument may not be provided by the calling routine. 
In the GCM, the optional argument is provided in both calls to StieglitzSnow_relayer().   In the LDAS (incl. catch_incr.F90), however, the optional argument is not always provided, which makes the "if" statement fail with a seg fault error. 
The fix is to use the local variable ("conserve_ice10_tzero_tmp"), which always exists, in the offending if statement.

Could be considered _trivially_ 0-diff based on inspection of code change.


sdrabenh (approved):

sdrabenh:
@gmao-rreichle @biljanaorescanin sounds reasonable to me. Thanks for the explanation. I approve, but still need Biljana's approval. Assuming no issues with nightlies (which I doubt) I'll tag for LDAS tomorrow.

biljanaorescanin (approved):

PR #909, opened 2024-02-23
Add offline choice for  landice_gridcomp

weiyuan-jiang:
The PR prepares landice_gridcomp to run offline with catch_gridcomp

PR #1028, opened 2024-11-05
add nc4-formatted tile file when generating bcs

weiyuan-jiang:
This PR modifies `make_bcs` to create an NC4-formatted tile file, in addition to the ASCII-formatted tile files. 

The NC4-formatted tile file also includes the supplemental tile attributes (min/max lat/lon and elevation of each tile) needed by GEOSldas, which are currently stored in the (land-only) "catchment.def" file.  The supplemental tile attributes are now also computed for landice and lake tiles.

The PR also includes a py script and an F90 utility program to retroactively create NC4-formatted tile files for existing bcs. 

Because the PR only touches make_bcs, it should be trivially 0-diff for the standard GCM and GEOSldas nightly tests.

The PR is also 0-diff for bcs generation (except for the version-specific README file generated by make_bcs).

In the future, the ASCII-formatted tile file and catchment.def file should become obsolete.

**Contingencies**:   https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/909 must be merged first.

@tclune @biljanaorescanin @sdrabenh @lcandre2


gmao-rreichle:
cc @gmao-rreichle

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rasterize.F90
```diff
@@ -188,22 +186,26 @@ subroutine WriteTilingIR(File, GridName, im, jm, ipx, nx, ny, iTable, rTable, Zi
...
   real(kind=8)   :: fr(size(GridName))
   real(kind=8)   :: xc, yc, area
   real(kind=8)   :: garea, ctg(size(Gridname))
   real(kind=8)   :: sphere, error
-  integer :: status
+  integer        :: status, tmp_in1, tmp_in2, ncat
+  logical        :: file_exists
+
+  character(len=:), allocatable :: filenameNC4, catch_file
+  real(kind=8),     allocatable :: rTableT(:, :)               ! extended and transpose of rTable
```
gmao-rreichle, line 201: Here we're duplicating half of rTable, which isn't memory-efficient.  But since rTable is intent(in), I can't think of an easy way to avoid the extra memory demand without changing subroutine interfaces.  Something to maybe revisit later

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
```diff
@@ -884,30 +911,10 @@ PROGRAM mkEASETilesParam
...
-        mny = clat
-        
-        jgv = real(jg-1) - 0.5
-        
-        call EASE_inverse( EASELabel, real(ig-1), jgv, clat, clon ) 
-
-        mxy = clat 
-        
-        ! write tile properties into catchment.def file
+        ! write tile properties into (ASCII-formatted) catchment.def file
```
gmao-rreichle, line 914: Would it be more efficient to move this write statement outside of this do loop and place it into a separate block after the do loop?  [Same for the write statement where the ASCII-formatted *.til file is written.]  Not sure if switching between computational  stuff and i/o in this way might adversely impacts execution speed

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rasterize.F90
```diff
@@ -305,9 +307,231 @@ subroutine WriteTilingIR(File, GridName, im, jm, ipx, nx, ny, iTable, rTable, Zi
...
+
+  allocate(rTableT(ip, 10))
+
+  rTableT        = MAPL_UNDEF
+  rTableT(:,1:5) = transpose(rTable)
+
+  catch_file = 'clsm/catchment.def'
+  inquire(file=catch_file, exist=file_exists) 
+  if (file_exists) then
+    open (newunit=unit,file=catch_file,action='read', form='formatted',status='old')
```
gmao-rreichle, line 320: It looks like we're getting the min/max lat/lon and elevation info from a catchment.def file (in ASCII format) that was written earlier.  Going via an ASCII file doesn't seem very efficient, and it wouldn't let us eliminate the ASCII-formatted files in the future.  Not sure if I understand the sequence of operations correctly

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
```diff
@@ -726,6 +724,12 @@ PROGRAM mkEASETilesParam
               
               tile_area(tileid_index(i,j)) = tile_area(tileid_index(i,j)) + pix_area     
               
+              ! sum up (area-weighted) elevation 
+                 
+              tile_elev(tileid_index(i,j)) = tile_elev(tileid_index(i,j)) + q0(i,j) * pix_area  ! q0 = elevation on 30-arcsec raster
```
gmao-rreichle, line 729: Moving the tile_elev calculation to here should give tile elevation info also for lake and landice tiles.  This needs verification!

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
```diff
@@ -875,6 +882,26 @@ PROGRAM mkEASETilesParam
...
+     
+     mny = clat
+     
+     jgv = real(jg-1) - 0.5
+     
+     call EASE_inverse( EASELabel, real(ig-1), jgv, clat, clon ) 
+     
+     mxy = clat 
      
      if ((nn>n_land) .and. (nn<=n_landlake)) typ = 19     ! Lake tile
```
gmao-rreichle, line 906: Moving the min/max lat/lon calcs out of the "land"  block should result in min/max lat/lon also for lake & landice tiles.  This needs verification!

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/TileFile_ASCII_to_nc4.F90
```diff
@@ -0,0 +1,301 @@
...
+     
+     ! EASE grid tile space
+
+     i = 1
+     read(tmpline,*)   iTable(i,0), iTable(i,4), rTable(i,1), rTable(i,2),   &
+          iTable(i,2), iTable(i,3), rTable(i,4)
+     do i = 2, N_tile
+        read (unit,*) tmpline
+        read(tmpline,*) iTable(i,0), iTable(i,4), rTable(i,1), rTable(i,2),  &
+             iTable(i,2), iTable(i,3), rTable(i,4)
```
gmao-rreichle, line 118: Can this be simplified as follows?
```
read (unit,*) iTable(i,0), iTable(i,4), rTable(i,1), rTable(i,2),  &
             iTable(i,2), iTable(i,3), rTable(i,4)
```

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/TileFile_ASCII_to_nc4.F90
```diff
@@ -0,0 +1,301 @@
...
+     i = 1
+     read(tmpline,*)   iTable(i,0), rTable(i,3), rTable(i,1), rTable(i,2),   &
+          iTable(i,2), iTable(i,3), rTable(i,4), iTable(i,6),                &
+          iTable(i,4), iTable(i,5), rTable(i,5), iTable(i,7)
+     
+     do i = 2, N_tile
+        read (unit,*)   tmpline
+        read(tmpline,*) iTable(i,0), rTable(i,3), rTable(i,1), rTable(i,2),  &
+             iTable(i,2), iTable(i,3), rTable(i,4), iTable(i,6),             &
+             iTable(i,4), iTable(i,5), rTable(i,5), iTable(i,7)
```
gmao-rreichle, line 140: Can this be simplified as follows?
```
        read (unit,*)  iTable(i,0), rTable(i,3), rTable(i,1), rTable(i,2),  &
             iTable(i,2), iTable(i,3), rTable(i,4), iTable(i,6),             &
             iTable(i,4), iTable(i,5), rTable(i,5), iTable(i,7)
```

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rasterize.F90
weiyuan-jiang: Changing the interface would work. Also the Table should be in transposed form at the very beginning.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
weiyuan-jiang: Probably not much impact

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rasterize.F90
weiyuan-jiang: Good point. I will figure out the sequence of the operation

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/TileFile_ASCII_to_nc4.F90
weiyuan-jiang: good catch

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90
weiyuan-jiang: I believe so

weiyuan-jiang:
@gmao-rreichle It is ready for your review

gmao-rreichle:
@weiyuan-jiang:  To compute the supplemental attributes for landice and lake tiles, I suspect we need to modify the (renamed) subroutine supplemental_tile_attributes().  Specifically, I think the indices ip1 and ip2 appear to constrain the calculations to just land tiles.  Perhaps it would not be difficult to calculate the supplemental attributes for landice and lake tiles.   We do **not** want them for ocean tiles, and we do **not** want them written into the legacy "catchment.def" file.  The idea is to have good values for min/max lat/lon and elevation of landice and lake tiles in the new, nc4-formatted tile file.

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/19eb1cb0ef0c2086ae30e008e56d283a5f71f247/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90#L2138

weiyuan-jiang:
> @weiyuan-jiang: To compute the supplemental attributes for landice and lake tiles, I suspect we need to modify the (renamed) subroutine supplemental_tile_attributes(). Specifically, I think the indices ip1 and ip2 appear to constrain the calculations to just land tiles. Perhaps it would not be difficult to calculate the supplemental attributes for landice and lake tiles. We do **not** want them for ocean tiles, and we do **not** want them written into the legacy "catchment.def" file. The idea is to have good values for min/max lat/lon and elevation of landice and lake tiles in the new, nc4-formatted tile file.
> 
> https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/19eb1cb0ef0c2086ae30e008e56d283a5f71f247/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90#L2138

Is the min/max of lat/lon of landice and lake  calculated by the same way as  land ?

gmao-rreichle:
> Is the min/max of lat/lon of landice and lake calculated by the same way as land ?

I would think so.  The information from the raster (*.rst) file, which is the basis for the min/max lat/lon calculations, isn't specific to land and should work for landice and lake in the same way.

lcandre2:
> > Is the min/max of lat/lon of landice and lake calculated by the same way as land ?
> 
> I would think so. The information from the raster (*.rst) file, which is the basis for the min/max lat/lon calculations, isn't specific to land and should work for landice and lake in the same way.

I agree with Rolf. The lake and landice tile min and max should be calculated the same way as land. Like land, both are irregular.

weiyuan-jiang:
> > > Is the min/max of lat/lon of landice and lake calculated by the same way as land ?
> > 
> > 
> > I would think so. The information from the raster (*.rst) file, which is the basis for the min/max lat/lon calculations, isn't specific to land and should work for landice and lake in the same way.
> 
> I agree with Rolf. The lake and landice tile min and max should be calculated the same way as land. Like land, both are irregular.

OK, I will take care of it

gmao-rreichle:
> The min/max of lat lon are added to the nc4 tile file. But I am not sure if we need to re-adjust the elevation here for the other type. @gmao-rreichle [ tile_ele ](https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/de5c495b04d0a6db63a7a2321d2cf444341e2841/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90#L2158)

Thanks!  I think the mean elevation adjustment is done only for land tiles, and we need to keep it this way for now, otherwise the existing bcs cannot be reproduced identically.  This will require attention, and @biljanaorescanin is already working on this.  Note that the mean elevation adjustment in rmTinyCatchParaMod.F90 is also done in mkEASETilesParam.F90:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/de5c495b04d0a6db63a7a2321d2cf444341e2841/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkEASETilesParam.F90#L800C5-L800C25

I added two commits that should be 0-diff w.r.t. the preceding commit.  The goal was for me to understand better the sequence of operations.  There's a lot of cleanup that could (and probably should) still be done.  I add a couple of quick inline comments, but I don't have time this afternoon to get into a more comprehensive analysis.   

@weiyuan-jiang : For now, please take a look at my commits and let me know if you see anything that's wrong or needs improvement. Thanks

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90
```diff
@@ -2053,6 +1653,14 @@ SUBROUTINE catchment_def (nx,ny,regrid,dateline,gfilet,gfiler)
        call RegridRasterReal(q0,raster)
     endif
 
+    ! -----------------------------------------------------------
+    !
+    ! read ASCII-formatted tile file (*.til)
```
gmao-rreichle, line 1658: Here we're reading the tile info from the ASCII-formatted tile file generated by MAPL.  If this is the first time, then that's unavoidable.   But then we have all the tile info in memory and could pass it into the other subroutines that read the ASCII tile file.  Maybe we can avoid reading the ASCII tile file repeatedly as the bcs are generated.

weiyuan-jiang: Those are intermediate tile files. They are not generated by MAPL but by the program here. We now only produce the last tile file with nc4 format.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90
```diff
@@ -2106,43 +1747,56 @@ SUBROUTINE catchment_def (nx,ny,regrid,dateline,gfilet,gfiler)
...
-    limits(:,1)=360.
+    allocate(limits(1:ip,1:4))
+    limits(:,1)= 360.
     limits(:,2)=-360.
-    limits(:,3)=90.
-    limits(:,4)=-90.
+    limits(:,3)=  90.
+    limits(:,4)= -90.
+
+    ! read raster file with tile IDs
```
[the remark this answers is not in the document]
weiyuan-jiang: it may help save the memory though

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90
gmao-rreichle: @weiyuan-jiang: I'm always a bit confused about the various `*.til` files produced by MAPL and make_bcs in the `./geometry/[RESOLUTION]/` and `./geometry/[RESOLUTION]/til/` directories, but in essence they should all be the same.  

In any case, once we have a "final" `*.ti`l file that defines the geometry, subroutines from make_bcs such as `modis_alb_on_tiles()`, `supplemental_tile_attributes()` [formerly `catchment_def()`], `create_soil_types_files()`, and `compute_mosaic_veg_types()` in rmTinyCatchParaMod.F90 presumably all read the same exact `*.til` file from scratch.  The same file should also be read by `modis_scale_para_high()` in mod_process_hres_data.F90.  

If we can fit the `rTable` and `iTable` variables into memory (which should contain the entire `*.til` file, incl. the "supplemental" fields, except the header info), we shouldn't need to re-read the `*.til` file again from scratch in each subroutine. 

I'm seeing three action items here:

1) Figure out and **document** the differences and purpose of the various flavors of the `*.til` files in the `./geometry/[RESOLUTION]/` and `./geometry/[RESOLUTION]/til/` directories of each set of bcs.  There is some documentation in section 2.2 of the README file that is generated by create_README.csh for each set of bcs in `./land/[RESOLUTION]/clsm`/, but it doesn't say anything about the various flavors of `*.ti`l files.  My best guess is that -- at least for generating the land model parameters of the bcs -- the relevant file is `./geometry/[RESOLUTION]/til/[RESOLUTION]-Pfafstetter.ti`l.

2) Update `create_README.csh` with info about the new nc4-formatted tile file.

3) Convert the tile info contained in `rTable` and `iTable` into a set of variables with interpretable names (or a structure or other object) that can be more readily understood by humans, and keep the info in memory for use by the above-mentioned subroutines.  

The above should work for the resolutions we're most concerned with on a daily basis (c90...c720).  It's unclear to me what the memory implications are for the high-res bcs, say EASEv2 M03 and M01 and the c2880.  Before making any changes, please make sure the logic above doesn't get us into trouble for the high-res bcs.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90
```diff
@@ -2002,31 +1589,44 @@ END SUBROUTINE modis_alb_on_tiles
...
     character*200 :: gtopo30
     character (*) :: gfilet,gfiler
     character*10 :: dline
     CHARACTER*20 :: version,resoln,continent
     REAL, ALLOCATABLE :: limits(:,:)
-    REAL :: mnx,mxx,mny,mxy,dx,dy,d2r,lats,sum1,sum2,dx_gcm,dy_gcm
+    REAL :: mnx,mxx,mny,mxy,dx,dy,d2r,lats,sum1,dx_gcm,dy_gcm
     REAL, dimension (:), allocatable :: tile_ele, tile_area,tile_area_land  
-    integer :: nx,ny,status
+    integer :: nx,ny,IM(2), JM(2)
```
gmao-rreichle, line 1615: This subroutine as well as several others in this F90 file (and presumably other subroutines elsewhere in make_bcs) are still lacking intent() statements.  I suspect this task was part of the Github "bcs cleanup" project, which we never quite completed.  I can't even find the location of the project.  At the very least, we should add intent() statements throughout the subroutines of at least this F90 file.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90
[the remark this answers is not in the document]
gmao-rreichle: I'm not suggesting that we try to keep the entire raster data in memory.  Rather, there should be a **single** do loop:
```
do j=1,j_sib
       read (10)(catid(i),i=1,i_sib)
```
and within the this loop through j we can then do loops through i as needed for (i) the tile elevation calcs and (ii) the limits calcs.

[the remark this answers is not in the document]
gmao-rreichle: addressed in https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1028/commits/b3ce2ca84b8601e938ffac8c6027e48785eb3ea6

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90
gmao-rreichle: Added intent statements, fixed indentation and other cleanup in https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1028/commits/1bc466c35051879e5155148bd3692e7e77bc8fa6

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90
weiyuan-jiang: The new commit will avoid reading the tile file, catchment file and raster file repeatedly @gmao-rreichle

weiyuan-jiang:
It seems to me we should not pull this section out of the loop above. @gmao-rreichle [
](https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/d9d661c529d008707963c0bf79b529f89514f5fa/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rasterize.F90#L435-L438)

gmao-rreichle:
> It seems to me we should not pull this section out of the loop above. @gmao-rreichle [here ](https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/d9d661c529d008707963c0bf79b529f89514f5fa/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rasterize.F90#L435-L438)

@weiyuan-jiang: I still don't understand why this variable should be inside the loop (do ll=1,ng). The loop is for the two grids (ll=1: atm grid, ll=2: ocean grid) that co-determine the global tile space.  The tile-space variable in question is the Pfafstetter index, which shouldn't depend on either grid, and I don't see an explicit dependence on ll for this variable.   What am I missing?

weiyuan-jiang:
> > It seems to me we should not pull this section out of the loop above. @gmao-rreichle [here ](https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/d9d661c529d008707963c0bf79b529f89514f5fa/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rasterize.F90#L435-L438)
> 
> @weiyuan-jiang: I still don't understand why this variable should be inside the loop (do ll=1,ng). The loop is for the two grids (ll=1: atm grid, ll=2: ocean grid) that co-determine the global tile space. The tile-space variable in question is the Pfafstetter index, which shouldn't depend on either grid, and I don't see an explicit dependence on ll for this variable. What am I missing?

@gmao-rreichle , I am not sure either. But the local column number was change to 4 ( I remember it was 3) 
 (https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/86f63ad9a7ce15abc3aa27ae287d2798115f7cec/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rasterize.F90#L252)

I  think the kk is the index. If there are 2 grids, there are two KKs

weiyuan-jiang:
@gmao-rreichle , 
I am very confused by the pfaf number now. Here is how it writes to the text tile file:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/eb3562fb845b2dd0ce3f421b23903342e2d59562/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rasterize.F90#L260-L261

And here is how it reads back from text tile file:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/eb3562fb845b2dd0ce3f421b23903342e2d59562/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90#L1013-L1014

The pfs seems i_index2 to me and it is used as pfs for catchment.def

I guess that number is not important and as long as it is consitent, it should be though.

gmao-rreichle:
> @gmao-rreichle , I am very confused by the pfaf number now. Here is how it writes to the text tile file:
> 
> https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/eb3562fb845b2dd0ce3f421b23903342e2d59562/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rasterize.F90#L260-L261
> 
> And here is how it reads back from text tile file:
> 
> https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/eb3562fb845b2dd0ce3f421b23903342e2d59562/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90#L1013-L1014
> 
> The pfs seems i_index2 to me and it is used as pfs for catchment.def
> 
> I guess that number is not important and as long as it is consitent, it should be though.

@weiyuan-jiang : I can't make sense of this.  As mentioned when we talked earlier, everything looks ok for EASE tiles.  For cube-sphere tiles, I verified (for the existing "v12" c360 bcs) that column 9 of the *.til file does indeed contain the index of the Pfafstetter (hydrological) catchment to which the tile belongs.  In the code, there are tons of variable names for this, incl. "pfs", "pfaf", "kk(1)", "cid", "pfaf1", "pfafindex", "pfaf_index", "catid_index", ...
What beats me is why the existing *.til file do have the Pfafstetter index in column 9, when the write statement you referenced above appears to be putting it into column 8.   Is it possible that the iTable information gets messed up somewhere?  In the do-loop below, kk(1)=iTable(6,K) and ii(2) = iTable(4,K).  I'm having a hard time understanding how iTable gets filled; perhaps that's where the compensating error is? 

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/04e29b34082b2f79d09b6ceaa4cb352e7ab81add/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rasterize.F90#L264

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkCatchParam.F90
```diff
@@ -230,33 +238,55 @@ PROGRAM mkCatchParam
...
        endif
        write (log_file,'(a)')' '
 
-        open (10, file = 'clsm/catchment.def', form = 'formatted', status = 'old', &
-              action =  'read')
-        read (10, *) N_tile
-        close (10, status = 'keep')
+       call ReadTilingNC4( trim(fnameTil)//".nc4", iTable = iTable, rTable = rTable) 
+       N_land = count(iTable(:,0) == 100)                        ! n_land = number of land tiles
+       allocate(tile_pfs,   source = iTable(1:n_land,4))
```
gmao-rreichle, line 274: @weiyuan-jiang : `tile_pfs=iTable(:,4)` seems inconsistent with the documentation in rasterize.F90 (which may or may not be correct):
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/eb3562fb845b2dd0ce3f421b23903342e2d59562/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rasterize.F90#L190

atrayano:
Let me jump in the discussion and add my 2 cents. 
We have ascii tile files that support either 1 or 2 grid. The most typcal case for GEOS is a tile file with 2 grids (although to generate the final file, we generate a lot of intermediate tile files with single grid)
For the 2-grid tile file, the pfaf index number should be in column 9 (basically that file has 3 groups, each of which has 4 columns: the "global" info part (type, area, lat, lon), and then for each grid we have (index1 (i.e. "i"), index2 (i.e. "j"), weight, dummy). Here "dummy" is a variable, used internally for bookkeeping purposes, but it is totally ignored by GEOS, MAPL, etc. So, for the typical case, ATM and OCN grids, columns 1-4 represent the global variables, then the next 4 columns refer to the ATM grid (but this is to a large extend an artifact of the ordering of the "combine" calls that generate the final tile file). Then for type=0 (i.e. "ocean") the last 4 columns are the i, j, weight, dummy of the ocean grid. But for type=100 (i.e. land) the convention is the first index, i.e. column 9, is the pfaf index. I do not think we use the content of column 10 anywhere in the model.
So my bottom line is the pfaf index should be in column 9. If it appears in column 8, it wont do any harm to the atmosphere, but we cannot use it properly to do river routing inside the land model.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkCatchParam.F90
gmao-rreichle: Thanks, @atrayano, for your elucidating comments below on sticking the pfaf index number into column 9 of the cs tile file, which is nominally for the i-index w.r.t. the second (i.e., ocean) grid.  Is there any documentation of this in the code or elsewhere?  I might have missed it.  If not, it is imperative that we add this documentation.  It's extremely time-consuming and nearly impossible to reverse-engineer this kind of repurposing of variables without some external knowledge.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/create_README.csh
```diff
@@ -170,33 +170,33 @@ endif
...
-	    of the ocean grid cell where the ocean tile is located while column 11 contains the fraction\\n \
+	    of the ocean grid cell where the ocean tile is located, while column 11 contains the fraction\\n \
 	    of the ocean grid cell. '`" 
 set rout_smap
 
-if(`echo $gfile | cut -d '_' -f1` == SMAP | $ease == EASE) then
-   set WGRID=SMAP
-   set sec2_til="`printf 'pfaf_code, longitude, latitude, ig, jg, cell_frac, pfaf_index'`"
+if(`echo $gfile | cut -d '_' -f1` == SMAP | $ease == EASE) then   
+   set WGRID=SMAP
```
gmao-rreichle, line 187: @weiyuan-jiang : I think the logic in the "if" statement no longer works when $gfile starts with "EASEv2" (that is, after we dropped the leading "SMAP_" from the grid name).   That is, in recently generated bcs, the README file for the EASE grid bcs shows the wrong read statement.  
E.g., /gpfsm/dnb06/projects/p185/fvInput/ExtData/esm/tiles/v12/land/EASEv2_M36/clsm/README
has the following:
```
       2.2.1 Tile types, location, area and Pfafstetter catchment mapping
         file name: ../til/EASEv2_M36.til
         The 8-line header is followed by  number of rows.
         do n = 1,
                read ([UNIT],*) type, area, longitude, latitude, ig, jg, cell_frac
         end do
```
which is not correct for the EASE grid tile file.

gmao-rreichle:
@atrayano, @weiyuan-jiang, I took a stab at fixing some of the [documentation issues](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1028#discussion_r1920813498) in the following [commit](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1028/commits/e93c98de6bfdfa201acdeaee03c27107fc344102).  Note that the README file in question is only meant to document land model parameters and is not suitable for documenting the tile file.  The tile file documentation needs to cover all tile types (land, ocean, lake, landice).  That is, more documentation is needed.  We also need to add comments into the src code in strategic places so future developers have at least a chance at understanding the repurposing of "i-index for grid 2" as "pfaf_index" for land tiles.

@weiyuan-jiang, I'm a bit confused about your commit that [merged branch 'feature/wjiang/landice_offline' into feature/wjiang/nc4_tilefile](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1028/commits/e71d19ad4588561ee5ff9f299645e79e09bcd47f).  Both branches ("nc4_tilefile" and "landice_offline") are currently out of sync w/ develop.  Can we update these branches to develop?  Do we really want "landice_offline" to be merged into "nc4_tilefile"?  The latter is trivially 0-diff for the GCM and would be easier to get in.  The "landice_offline" branch will take testing, and it might take longer to get it merged.

weiyuan-jiang:
> @atrayano, @weiyuan-jiang, I took a stab at fixing some of the [documentation issues](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1028#discussion_r1920813498) in the following [commit](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1028/commits/e93c98de6bfdfa201acdeaee03c27107fc344102). Note that the README file in question is only meant to document land model parameters and is not suitable for documenting the tile file. The tile file documentation needs to cover all tile types (land, ocean, lake, landice). That is, more documentation is needed. We also need to add comments into the src code in strategic places so future developers have at least a chance at understanding the repurposing of "i-index for grid 2" as "pfaf_index" for land tiles.
> 
> @weiyuan-jiang, I'm a bit confused about your commit that [merged branch 'feature/wjiang/landice_offline' into feature/wjiang/nc4_tilefile](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1028/commits/e71d19ad4588561ee5ff9f299645e79e09bcd47f). Both branches ("nc4_tilefile" and "landice_offline") are currently out of sync w/ develop. Can we update these branches to develop? Do we really want "landice_offline" to be merged into "nc4_tilefile"? The latter is trivially 0-diff for the GCM and would be easier to get in. The "landice_offline" branch will take testing, and it might take longer to get it merged.

To use nc4 tile file in GEOSldas, I have to develop it on the top of add_landice branch in GEOSldas_GridComp, otherwise there would be too much conflicts when merging.  I will update the two branches.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkCatchParam.F90
gmao-rreichle: I added some documentation here: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1028/commits/7614bfa8bf5e1b6e8f8edb43190d03bd22551a11

We're still missing documentation on the various intermediate tile files in the `./geometry/[RESOLUTION]/til/` subdirectory  https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1028#discussion_r1894455505

weiyuan-jiang:
> Let me jump in the discussion and add my 2 cents. We have ascii tile files that support either 1 or 2 grid. The most typcal case for GEOS is a tile file with 2 grids (although to generate the final file, we generate a lot of intermediate tile files with single grid) For the 2-grid tile file, the pfaf index number should be in column 9 (basically that file has 3 groups, each of which has 4 columns: the "global" info part (type, area, lat, lon), and then for each grid we have (index1 (i.e. "i"), index2 (i.e. "j"), weight, dummy). Here "dummy" is a variable, used internally for bookkeeping purposes, but it is totally ignored by GEOS, MAPL, etc. So, for the typical case, ATM and OCN grids, columns 1-4 represent the global variables, then the next 4 columns refer to the ATM grid (but this is to a large extend an artifact of the ordering of the "combine" calls that generate the final tile file). Then for type=0 (i.e. "ocean") the last 4 columns are the i, j, weight, dummy of the ocean grid. But for type=100 (i.e. land) the convention is the first index, i.e. column 9, is the pfaf index. I do not think we use the content of column 10 anywhere in the model. So my bottom line is the pfaf index should be in column 9. If it appears in column 8, it wont do any harm to the atmosphere, but we cannot use it properly to do river routing inside the land model.

Since we are going to add nc4 format, what do you think of the names for those columns? Do you think we should ignore some columns ?  @gmao-rreichle @atrayano

gmao-rreichle:
@weiyuan-jiang : We can discuss in more detail tomorrow.  For now, I'm thinking we keep all the variables and make sure to use no-data-values where appropriate.  For example, for a land tile I would fill ii_1, jj_1, frac_1 with the relevant numbers, fill ii_2, jj_2, frac_2 with non-data values, and store the Pfafstetter index in the appropriate variable.  For the ocean, we fill ii_1, ii_2, jj_1, jj_2, frac_1, frac_2 with the relevant numbers and use no-data for the Pfafstetter index.   If we use nc4-internal compression, the no-data-values won't hurt us in terms of file size.

gmao-rreichle:
Lake and landice are more like land than ocean. We may already have Pfafstetter indices for lakes (from the raster file that is used for land).  We should also have some routing info for landice, but it's a bit heterogeneous.  For Greenland, it's from the landice group.  For continental glaciers, there should be information in the same raster file that is used for land.  Not sure about Antarctica.  For now, it doesn't really matter what we do with lake and landice, so long as the nc4 file reproduces the ASCII tile file for existing bcs.  We'll need a separate PR for fleshing out the routing system across landice and lake

weiyuan-jiang:
> Lake and landice are more like land than ocean. We may already have Pfafstetter indices for lakes (from the raster file that is used for land). We should also have some routing info for landice, but it's a bit heterogeneous. For Greenland, it's from the landice group. For continental glaciers, there should be information in the same raster file that is used for land. Not sure about Antarctica. For now, it doesn't really matter what we do with lake and landice, so long as the nc4 file reproduces the ASCII tile file for existing bcs. We'll need a separate PR for fleshing out the routing system across landice and lake

I will check the program. It is clear now the column 9 has at least two meanings.  When it is land, it represents pfs, when it is ocean , it represent I2.  That is why I am asking what it is for the other types. @gmao-rreichle

gmao-rreichle:
@weiyuan-jiang : Currently, column 9 for lake and landice in cs-based tile files is filled with essentially useless constants: 190000000 for lake and 200000000 for landice.   For existing bcs, we should keep these in the ASCII-based tile file to maintain 0-diff.  In the nc4 tile file, I would fill ii_2 with no-data and pfs with the aforementioned (useless) constants.  Looking further ahead, we need a new PR that properly processes the routing info, as mentioned above.

biljanaorescanin (dismissed):

gmao-rreichle (dismissed):

lcandre2 (approved):

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_shared.py
```diff
@@ -61,15 +62,19 @@ def get_change_til_file(grid_type):
...
 /bin/mv -f tile.file {GRIDNAME}{RS}.til
 /bin/rm -f sedfile
+ncatted -a Grid_Name,global,o,c,'PE{nc}x{nc6}-CF' {GRIDNAME}{RS}.nc4
+ncatted -a Grid_ocn_Name,global,o,c,'PE{imo}x{jmo}-{OCEAN_VERSION}' {GRIDNAME}{RS}.nc4
 endif
 if( {CUBED_SPHERE_OCEAN} == True ) then
 cat > sedfile << EOF
-s/{DATENAME}{IMO}x{POLENAME}{JMO}-Pfafstetter/OC{nc}x{nc6}-CF/g
 s/CF{NC}x6C{SGNAME}/PE{nc}x{nc6}-CF/g
+s/{DATENAME}{IMO}x{POLENAME}{JMO}-Pfafstetter/OC{nc}x{nc6}-CF/g
```
biljanaorescanin, line 71: @weiyuan-jiang For some reason this line is not working on the PR.

mathomp4: @weiyuan-jiang and I looked and I think the "fix" is that line 71 needs to be above line 70 (the way it was before).

What happens line 70 changes the file enough that line 71 doesn't do anything. But when they were reversed in order, line 71 did it's bit and then line 70 worked.

biljanaorescanin:
We fixed all the labels so we are good to be merged.

biljanaorescanin (approved):

sdrabenh (approved):

bena-nasa (approved):

biljanaorescanin (approved):

PR #1034, opened 2024-11-25
[MKIAU] Initial prototype of a CFFI-based fortran/python bridge

FlorianDeconinck:
:warning: This might not be PR'ed back into the right branch :warning:

- Fortran-first bridge that call down to a `pyMKIAU` package that can run Fortran
- Bogus data & code to show usage in `GEOSmkiau_GridComp/GEOS_mkiauGridComp.F90`
- Ships with a Fortran/Python memory converted and a Timer capable of GPU via `cupy`

**Dev note**: this is the current strategy in the SMT to integrate Fortran & Python as it gives us control over overhead. We might evolve the technique as we look more into CPU-CPU bridging. We also aim to deliver a generator so that the boilerplate code is can be code generated and reduce bugs.

Poke @jli-99 
Poke @katherbreen

### DIFF
--- a/GEOSmkiau_GridComp/CMakeLists.txt
+++ b/GEOSmkiau_GridComp/CMakeLists.txt
@@ -1,5 +1,7 @@
 esma_set_this()
 
+option(BUILD_PYMKIAU_INTERFACE "Build pyMKIAU interface" OFF)
+
 set (srcs
   IAU_GridCompMod.F90
   GEOS_mkiauGridComp.F90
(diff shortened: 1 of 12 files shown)


FlorianDeconinck:
The entire code is linked to a `cmake` flag and defaults to not be compiled: no change to default GEOS behavior.

mathomp4:
@FlorianDeconinck Can you make a minor tweak to some file (say the README) and see if that will trigger our CI? I think because this came from a fork, the CI wasn't triggered. I updated a setting and maybe it now will

FlorianDeconinck:
> @FlorianDeconinck Can you make a minor tweak to some file (say the README) and see if that will trigger our CI? I think because this came from a fork, the CI wasn't triggered. I updated a setting and maybe it now will

Looking good!

sdrabenh (approved):

mathomp4 (approved):

PR #1050, opened 2025-01-31
turn off GOSWIM via N_constit=0

gmao-rreichle:
The present PR attempts to turn off GOSWIM by setting the number of constituents to 0. 

GOSWIM is a snow impurity scheme that was implemented in GEOS but has not been used.

Previously, N_constit=9 was hardwired and GOSWIM calculations were executed even though GOSWIM was never used.  To turn off GOSWIM, the GCM's rc parameter AEROSOL_DEPOSITION was set to 0, which forced the constituent mass and the deposition rates to remain zero, but the many GOSWIM do loops through the 9 constituents were still executed, thus multiplying and adding lots zeros many times.

If needed, original behavior can be recovered by setting N_constit=N_constit_GOSWIM=9 in StieglitzSnow.F90

Successfully 0-diff tested for GEOSldas after https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1050/commits/8254a1032a6c45aa30b83c48dd4e73e0b11010e8 (@gmao-rreichle)

If GOSWIM or a revised snow impurity scheme is used in future, N_constit should be configurable (see also the existing rc variables `N_CONST_LAND[ICE]4SNWALB`).


weiyuan-jiang (dismissed):

biljanaorescanin:
Testing summary:
1. @gmao-rreichle did regression tests and they all passed
2. 1-Day AMIP, REPLAY and Inc REPLAY were all zero diff. 
@sdrabenh  this PR is also ready for you

biljanaorescanin (approved):

lcandre2 (approved):

sdrabenh (approved):

biljanaorescanin (approved):

sdrabenh (approved):

mathomp4 (approved):

PR #1052, opened 2025-02-04
v11: Make w3adc more portable

mathomp4:
As detailed by @atrayano, the `-march=core-avx2` flag can cause issues with WW3 building on older nodes (say pfe nodes at NAS which are ooooooooooooooold). So this PR removes the flag from `w3adc` but then re-adds it for the rest of WW3.

I think this should be zero-diff as long as you don't use WW3. If you do use it...not sure.


mathomp4:
Superseded by #1055. Closing

PR #1054, opened 2025-02-07
add make_topo scripts to generate topography files on BC

weiyuan-jiang:
Add make_topo scripts to generate topography files on BC.  

PR needs this topo branch to work for stretched grids: 
https://github.com/GEOS-ESM/Topo/pull/1


biljanaorescanin:
Quick update:
1. All runs regular and stretch run for less then 1h so I reduced time in script. 
2. I've disabled c5760 for now. 
3. We should decide where baseline should be created there is roundoff difference for running on CAS vs MIL. 
What do you think @mathomp4 I could limit script just to mil and rerun ... ?

biljanaorescanin:
@mathomp4 @bena-nasa I've made changes and tested it will add attributes like this at the end, for example here is from c540 : 
`// global attributes:`
		`:STRETCH_FACTOR = 2.5 ;`
		`:TARGET_LAT = 39.5 ;`
		`:TARGET_LON = -98.35 ;`

and regular grid is same as before I've tried c24. I hope this is what you had in mind.

mathomp4:
@biljanaorescanin I think you got it!

biljanaorescanin (dismissed):

sdrabenh (dismissed):

biljanaorescanin:
@sdrabenh @mathomp4 

I've run all resolutions and updated TOPO set on bcs_shared path (v14_BETA/TOPO) with this branch as it is now and TOPO tag 2.1.0.  
Once this is merged and we have a tag we can add a README to the set so it is clear how to reproduce set if needed. 

I think at this point is at good stage to test and see how it behaves in runs. 

All in magnitude compares well to what Replay_Experiments/c360 for example has in scratch/gwd_internal_rst file. 

_Few notes for future regarding **gwd_internal_rst**:_
-------------------------------------------------------------
More  cleanup in sense more metadata inconstancies I didn't catch at first but it is really not going to affect the model it is just truthful representation what nc4 values are as far as description in metadata. 

`1. ANGLL and ANGLX` - sentinels for these two angle variables are wrong 
		`ANGLL:_FillValue = 1.e+15 ;`
		`ANGLL:missing_value = 1.e+36 ;`

 it should be ` -9000`. 

Line for example:
    `64 : 0000-00-00 00:00:00       8    13824       0 :     -9000.0     -8705.5      178.91 : ANGLL`

FYI: GEOS only uses ANGLL and this won't be an issue since from code on develop it has this check in code: 
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/75599dea3070dfc23c9887eaba59db32fe7bbf65/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSgwd_GridComp/GEOS_GwdGridComp.F90#L613-L617

Other ANGLX is never used in GEOS it is internal for topo package. 

Still right way is to fix sentinel to -9000 and possibly normalize so value is right range for angle:  0-180. 

`2.  LANDFRAC` variable -  cosmetic purely. 

GEOS doesn't use this var it is for NCAR plotting package only (ncl script).

Name is here misleading it says in metadata LANDFRAC is fraction of grid cell leading us to belive values should then be 0-1 but we don't see that in gwd_internal_rst file we see: 
    ` 2 : 0000-00-00 00:00:00       0    13824       0 :      0.0000     0.30564      3.0000 : LANDFRAC `
and
	double LANDFRAC(lat, lon) ;
		LANDFRAC:_FillValue = 1.e+15 ;
		LANDFRAC:missing_value = 1.e+36 ;
		LANDFRAC:long_name = "gridbox land fraction" ;
		LANDFRAC:units = "1" ;

so we can fix this in code but we don't use it so I didn't mess with it. 

in this code this line:
https://github.com/GEOS-ESM/Topo/blob/b1f414f0e089af6139404aae689d2caa34886aa1/bin_to_cube/bin_to_cube.F90#L368

We need to use the decoded fraction in the binning, not the raw byte. 

`3. I would remove variable ISOVAR `(or give it real values) has no sense to store in restart something we don't really even populate like place holder. 

     7 : 0000-00-00 00:00:00       0   777600       0 :  1.0000e+36  1.0000e+36  1.0000e+36 : ISOVAR   

	double ISOVAR(lat, lon) ;
		ISOVAR:_FillValue = 1.e+15 ;
		ISOVAR:missing_value = 1.e+36 ;
		ISOVAR:long_name = "SQRT(Variance) from topo NOT represented by ridges" ;
		ISOVAR:units = "m" ;
		ISOVAR:filter = "none" ;

`4.  Pick either area or GBXAR`, not both to be written.  

     6 : 0000-00-00 00:00:00       0   777600       0 :  1.0150e-05  1.6160e-05  2.3383e-05 : area          
     8 : 0000-00-00 00:00:00       0   777600       0 :  1.0150e-05  1.6160e-05  2.3384e-05 : GBXAR 

	double area(lat, lon) ;
		area:_FillValue = 1.e+15 ;
		area:missing_value = 1.e+36 ;
		area:long_name = "angular area of target grid cell" ;
		area:units = "sr" ;

	double GBXAR(lat, lon) ;
		GBXAR:_FillValue = 1.e+15 ;
		GBXAR:missing_value = 1.e+36 ;
		GBXAR:long_name = "angular area of target grid cell from scheme" ;
		GBXAR:units = "sr" ;
		GBXAR:filter = "none" ;

so one option is to make area like m**2? That info feels maybe more useful so we don't repeat same info in restart.   
Area would then be GBXAR* Earth Radius and that will give us Area in square meter.
Or we remove one. 

Again now all of this is mostly cleanup and cosmetic. 

Final taught: 
Other possible major cleanup, there are codes we don't use from package and don't run anymore, we could remove all those if we are making this into stand alone "our" package version.  It would create a significantly smaller code base if we remove all stuff NCAR keeps for their special cases and most of those codes they can't even run from our base if they don't echo our changes there.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/utils_topo/scrip_to_cube_topo.py
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env python
```
mathomp4:
```suggestion
#!/usr/bin/env python3
```

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/utils_topo/scrip_to_restart_topo.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python
```
mathomp4:
```suggestion
#!/usr/bin/env python3
```

sdrabenh (dismissed):

mathomp4 (dismissed):

mathomp4 (approved):

PR #1056, opened 2025-02-12
v12: Fix turb bug

mathomp4:
This fixes a bug found by debug runs of GEOSgcm v12 in CI. Fix via @wmputman and @narnold1


narnold1 (approved):
Looks good

PR #1081, opened 2025-03-30
DO NOT MERGE - Porting GWD to mapl3

pchakraborty:
Illustrating the changes needed for porting GWD to MAPL3


pchakraborty:
> The main thing that catches my eye is the `USE` statements.
> 
> Please open another ticket (maybe already exists) that would target the ability to replace all `USE MAPL*` and `USE ESMF` with a single `USE MAPL3`.

Create #1107

PR #1082, opened 2025-04-01
Optimization of aerosol callback

tclune:
This implementation fuses the loops over modes and thereby
significantly reduces memory copies.   It also potentially reduces
memory storage, as the AeroProps data structure is not needed. (At
least here.)

@wmputman 
@dbarahon


tclune:
Tagging @aoloso who will likely be called upon to test these changes.

dbarahon:
@tclune looks good. The deleted code however is needed for the 2 moment microphysics. Although if we are doing something like this, it is probably better to call the callback from MGB making a copy there, and forget about the structure all together. Old code like RAS and MG1 would need to be updated.

PR #1103, opened 2025-05-09
[NDSL] Update GEOS integration to GFDL_1M

FlorianDeconinck:
- Update to `MicrophysicsConfiguration`
- Automatic bridge between Fortran flag and Configuration


CharlesKrop (approved):
I see no reason why this would not work. All of these changes are things I wanted to do at some point anyway - although admittedly I don't think I would have come up with the `upper_case_dict` thing. That level of cleverness is still a bit beyond me, but I will remember that trick. I didn't go through each of the names in the config files to make sure they line up properly - but I don't recall changing them, and I'm assuming that would have broken at least some of my GFDL_1M driver tests so all should be fine.

FlorianDeconinck:
> I see no reason why this would not work. All of these changes are things I wanted to do at some point anyway - although admittedly I don't think I would have come up with the `upper_case_dict` thing. That level of cleverness is still a bit beyond me, but I will remember that trick. I didn't go through each of the names in the config files to make sure they line up properly - but I don't recall changing them, and I'm assuming that would have broken at least some of my GFDL_1M driver tests so all should be fine.

The upper casing things is fragile as hell though, I might put a warning in there to feedback that some configuration might not have been set downstream from the upstream

PR #1120, opened 2025-06-27
Turbulence and SCM updates on v12rc16

narnold1:
Fixes TKE transport diagnostic, adds ice water source term to EDMF, adds EDMF tracer transport, adds SCM_AREA parameter for single column model, adjustments in SHOC, and code cleanup.


mathomp4:
I'm pushing up a fix for the CI. This is weird.

PR #1136, opened 2025-07-11
improve runoff distribution

zhaobin74:
This PR complements https://github.com/GEOS-ESM/GEOS_OceanGridComp/pull/105 to improve the runoff distribution. 

@mfmehari, I think OBIO needs to be addressed separately as it also has a DISCHARGE import. I can't do it since I am not running the OBIO model. 

cc' @amolod


mfmehari:
@zhaobin74  Ah with "OGCM GC 's grid is not tripolar" it won't work. I will have it imported from ExtData from OBIO GC.

zhaobin74:
> @zhaobin74 Ah with "OGCM GC 's grid is not tripolar" it won't work. I will have it imported from ExtData from OBIO GC.

Thanks @mfmehari. Please submit another PR for OBIO GC.

sdrabenh (approved):

mathomp4 (approved):

PR #1149, opened 2025-08-28
Boundary conditions - v13 based on MOM6/v2 ocean bathymetry (OM4) 

sinakhani:
- This PR adds updates for files in makebcs directory to include MOM6/v2 ocean bathymetry. 
- This PR also adds the wiring for handling (atmos) TOPO versions (but v13 bcs still have v1 TOPO). 
- This is non-zero diff for MOM6 in coupled mode, but zero-diff otherwise (AMIP). 

Successfully 0-diff tested by @biljanaorescanin on 10-Apr-2026 https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1149#issuecomment-4223649739

Related PRs: 
https://github.com/GEOS-ESM/GEOS_Util/pull/201
https://github.com/GEOS-ESM/Topo/pull/5


gmao-rreichle (commented):
see comments below

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90
```diff
@@ -5514,10 +5529,10 @@ SUBROUTINE svbksb(u,w,v,m,n,b,x)
     REAL*8 b(m),u(m,n),v(n,n),w(n),x(n) 
     PARAMETER (NMAX=500)  !Maximum anticipated value of n
     !------------------------------------------------------------------------------------------- 
-    ! Solves A · X = B for a vector X, where A is specified by the arrays u, w, v as returned by 
+    ! Solves A � X = B for a vector X, where A is specified by the arrays u, w, v as returned by 
     ! svdcmp. m and n are the dimensions of a, and will be equal for square matrices. b(1:m) is 
     ! the input right-hand side. x(1:n) is the output solution vector. No input quantities are 
-    ! destroyed, so the routine may be called sequentially with different b’s. 
+    ! destroyed, so the routine may be called sequentially with different b?s.
```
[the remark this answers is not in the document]
mathomp4: If I had to guess, an editor inserted a "smart quote". Or maybe the *OS* inserted it? macOS is "fun" that way at times.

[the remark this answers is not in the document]
sinakhani: I think this is an OS issue. I tried to somehow change the text to avoid this (sorry, if it is old-fashioned!)

gmao-rreichle (commented):
Thanks, @sinakhani.  See more comments below

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_questionary.py
```diff
@@ -174,6 +174,14 @@ def ask_questions(default_grid="Cubed-Sphere"):
...
             "choices": ["No", "Yes"],
         },
+   
+        {"type": "select",
+         "name": "ocean_bathy_version",
+         "message": "Select ocean bathymetry version in coupled mode.",
+         "choices": [ \
+    "v1  : MOM5 bathymetries and MOM6 bathymetries with v1 ocean grids",\
+    "v2  : MOM6 bathymetries with OM4  ocean grids. For now, only  available at 1440x1080. Select v14 for next question"],  
+        },
```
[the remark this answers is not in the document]
sinakhani: I removed this. Thanks for the clarification.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_questionary.py
```diff
@@ -199,6 +207,7 @@ def ask_questions(default_grid="Cubed-Sphere"):
    "v11 : NL3 + JPL veg height + PEATMAP + MODIS snow alb v2", \
    "v12 : NL3 + JPL veg height + PEATMAP + MODIS snow alb v2 + Argentina peatland fix", \
    "v13 : NL3 + JPL veg height + PEATMAP + MODIS snow alb v2 + Argentina peatland fix + mean land elevation fix", \
+   "v14 : NL3 + JPL veg height + PEATMAP + MODIS snow alb v2 + Argentina peatland fix + mean land elevation fix + v2 (OM4) ocean-seaice bathymetry", \
```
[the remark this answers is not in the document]
sinakhani: The suggested change has been made.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90
```diff
@@ -242,6 +242,21 @@ SUBROUTINE init_bcs_config(LBCSV)
...
+
+       LAIBCS  = 'MODGEO'
+       SOILBCS = 'HWSD_b'
+       MODALB  = 'MODIS2'
+       SNOWALB = 'MODC061v2'
+       OUTLETV = "v2"
+       GNU     = 1.0
+       use_PEATMAP = .true.
+       jpl_height  = .true.
+
```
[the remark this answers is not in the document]
gmao-rreichle: I merged the "case" statements for clarity.  @biljanaorescanin: The comments still need to be merged and edited

sinakhani:
I just cleaned up the text for merged cases "v12", "v13", and "v14". Please feel free to make further changes. Thank you.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90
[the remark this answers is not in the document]
gmao-rreichle: > I just cleaned up the text for merged cases "v12", "v13", and "v14". Please feel free to make further changes. Thank you.

Thanks, Sina.  @biljanaorescanin: What I'm looking for here is a more detailed set of comments that specify what exactly is in each version for each of the bullet points (and possibly additional bullet points with differences that aren't mentioned yet -- I forget if there are more differences).  E.g., the catchment.def topo fix is not in v12 but is in v13 and v14.

biljanaorescanin:
This PR was tested for BCS creation for both old `v12` set and new `v14`  . 
Zero diff to `v12` with addition to `TOPO` linking.  New version `v14` is sourcing right files. 
BCS run's for resolutions we do not have new bathymetry file will use `v1` so we can create full set of BCS.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_cube.py
```diff
@@ -5,15 +5,58 @@
 import os
 from make_bcs_questionary import *
 from make_bcs_shared import * 
+import re
+from datetime import datetime
+import subprocess
+
+def _resolve_versions(bcs_version: str):
+    """
+    Returns (mom6_bathy_version, topo_version) as 'v1' or 'v2'
```
gmao-rreichle: The fact that bcs "v14" happen to use "v2" for both TOPO **and** MOM6_BATHY_VERSION is pure coincidence.  If in the next bcs version we update TOPO but not MOM6_BATHY_VERSION (or vice versa), this code won't work anymore.   It may be clearer to use a look-up table approach as in https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/a5871db4d1ead498ac72eb26e604475f567bc1d7/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90#L117.  I assume python has something like the "select case" construct.  Do we have anything other sub-versions that are controlled in the python portion of the code (besides TOPO and MOM6_BATHY_VERSION)?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_cube.py
```diff
@@ -5,15 +5,58 @@
...
+    for size in sizes:
+        src_pref = os.path.join(inputdir, "ocean", "MOM6", preferred_version, size)
+        src_other = os.path.join(inputdir, "ocean", "MOM6", other,            size)
+
+        if os.path.isdir(src_pref):
+            src = src_pref
+            note = ""
+        elif os.path.isdir(src_other):
+            src = src_other
+            note = f'echo "NOTE: MOM6 {preferred_version}/{size} not found; using {other}/{size}"'
```
gmao-rreichle: Are you linking v1 MOM6_BATHY_VERSION into v14 because we don't have v2 data?  That will create all sorts of confusion.  Why can't we generate v2  bathymetry for all resolutions?  Rather than linking the "wrong" version, I think we should omit the link if the v2 bathymetry isn't available for a given resolution.

gmao-rreichle (commented):
see comments below

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_cube.py
biljanaorescanin: That would mean we can't run GCM v12 model version at c90 or c12 resolutions. Not even for nightly testing we have to do for GCM. Since I won't be able to create bcs files for them.  So that is not an option to my understanding. Why there is no version 2 of files for those resolutions maybe @sinakhani can chime in. In general, as soon as he ads them code will always grab v2.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_cube.py
biljanaorescanin: Good point, I missed that one...  I will decouple them with a small, explicit lookup table.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_cube.py
gmao-rreichle: For now, I assume nightly tests don't run with v14 bcs.   The AGCM can run just fine at all resolutions with v14 bcs, independent of the presence of bathymetry files.  If v14 bcs aren't ready for the coupled model, then they're not ready.  It is not ok to mix and match bathymetry versions within a single bcs version.

biljanaorescanin: @mathomp4 can you chime in and and let us know is there plan to run coupled runs (testing like nightly or any runs at all really) with new bcs (**v14**) `GCMv12 will use` at c12 or c90 res?  @gmao-rreichle proposed edit will create for v14 bcs only one resolution for coupled runs and that is `CF0180x6C_M6TP1440x1080` if that is what group wants I can edit code.  

To me easiest resolution so everyone is happy is for @sinakhani to create other resolutions (c90 and c12) . But I don't know how fast he can do that or can he do that.

sinakhani: The new ocean bathymetry (MOM6/v2) is at 1/4 deg for now (this is an adoption from GFDL-OM4_025). My plan is to have it at 1 deg as well (this has been in my to-do list for some time!). And probably at 1/8 deg at some point. I am not sure if we need c12 with the new bathymetry (it is rather too coarse for ocean test exp).

gmao-rreichle: @sinakhani : Keep in mind that the c12 tests are **not** for **science** testing, they're for 0-diff testing.   I'll defer to @mathomp4 about what makes sense here, but I would strongly advise against mixing bathymetry versions within a single bcs version.  If the easiest way to create "v2" bathymetry for c12 is to copy the "v1" file, so be it.  But for any resolution that may be useful for science, it's a hard no from my end.

biljanaorescanin: I've adopted suggested change in last commit (https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1149/commits/c16644cc356a765a28f42f7c03c98d44fea9ad0a), now code will abort if that res is not present. Once Sina adds it, code will work with no additional changes. 
`Grid_type: Cubed-Sphere`
`orslv: T3MOM6`
`resolution: c90`
`ABORT: Missing MOM6 bathymetry: /discover/nobackup/projects/gmao/bcs_shared/make_bcs_inputs/ocean/MOM6/v2/540x458 (LBCSV=v14 MOM6=v2)`

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_cube.py
biljanaorescanin: Done in this commit: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1149/commits/c16644cc356a765a28f42f7c03c98d44fea9ad0a

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_ease.py
```diff
@@ -15,6 +15,10 @@
 chmod 755 bin/create_README.csh
 bin/create_README.csh
 
+# Prevent shared CF TOPO linking for non-cube grids:
+if ( ! -d TOPO ) mkdir -p TOPO
+if ( ! -e TOPO/CF{NC}x6C{SGNAME} ) mkdir -p TOPO/CF{NC}x6C{SGNAME}
+
```
gmao-rreichle: @biljanaorescanin : Is this shell scripting within a python file?  Also, are we adding TOPO files for EASE bcs?  There's no need to add (atm) TOPO files to EASE bcs, which are used only for offline work.

biljanaorescanin: It is dummy so code works. We don't get topo for lat-lon and ease. 
Here is how once it is done it looks like (I'm still running them for some resolutions): 
[BCS_PATH]//fvInput/ExtData/esm/tiles/v13/

gmao-rreichle: > It is dummy so code works.

What does this mean?  If the code does nothing, why is it there? 

> Here is how once it is done it looks like (I'm still running them for some resolutions):
[BCS_PATH]//fvInput/ExtData/esm/tiles/v13/

The top-level v13 dir now shows lots of new resolution-specific dirs.  This doesn't make sense.  If these new dirs are geometry-related, they should be under ./geometry

biljanaorescanin: **Second part (why you currently see resolution directories under v13/)**
I’m currently running the package and the runs are not finished yet,  I mentioned this in my GitHub message.
While jobs are running, temporary resolution specific directories appear under the v13/ path. Once the runs complete, the job’s final move step sorts outputs into the correct geometry/, land/, and logs/ subdirectories and removes the temporary run directory.

I don’t have enough disk space to keep large resolution outputs under my user at the moment. What you’re seeing right now is intermediate state, not the final directory layout.

**First part (why the dummy TOPO/CF0000x6C directory exists)**

The dummy `TOPO/CF0000x6C` directory is intentional and is a protection for non cubed sphere grids. The shared job tail (get_script_mv() in make_bcs_shared.py) was updated to always execute a CF TOPO link block at the end of every job.
It requires placeholders such as {TOPO_VERSION}, {NC}, and {SGNAME} during Python string formatting. It attempts to link a CF topo directory named CF{NC}x6C{SGNAME} into TOPO/.

For CF runs this behavior is correct. For EASE and Lat-Lon runs, it is not needed (and the corresponding CF… topo directory does not exist), but the shared script still executes the block.

The dummy directory serves two purposes:
1)  It prevents job generation failures: without supplying dummy {NC}, {SGNAME}, and {TOPO_VERSION} values, Python .format() would raise a KeyError when generating the job script.

2) It neutralizes the shared TOPO link step at runtime: by pre creating TOPO/CF0000x6C in the scratch run directory, the shared block detects that the target already exists and becomes a no op, no symlink is created, no “missing topo source” warning is printed, and no outputs are affected.

Although the dummy directory `does nothing` from a science perspective, it keeps the pipeline stable without modifying shared logic and avoids spurious warnings or incorrect TOPO links for grid types that do not use topography. The dummy directory exists only in the temporary run directory and is removed when the run completes. That is why I pointed you to final grids dir to see they are clean. 

`But if you don't like this the alternative is to make the TOPO link block cube only in get_script_mv() (or use separate mv tails per grid), so EASE/Lat-Lon never include cube only placeholders and we don’t need dummy directories... `  Downside is reprocessing but that is ok I wasn't done anyway I can stop jobs.

gmao-rreichle: @biljanaorescanin:

Re. "second part" (v13 bcs dir):

> What you’re seeing right now is intermediate state, not the final directory layout.

I'm surprised you're creating the bcs directly in the "final/official" directory.  I thought we wouldn't put bcs into this dir until they're fully vetted.  Rather, we would (i) create the bcs in some temp dir (which could be on the "shared_bcs" disk if you don't have space in your personal nobackup), (ii) verify the bcs, and (iii) mv the bcs into the final/official directory.

Re. "first part" (dummy code):
In make_bcs_shared.py, why do we execute code for all grid types when it only applies to some grid types?  Wouldn't it make more sense to execute (atm) TOPO stuff only for cube-sphere grids?  It looks like make_bcs_shared.py knows the grid type.  Maybe I'm missing something.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_shared.py
```diff
@@ -172,6 +216,29 @@ def get_script_mv(grid_type):
...
 endif
 
+
+# Link TOPO into this BCS directory based on bcs_version
+# You can override with: setenv TOPO_VERSION v1|v2  before launching the job
+if ( $?TOPO_VERSION ) then
+    set topo_version = $TOPO_VERSION
+else
+    set topo_version = {TOPO_VERSION}
+endif
```
gmao-rreichle: Why do we allow the override with setenv?  Don't we have a yaml file that controls the config?  Adding an env variable that overrides yaml inputs seems to violate the original config design.

biljanaorescanin:
@gmao-rreichle I think I implemented all your comments but please let me know if I missed something. 
I am retesting it now.

biljanaorescanin:
I've re-tested few grids and res just to confirm it is zero diff to before and what we want as far as behavior. And all looks good.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_cube.py
```diff
@@ -175,6 +226,8 @@ def make_bcs_cube(config):
            STRETCH = STRETCH, \
            SGNAME  = SGNAME, \
            SGPARAM = SGPARAM, \
+           TOPO_VERSION = TOPO_VERSION,
+           mom6_bathy_version = MOM6_BATHY_VERSION,
```
gmao-rreichle: Don't we need to end these lines with "\\"?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_shared.py
```diff
@@ -4,6 +4,50 @@
...
+def _normalize_lbcsv(label: str) -> str:
+    v = (label or "").strip()
+    m = re.match(r'[vV]?\d+|[A-Za-z0-9]+', v)  # accepts v13 / 13 / NL3 / etc.
+    if not m:
+        return v
+    key = m.group(0)
+    # normalize numeric like '13' -> 'v13'
+    if key.isdigit():
+        key = 'v' + key
+    return key
```
gmao-rreichle: Why do we need this "normalize" function?  Don't we always have a 3-char alphanumeric string for "bcs_version"?   When would we have "bcs_version=13" (without the leading "v")?

gmao-rreichle (commented):
@biljanaorescanin :  I just pushed a commit with a few minor updates to the comments and white space.  One of the updates might affect a log file, so technically it may not be zero-diff w.r.t. the v13 bcs that you already generated.  But there's no need to re-generate the v13 bcs just because of the minimal log file change. 
See also in-line comments below.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_shared.py
biljanaorescanin: I've removed normalization with last commit.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_questionary.py
```diff
@@ -198,7 +198,7 @@ def ask_questions(default_grid="Cubed-Sphere"):
    "v10 : NL3 + PEATMAP + MODIS snow alb v2", \
    "v11 : NL3 + JPL veg height + PEATMAP + MODIS snow alb v2", \
    "v12 : NL3 + JPL veg height + PEATMAP + MODIS snow alb v2 + Argentina peatland fix", \
-   "v13 : NL3 + JPL veg height + PEATMAP + MODIS snow alb v2 + Argentina peatland fix + mean land elevation fix", \
+   "v13 : As in v13 + mean land elevation fix + MOM6 v2 (OM4) ocean-seaice bathymetry", \
```
gmao-rreichle: @biljanaorescanin : I think this should read:
"v13 : As in **v12** + mean land elevation fix + MOM6 v2 (OM4) ocean-seaice bathymetry"

biljanaorescanin:
BCs package was tested and all grids and resolutions are zero diff to what we have already produced on  `bcs_shared/*/v13` . Matching PR on GEOS_Util is also marked as ready. 

@sinakhani let me know if on final path you are missing any coupled resolution combination, I still didn't hear from you for that?  (Or I should remove if I have some we don't plan to use. )

biljanaorescanin (approved):

gmao-rreichle (approved):

sdrabenh (approved):

sdrabenh (approved):

PR #1152, opened 2025-09-04
CatchCN CLM51 fix endrun exit issue

biljanaorescanin:
`NOTE: We can't merge this until Nitrogen balance error is resolved. `

**Issue :** 
During CatchCN run we see error message but run doesn't exit it proceeds and finishes successfully. 

In code we have calls to endrun in right locations but down the stream final code just did assert print /log not exit.

**With Branch fix:** 
**Code stops** and prints error and trace to error, example from my 1day run:
**`In GEOSldas_log_txt trace:`**
 ENDRUN:
 ERROR: 
 ERROR in CNBalanceCheckMod.F90 at line                                      548
 
===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   RANK 0 PID 27588 RUNNING AT borgk145
=   KILLED BY SIGNAL: 9 (Killed)

**`In GEOSldas_err_txt trace:`**
No errorforrtl: severe (174): SIGSEGV, segmentation fault occurred
libpthread-2.31.s  00001495FB25A910  Unknown               Unknown  Unknown
GEOSldas.x         0000000000445658  Unknown               Unknown  Unknown
libpthread-2.31.s  00001495FB25A910  Unknown               Unknown  Unknown
No errorshr_abort_abort: hard abort
shr_abort_abort: hard abort
shr_abort_abort: hard abort

@gmao-rreichle  @weiyuan-jiang

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM51_GridComp/CLM51/CNBalanceCheckMod.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM51_GridComp/CLM51/CNBalanceCheckMod.F90
@@ -5,6 +5,9 @@ module CNBalanceCheckMod
   ! Module for carbon/nitrogen mass balance checking.
   !
   ! !USES:
+
+  use ESMF
+  
   use shr_kind_mod                    , only : r8 => shr_kind_r8
   use nanMod                          , only : nan
   use shr_log_mod                     , only : errMsg => shr_log_errMsg
@@ -220,6 +223,7 @@ contains
     real(r8) :: som_c_leached_grc(bounds%begg:bounds%endg)
     real(r8) :: hrv_xsmrpool_amount_left_to_dribble(bounds%begg:bounds%endg)
     real(r8) :: dwt_conv_cflux_amount_left_to_dribble(bounds%begg:bounds%endg)
+    integer  :: rc       
     !-----------------------------------------------------------------------
 
     associate(                                                                            & 
@@ -307,7 +311,7 @@ contains
          write(iulog,*)'wood_harvestc            = ',wood_harvestc(c)*dt
          write(iulog,*)'grainc_to_cropprodc      = ',grainc_to_cropprodc(c)*dt
          write(iulog,*)'-1*som_c_leached         = ',som_c_leached(c)*dt
-         call endrun(msg=errMsg(sourcefile, __LINE__))
+         call endrun(msg=errMsg(sourcefile, __LINE__),rc)
       end if
 
       ! Repeat error check at the gridcell level
@@ -379,11 +383,13 @@ contains
          write(iulog,*)'dwt_seedc_to_deadstem_grc =', dwt_seedc_to_deadstem_grc(g) * dt
          write(iulog,*)'--- Outputs ---'
          write(iulog,*)'-1*som_c_leached_grc    = ', som_c_leached_grc(g) * dt
-         call endrun(msg=errMsg(sourcefile, __LINE__))
+         call endrun(msg=errMsg(sourcefile, __LINE__),rc)
       end if
 
     end associate
 
+    RETURN_(ESMF_SUCCESS)
+    
   end subroutine CBalanceCheck
 
   !-----------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


weiyuan-jiang:
I am not sure if it is a right fix. The optional argument " ec" and "rc" are never used in the actual calls.  Does the "stop" really stop the program ? Should we remove the ASSERT and bring back the original abort?

gmao-rreichle:
> I am not sure if it is a right fix. The optional argument " ec" and "rc" are never used in the actual calls. Does the "stop" really stop the program ? Should we remove the ASSERT and bring back the original abort?

@weiyuan-jiang : We talked with Tom, and I just pushed what I hope is a cleaner fix.  If it builds, @biljanaorescanin can give it a try.  In the meantime, don't hesitate to let me know if you see anything wrong.  I always feel a bit lost when it comes to ESMF and MAPL

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM51_GridComp/CLM51/CNBalanceCheckMod.F90
```diff
@@ -307,7 +311,7 @@ subroutine CBalanceCheck(this, bounds, num_soilc, filter_soilc, &
          write(iulog,*)'wood_harvestc            = ',wood_harvestc(c)*dt
          write(iulog,*)'grainc_to_cropprodc      = ',grainc_to_cropprodc(c)*dt
          write(iulog,*)'-1*som_c_leached         = ',som_c_leached(c)*dt
-         call endrun(msg=errMsg(sourcefile, __LINE__))
+         call endrun(msg=errMsg(sourcefile, __LINE__),rc)
```
gmao-rreichle: Come to think of it, why not just use _ASSERT(.false.) here (and below)??  In any case, we probably need to take a look at error handling throughout the code in ./CLM51

weiyuan-jiang:
I think the change is not enough.  1) When this abort is called, the program should stop running so it should not return SUCCESS.  2) We need to bubble up all the rc s in every call to get back the trace if we use MAPL's error handling. It is not hard, but there would be a lot of changes

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM51_GridComp/CLM51/CNBalanceCheckMod.F90
gmao-rreichle: My initial commit doesn't build because I didn't specify "Iam" in the CTSM abort routines.  It doesn't make much sense to do that.  I'll fix it in a bit along the lines of doing _ASSERT instead of endrun() in the science code

gmao-rreichle:
@weiyuan-jiang, @biljanaorescanin : I reverted "abort" to that the original CTSM code (6e02675b47f6236964a200d8d1b04507ca10b397).  This builds, and it might be worth seeing if it works.
However, I'm really confused about `shr_sys_mod.F90`:
1) Jana's version of this file comments out almost everything.  For example, subroutine shr_sys_system() (to make. a system call) has been commented out.  But this subroutine is called in `shr_file_mod.F90`.  I don't see why this builds.  I must be missing something.  (This issue has nothing to do with the "abort" problem addressed in the PR.)
2) Per Jana's [README_CN51.md](https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/feature/jkolassa_cnclm51/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM51_GridComp/README_CN51.md), the original files are from CTSM tag [branch_tags/PPE.n08_ctsm5.1.dev023](https://github.com/ESCOMP/CTSM/releases/tag/branch_tags/PPE.n08_ctsm5.1.dev023).  But the 'shr_sys_mod.F90' file [from this tag](https://github.com/ESCOMP/CTSM/blob/branch_tags/PPE.n08_ctsm5.1.dev023/tools/mksurfdata_map/src/shr_sys_mod.F90) is different from the [file in ./CLM51_orig_files](https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/feature/jkolassa_cnclm51/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatchCN_GridComp/GEOScatchCNCLM51_GridComp/CLM51_orig_files/shr_sys_mod.F90). 

Ideally, one of the two "original" files could be used instead of the modified file.  We should avoid changes from the original CTSM code if we can help it.  (It doesn't help that it's not clear what the original code is...)

biljanaorescanin:
I ran a sanity test to confirm that we exit when an error occurs. 
Once we’re ready, this will be good to go.

mathomp4:
Note: The Spack CI will (eventually) work once `develop` is merged into this branch (or it's parent branch)

PR #1159, opened 2025-09-17
update a couple of CICE6 forcing nml variables

zhaobin74:
This PR updated a few forcing nml parameters to be compatible with mushy layer scheme now as default.


mathomp4:
Ack. I need to fix the label enforcer to know about AMIP vs coupled. For now, I'll also label with 0-diff.

zhaobin74:
Thanks, @mathomp4.

sdrabenh (approved):

mathomp4 (approved):

PR #1160, opened 2025-09-17
v12: update a couple of CICE6 forcing nml variables 

zhaobin74:
This is the same as https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1159 but for v12.


mathomp4:
Ack. I need to fix the label enforcer to know about AMIP vs coupled. For now, I'll also label with 0-diff.

PR #1184, opened 2025-12-19
Updating to new MAPL3

tclune:
Just ensuring that the GCM can build with latest MAPL.   Lots of work needed for running, of course.

### DIFF
--- a/GEOS_GcmGridComp.F90
+++ b/GEOS_GcmGridComp.F90
@@ -26,8 +26,6 @@ module GEOS_GcmGridCompMod
 
    use GEOS_OgcmGridCompMod,     only:  OGCM_SetServices => SetServices
    use GEOS_WgcmGridCompMod,     only:  WGCM_SetServices => SetServices
-   use MAPL_HistoryGridCompMod,  only:  Hist_SetServices => SetServices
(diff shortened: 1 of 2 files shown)


GEOS_GcmGridComp.F90
```diff
@@ -150,8 +147,8 @@ subroutine SetServices ( GC, RC )
 
     type (T_GCM_STATE), pointer         :: gcm_internal_state => null()
     type (GCM_wrap)                     :: wrap
-    type (T_EXTDATA_STATE), pointer     :: extdata_internal_state
-    type (EXTDATA_wrap)                 :: ExtDataWrap
+!#    type (T_EXTDATA_STATE), pointer     :: extdata_internal_state
```
pchakraborty, line 150: Should be remove the commented out code?

tclune: If you can think of a way to ensure that we know what is missing when we ultimately port this to work with MAPL3, I'm fine with deleting.    Otherwise, I think we need the proverbial bread crumbs here.

pchakraborty: Let’s leave it in for now.

pchakraborty (approved):

PR #1193, opened 2026-01-07
dsl: Try out per-branch codeowners

mathomp4:
Hmm. I'm trying this out to see if we can now do per-branch CODEOWNERS...


mathomp4:
Hmm. Okay. So I think it needs to be *in* `dsl/develop`. We learn...

PR #1196, opened 2026-01-07
BCS processing to support HWSDv2 soil preprocessing and new Peat GPA22

biljanaorescanin:
This PR has 2 parts:

1) Add HWSDv2 soil preprocessing workflow (STEP1–STEP5), including supporting QA utilities and documentation. Details on the HWSDv2 workflow, inputs, and processing sequence are documented in a README file.  This workflow is included in the PR for continued development and testing, but it is not intended to be enabled for new  BCS releases at this time, since current testing shows it produces a substantially different soil moisture climatology relative to that associated with existing bcs while soil moisture simulation skill is neutral.   

2) Update raster BCS soil processing for peatlands based on Global Peatland Map (GPM) 2.0 and supporting MODIS MCD12Q1 crop land cover information. Also fixes problem with inconsistent soil parameters in old processing logic for peatlands as the union/merger of PEATMAP and HWSD information.  Add required preprocessing utilities under `preproc/soil/`.  A new "v14" release of bcs will include the GPM 2.0-based processing of peatlands. 

PR is 0-diff trivial for GCM and LDAS simulations because it only touches "Raster" utilities.

PR is 0-diff for make_bcs of exiting bcs versions. 

Related PRs:
https://github.com/GEOS-ESM/GEOS_Util/pull/156


biljanaorescanin:
@gmao-rreichle I've tested with last commit BCS for v12 EASE M36 vs what is on develop now is zero diff and new ones are doing what they suppose to. So backward compatibility is preserved.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90
```diff
@@ -3761,28 +3771,69 @@ SUBROUTINE soil_para_hwsd (nx,ny, n_land, tile_pfs, tile_id)
...
+               NF_NOWRITE, ncid) ; VERIFY_(STATUS)
+       endif
+       
+       status = NF_GET_VARA_REAL(ncid, NC_VarID(ncid,'PEATMAP'), (/1,1/), (/i_highd,j_highd/), pmapr) ; VERIFY_(STATUS)
+       
+       ! Strict GPA22
+       ! prevent HWSD oc_top from generating peat outside the GPA22 peat mask.
+       if (PEATMAP_STRICT_GPA22) then
+          where (pmapr < PEATMAP_THRESHOLD_1)
+             oc_top = min(oc_top, NINT((cF_lim(4) - 1.0e-5)/sf))  ! cap oc_top just below cF_lim(4)=~8.72
```
gmao-rreichle: Since cF_Lim(4)~8.72, subtracting 1.0e-6 probably cannot be resolved within machine precision for single-precision variables, and the subtraction of 1.0e-6 was essentially ignored.  I changed the value to 1.0e-5 for consistency with the intent based on the comments.  It's possible that simply computing 
`min( oc_top, cF_Lim(4)/sf)` 
works, depending on whether later "if" conditions use "<" or "<=".  The safer option is probably to subtract 1.0e-5, but that's unlikely to be zero-diff w.r.t. the branch before my change.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90
```diff
@@ -3761,28 +3771,69 @@ SUBROUTINE soil_para_hwsd (nx,ny, n_land, tile_pfs, tile_id)
...
+          endwhere
+       endif
+       
+       ! In all PEATMAP variants, subsurface orgC must *not* fall into peat class.
+       ! Reassign HWSD subsurface peat to peat-rich mineral Group 3.
```
gmao-rreichle: I am still wondering if this legacy behavior is what we wanted originally, and if it makes sense for the new processing approach.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90
```diff
@@ -4148,19 +4192,18 @@ SUBROUTINE soil_para_hwsd (nx,ny, n_land, tile_pfs, tile_id)
     !$OMP         sf,data_vec1,data_vec2,data_vec3,         &
     !$OMP         data_vec4,data_vec5,data_vec6,cF_lim,     &
     !$OMP         table_map,soil_class_top,soil_class_com,  &
-    !$OMP         soc_vec,poc_vec,use_PEATMAP)              &
-    !ncells_* not used !$OMP         soc_vec,poc_vec,ncells_top,ncells_top_pro,&
-    !ncells_* not used !$OMP         ncells_sub_pro,use_PEATMAP)               &
+    !$OMP         soc_vec,poc_vec,use_PEATMAP,              &
+    !$OMP         PEATMAP_STRICT_GPA22)                     &
```
gmao-rreichle: FYI, I removed "tiles_pfs" from the "!$OMP SHARED" statement (w.r.t. to the preceding commit) because "tiles_pfs" was not used within the scope of the !$OMP PARALLELDO

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90
```diff
@@ -4225,104 +4267,158 @@ SUBROUTINE soil_para_hwsd (nx,ny, n_land, tile_pfs, tile_id)
...
-             if (cFamily(4)/real(i) > PEATMAP_THRESHOLD_2) then 
-                o_cl  = 4
+          else if (PEATMAP_STRICT_GPA22) then
+             !
+             ! Strict GPA22 behavior:
+             ! classify peat only when peat support exceeds the threshold
+             ! relative to the full tile pixel count. Otherwise choose the
+             ! dominant mineral orgC class (1–3).
+             !            
+             if (cFamily(4)/real(i) > PEATMAP_THRESHOLD_2) then
```
gmao-rreichle, line 4345: I'm not sure this is the desired behavior. Why do we need a majority w.r.t. all raster grid cells, and not just the ones for which we have data?  If, say, the tile is made up of 3 raster grid cells, and we know that one of them is peat, and we don't have data for the other two, this statement would assigns a mineral class to the tile, right?  That seems counter-intuitive.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90
```diff
@@ -4225,104 +4267,158 @@ SUBROUTINE soil_para_hwsd (nx,ny, n_land, tile_pfs, tile_id)
...
-          if (sum(cFamily) == 0.) o_clp = 1
-          if (sum(cFamily)  > 0.) o_clp = maxloc(cFamily, dim = 1)        
+          else if (PEATMAP_STRICT_GPA22) then
+             !
+             ! Strict GPA22 behavior:
+             ! classify profile peat only when weighted peat support exceeds the
+             ! threshold relative to the full weighted tile support. Otherwise
+             ! choose the dominant mineral orgC class (1–3).
+             !            
+             if (cFamily(4)/(real(i) + 2.33*real(i)) > PEATMAP_THRESHOLD_2) then
```
gmao-rreichle: Two comments here:

1) Like for the top layer, for the tile to have o_clp=4 (i.e., profile is peat), the if condition requires that an absolute majority of all raster grid cells that make up the tile is peat, regardless of how many raster grid cells are no-data. 

2) I'm still wondering how we want to handle the peat class for the sub-layer.  Early on in the processing, we limit oc_sub to 8/sf (peat-rich mineral class).  Then here we compute a weighted average of orgC in the top and sub layers with the goal to determine the orgC profile class.  This makes little sense.  I guess our choice of oc_top=33/sf for peat raster grid cells ensure that the profile class becomes peat regardless of the value of oc_sub: even when oc_sub =0, we have oc_profile = (0.3*33+0.7*0)=9.9 > cF_Lim(4).   This works except for the case when the sub-layer orgC is good data but the top layer orgC is missing (e.g., outside the coverage of PEATMAP or GPA22).  Then the result is case-dependent.  
Would it make more sense to simply assign the peat class of the tile based on oc_top alone?  That is, if the tile has oc_top>cF_Lim(4), then o_cl = 4 **and** o_clp = 4 ?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90
```diff
@@ -4544,59 +4633,78 @@ SUBROUTINE soil_para_hwsd (nx,ny, n_land, tile_pfs, tile_id)
...
+          ! Legacy fallback for the case where only the top-layer class is missing
+          ! but the profile class is already defined.
+          if ((soil_class_top(n) == -9999) .and. (soil_class_com(n) >= 1)) then
+             if (.not. PEATMAP_STRICT_GPA22) then
+                soil_class_top(n) = soil_class_com(n)
+             endif
+          endif
+          
+          ! If the profile class is missing, search "neighboring" tiles for a fill value,
+          !   where "neighbors" are defined in terms of proximity of tile index (not necessarily geographical proximity)
```
gmao-rreichle: Note that "neighbor" here is simply a neighbor in indexing space, and not necessarily in geographic space.  This approach is used in the legacy approach and can't be changed for the legacy approach, but it may be worth thinking about something more geographically meaningful for "STRICT_GPA22" processing.  Not a high priority though, and probably doesn't impact many tiles anyway.  I'm adding the comment mostly so we can revisit the question later.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90
```diff
@@ -105,6 +106,7 @@ SUBROUTINE init_bcs_config(LBCSV)
     !   NGDC      : Soil parameters from Reynolds et al. 2000, doi:10.1029/2000WR900130 (MERRA-2, Fortuna, Ganymed, Icarus)
     !   HWSD      : Merged HWSDv1.21-STATSGO2 soil properties on 43200x21600 with Woesten et al. (1999) parameters   
     !   HWSD_b    : As in HWSD but with surgical fix of Argentina peatland issue (38S,60W)
+    !   HWSD_c    : HWSD updated from v1.2 to v2. And sub and top are D2 layer
```
gmao-rreichle: Instead of HWSD_c we should call this HWSDv2 for clarity.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90
```diff
@@ -61,6 +61,7 @@ module rmTinyCatchParaMod
   
   logical,      public, save :: use_PEATMAP = .false.
   logical,      public, save :: jpl_height  = .false.
+  logical,      public, save :: PEATMAP_STRICT_GPA22 = .false.
```
gmao-rreichle: The config variables and the variable names in the code will need some cleanup.  
- We won't need to run the legacy PEATMAP inputs with the "STRICT_GPA22" processing tweaks.  It would be cleaner to identify the name of the peat map data product, and have the if conditions based on the product name.  
- The name "PEATMAP" is specific to the legacy peat map, and the name of the "GPA22" map in the published literature is "Global Peat Map 2.0".  (I think GPA22 was just a working name during the earlier phase of product development.)
-  So instead of the logical "use_PEATMAP", we may want to have an integer config variables that something like:
 ```
   PEAT_INFO =              0 (HWSD), 
                            1 (HWSD and PEATMAP)
                            2 (Global Peat Map 2.0)
```
  E.g., the if condition "if (use_PEATMAP)" would then translate to something like:
```
select case (PEAT_INFO)
case (0)
   ! use_PEATMAP=.false.
case (1,2)
   ! use_PEATMAP=.true.
```
etc

**Review by gmao-rreichle — COMMENTED** [review]
@biljanaorescanin : Many thanks for making the processing changes.  I looked through the code in some detail but haven't yet had a chance to inspect the new bcs datasets.  I'll do that later, but I'm confident that's just a formality since you already checked them.
See below for a few more comments and open questions. 
Also, I added one commit that mostly edits the in-code comments but also makes one (probably non-zero-diff) change: https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1196/changes/3d03f10766d01d56847ade7f0109d056e86476fd?w=1
(The link hides white-space changes.)
We can discuss at the next tag-up.

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90 (hunk 4 above)
Comment (biljanaorescanin) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90:4345]: In the strict GPA22 path this is intentional.  
The goal here is to decide peat from tile scale support, not from the fraction among the subset of raster cells with valid HWSD OC. Using only valid cells in the denominator reintroduces the sparse valid failure mode we were fixing, where one or two peat valid pixels can force a tile to peat even though peat does not dominate the tile. So the full tile denominator is a conservative choice and is meant to stay consistent with the tile-scale pmap interpretation.

`I can change in code comment to make this clearer to something like this`:
! Strict GPA22 behavior:
! classify peat only when peat support exceeds the threshold relative to the
! full tile pixel count, not just the subset with valid HWSD OC. This avoids
! assigning peat from sparse valid peat pixels on boundary / inconsistent-mask tiles.
! Otherwise choose the dominant mineral orgC class (1–3).

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90 (hunk 5 above)
Comment (biljanaorescanin) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90]: I think this is a fair point. In the strict GPA22 path, we already remove sub-layer peat support earlier by capping oc_sub below the peat threshold. That means the weighted profile peat test here is no longer really the mechanism that assigns peat profile class. In practice, peat profile output is currently coming from the later PEATMAP rule that forces written profile peat when the top class is peat. So the intended science is really that top-layer GPA22 peat support determines tile peat, and profile follows consistently. 
A cleaner strict-GPA22 implementation would therefore be to set o_clp = 4 whenever the strict top peat condition is met, and otherwise use the weighted profile support only to distinguish mineral classes 1–3. 

Is that the behavior we want in the strict GPA22 path? I can change this if it is.

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90 (hunk 6 above)
Comment (biljanaorescanin) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90]: Agree. In this section, “neighbor” is still defined in index space, not geographic space. For the legacy path that needs to stay unchanged for backward compatibility. In the strict GPA22 path we only tightened the consistency of what gets copied once a donor is chosen, we did not redesign the donor search itself. A geographically based search could be cleaner for strict GPA22, but it would also be more computationally expensive. Since this fill is rare and the current behavior is already much improved, I’d treat that as a follow-up rather than part of this PR. But we can change this if you insist.

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90 (hunk 8 above)
Comment (biljanaorescanin) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90]: Yes, that makes sense. The current use_PEATMAP + PEATMAP_STRICT_GPA22 split was mainly introduced to isolate and validate the GPA22 logic while preserving legacy behavior. I agree that the long term configuration would be cleaner if it keyed off the peat information product explicitly rather than overloading the legacy PEATMAP name. I can work on cleaning that up and push a follow-up commit once it is done and retested.

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90 (hunk 2 above)
Comment (biljanaorescanin) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90]: I think that’s a fair question. In this PR I kept that behavior because the intended rule in the strict GPA22 path was that tile peat should be controlled by mapped top-layer peat support, not created independently by HWSD subsurface OC. So carrying forward the oc_sub cap was a way to prevent HWSD sub-layer peat from reintroducing peat outside the mapped top-layer signal.

In that sense it is internally consistent with the current strict approach, but I agree the implementation is not very transparent, especially since profile peat is then effectively made to follow top peat later in the PEATMAP path. If we want to revisit this more cleanly, I think the right rule would be: top-layer peat support determines whether the tile is peat, and profile follows consistently, rather than relying on capped oc_sub plus later write-time forcing.

For this PR I left the legacy oc_sub handling unchanged to minimize scope, but I’m happy to change it if you want the new processing path to express that rule more explicitly. Just let me know should I change this?

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90 (hunk 1 above)
Comment (biljanaorescanin) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90]: I think you’re right that the 1.0e-6 version was too close to single-precision resolution at cF_lim(4) ≈ 8.72, so changing it to 1.0e-5 is at least more consistent with the stated intent.

That said, I think the deeper issue is that this is still using NINT(...), which rounds to nearest. The intent here is to guarantee that outside the GPA22 peat mask we cap oc_top so that oc_top*sf stays strictly below cF_lim(4), since the later peat test uses >= cF_lim(4).

In that sense, `min(oc_top, cF_lim(4)/sf)` is not really safe, and `NINT((cF_lim(4)-eps)/sf)` is still a bit fragile. Maybe better expression would be something like:

`oc_top = min(oc_top, ceiling(cF_lim(4)/sf) - 1)`

because that gives the largest integer coded value whose scaled value remains strictly below the peat threshold. If you prefer to minimize change for now, 1.0e-5 is better than 1.0e-6, but I agree it is not the cleanest formulation.

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90 (hunk 4 above)
Comment (gmao-rreichle) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90:4345]: To clarify, I had thought the spatial support should be based on that of the peat map inputs.  I realize now that the problem here is that the peat map is seemingly "complete" in that it doesn't have any no-data-values, at least not after preprocessing.  It's defined everywhere, including the middle of the ocean.  The original peat map tif file probably also doesn't have no-data values, so chances are we can't do better than what you have now.

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90 (hunk 1 above)
Comment (gmao-rreichle) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90]: Changed to `oc_top = min(oc_top, FLOOR((cF_lim(4)/sf) )  ! force oc_top<=872   (cF_lim(4)/sf=872.0930)`

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90 (hunk 4 above)
Comment (gmao-rreichle) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90:4345]: PS: The geotiff file contains only 255, 1, and 2.  That is, there is no distinction between ocean and "no peat" over land, and there is no sensible mask for the coverage of the peat map.

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90 (hunk 5 above)
Comment (gmao-rreichle) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90]: I'm wondering if this if condition:
`if (cFamily(4)/(real(i) + 2.33*real(i)) > PEATMAP_THRESHOLD_2)`
can ever be true. In the loop above where `cFamily` is computed, we should never get `cFamily>real(i)` because oc_sub was earlier capped at `cF_lim(4)`, so the sub layer should never contribute to the sum.  After normalization with `(real(i) + 2.33*real(i))`, we should then always be below `PEATMAP_THRESHOLD_2`.   In other words, we would never get `o_clp=4` from this section of the code.   
This would then always put us into the "mineral" block of the code below (~Line 4524), and `poc_vec` would be computed from the organic carbon content values of the contributing _mineral_ raster grid cells.
What I don't understand is how still we end up with `soil_class_com=253` even when `o_clp<=3`.

**Review by gmao-rreichle — COMMENTED** [review]
@biljanaorescanin : I still don't understand how this code works.  See comment below

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90 (hunk 5 above)
Comment (biljanaorescanin) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90]: Yes, I think that comment is right. 
In the strict GPA22 path, after oc_sub is capped below the peat threshold, the weighted profile peat test can no longer make o_clp = 4. 
The reason we still see profile peat in the output is that later in the PEATMAP path the written profile class fac is forced to 253 whenever the top class is peat. 
In addition, soil_class_com(n) itself can still become 253 through the strict fallback (soil_class_com < 1 -> soil_class_top) or neighbor fill. So the output is peat/peat, but the current implementation is still mixed.

I think I found fix in my notes and will try that and see.

Comment (gmao-rreichle) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90]: > In the strict GPA22 path, after oc_sub is capped below the peat threshold, the weighted profile peat test can no longer make o_clp = 4.

Then the if block makes no sense at all.

> The reason we still see profile peat in the output is that later in the PEATMAP path the written profile class fac is forced to 253 whenever the top class is peat.

Where is that exactly? 

> In addition, soil_class_com(n) itself can still become 253 through the strict fallback (soil_class_com < 1 -> soil_class_top) or neighbor fill. 

This should not happen very often.  But we get %OCProf<8.72 for nearly all peat tiles.

Comment (biljanaorescanin) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90]: The place where written profile peat is still being forced is the later PEATMAP output block:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/bb3f763ab94e2ec9c29fe0dc55e581c5bf505c31/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90#L4712-L4720
So the final written prof can become peat even when the upstream profile classification path remained on the mineral side.

Comment (gmao-rreichle) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90]: Ok, thanks.  But at this point in the code, only the profile soil class is changed, and the associated parameters (e.g., poc_vec [%OCProf]) remain unchanged, even though for peat tiles they were calculated from the _mineral_ raster grid cells.  
It will NOT work to add modifications of poc_vec etc here. For backward compatibility, we need to keep the legacy path as is.  For "STRICT GPA22", we need more invasive changes further up.   It makes no sense to reset oc_sub to mineral, and then diagnose o_clp from that.  After the reset, o_clp can never be peat, at least not if it's based on the logic in what's currently in L4347.
My hope is that with the appropriate in-line "if (STRICT GPA)" conditions, we can salvage the overall structure of the code and fix the new logic while maintaining legacy calcs in one subroutine.  If we had more time and resources, we'd want to rewrite all of this stuff from scratch in a new subroutine soil_para_hwsd_peat() or some such, but that would almost certainly require more effort than we can afford at this time.

Comment (biljanaorescanin) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90]: Last[ commit](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1196/changes/18a3ad45e911f15c2e07de01d165ae1d500c245e) should resolve this. I've shared BCs set produced with this commit with you on chat to confirm.

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90 (hunk 6 above)
Comment (biljanaorescanin) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90]: I need to revise my earlier response to this question after the follow-up investigation.

The concern about index-space “neighbors” is more relevant to STRICT_GPA22 than I first implied. The raw GPA22 → conservative PEATMAP rasterization checks out, and the first discrepancy appears later, in the PEATMAP → M09 stage. In the strict path, the early donor fill still uses index-space neighbor selection, and once a donor is chosen it can copy peat-defining state (soil_class_top, soil_class_com, pmap) from that donor. So this is now a plausible source of the most suspicious anomaly subset we found.

In particular, the strongest anomaly set is small: 20 tiles where the raw GPA22 tile box is fill-only but M09 still ends up peat; 18 of those 20 are in Maritime SE Asia. That makes the donor-fill behavior worth tightening in STRICT_GPA22.

If we want to resolve this in the PR, I see two options:

Option 1 (minimal fix): keep the legacy donor search, but in STRICT_GPA22 do not allow the fill step to create peat identity from the donor. In practice, that means tightening the strict-path fill so donor-selected values cannot newly assign peat-defining state (soil_class_top, soil_class_com, pmap) to a target tile without direct strict GPA22 / PEATMAP support.

Option 2 (cleaner fix): switch donor selection in STRICT_GPA22 from index-space to geographic-space.

My preference for this PR is Option 1, because it enforces the intended strict GPA22 logic with much less algorithmic risk. Option 2 is conceptually cleaner, but it is a broader redesign and computationally more expensive.

The main location that needs attention is here:
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/cbf7ca74fe8266d931f9ccb78aea60d7f2316be1/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90#L4660-L4682

I have an idea for the minimal fix and will test it to see whether it resolves this anomaly set.

Comment (biljanaorescanin) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90]: Update: Unfortunately, after testing, this block does not appear to be responsible for the 20-tile anomaly subset. The strict donor-fill path never triggered in this case, and replacing it did not change the results.

Comment (biljanaorescanin) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90]: Update2: I rechecked the strict GPA22 peat implementation and did not reproduce a code bug. The exact in-code tile validation passes with zero disagreements: tile-level pmap computed directly from the strict PEATMAP mask matches the model produced tile-level pmap exactly (model-only peat = 0, mask-only peat = 0, max |delta| = 0).

My earlier bug report came from a validation method bug on my side: I was comparing original raw GPA22 aggregated offline with a geometry based approximation to the final model peat decision, which is not the exact same object the code uses.

The correct validation is two-step. First, compare raw GPA22 class-1-only to the shared conservative PEATMAP raster on the same grid; that check also showed no mismatch. Second, compare tile-level pmap computed from that strict PEATMAP mask to the model’s tile-level pmap; that comparison is exact and shows zero difference.

The residual differences I still see in the rawGPA22-to-M09 offline plots are therefore diagnostic artifacts from the coarser offline aggregation, not evidence that the strict peat code is creating unsupported peat.

**Comment by biljanaorescanin** [comment]
For reference:
v12 BCs generated from this branch are currently identical (zero-diff) to those from the v12 develop baseline.

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90 (hunk 8 above)
Comment (biljanaorescanin) [reply GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/rmTinyCatchParaMod.F90]: This was addressed in below @gmao-rreichle commit

**Comment by biljanaorescanin** [comment]
Files are added here: `make_bcs_inputs/land/soil/SOIL-DATA/v2` 
I didn't remove `GPA22` from file names under upper path since `tiff` original data we download has `GPA22` in its name. 
But we can change this if its desirable.  
@gmao-rreichle  I've also resolved most of the comments since some were outdated and some you implemented in last set of commits.

File: GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/soil/peat/v2/convert_peat_tiff_to_nc4.py [hunk 9]
```diff
@@ -0,0 +1,45 @@
+import rasterio
+import numpy as np
+import netCDF4 as nc
+
+# Files
+tif_file = "/discover/nobackup/projects/gmao/bcs_shared/make_bcs_inputs/land/soil/SOIL-DATA/v2/peatGPA22WGS_2cl.tif"
+nc4_file = "/discover/nobackup/projects/gmao/bcs_shared/make_bcs_inputs/land/soil/SOIL-DATA/v2/peatGPA22WGS_2cl.nc4"
```
gmao-rreichle: The nc4 file does not exist in this location.  Do we really need to have an nc4 version of the original tif file?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/soil/peat/v2/build_gpa22_peat_rasters_30arcsec.py
```diff
@@ -0,0 +1,359 @@
...
+
+# Source choice:
+#   "nc"  -> read the existing GPM 2.0 (GPA22) NetCDF file
+#   "tif" -> read the original GeoTIFF directly
+# For strict reproduction of your current files, keep this as "nc" first.
+GPA_SOURCE_KIND = "nc"
+
+GPA_NC_FILE = Path(
+    "/discover/nobackup/projects/gmao/bcs_shared/make_bcs_inputs/land/soil/SOIL-DATA/v2/peatGPA22WGS_2cl.nc4"
+)
```
gmao-rreichle: Do we really need the nc4 version of the tif file?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/soil/peat/v2/build_gpa22_peat_rasters_30arcsec.py
```diff
@@ -0,0 +1,359 @@
...
+# Which products to create:
+#   "conservative" -> class 1 only
+#   "alpha050"     -> class 1 = 1.0, class 2 = 0.5
+#   "liberal"      -> class 1 or 2 both treated as 1.0
+#   "all"          -> write all three files
+PRODUCT_MODE = "all"
+
+OUTFILE_CONSERVATIVE = Path("PEATMAP_from_GPA22_like_old_conservative.nc4")
+OUTFILE_ALPHA050     = Path("PEATMAP_from_GPA22_like_old_alpha0.50.nc4")
+OUTFILE_LIBERAL      = Path("PEATMAP_from_GPA22_like_old_liberal.nc4")
```
gmao-rreichle: We only need the "alpha0.50" file.  If we ever wanted to switch to "conservative" or "liberal", the "alpha0.50" file provides complete information. 
We don't need to clean up the code and can leave the "liberal" and "conservative" processing in place, but we should only archive the "alpha0.50" file. 
For clarity, the file name for the "alpha0.50" should be something like:
`peatGPA22WGS_2cl_real_30arcsec.nc4`
The idea is that we really just remap the data from the original (0.01-deg) lat/lon grid to our 30-arcsec raster grid.  In doing so, we change the file format from tif to nc4, and we assign real numbers (1.0, 0.5) to classes (1,2).

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/soil/peat/v2/build_gpa22_peat_rasters_30arcsec.py
```diff
@@ -0,0 +1,359 @@
...
+)
+GPA_TIF_FILE = Path(
+    "/discover/nobackup/projects/gmao/bcs_shared/make_bcs_inputs/land/soil/SOIL-DATA/v2/peatGPA22WGS_2cl.tif"
+)
+
+# Target 30-arcsec grid used by the legacy PEATMAP product
+PEATMAP_GRID_FILE = Path(
+    "/discover/nobackup/projects/gmao/bcs_shared/make_bcs_inputs/"
+    "land/soil/SOIL-DATA/PEATMAP_mask.nc4"
+)
```
gmao-rreichle: The dependency on the old PEATMAP_mask.nc4 file should be removed.  The important thing is that the GPM 2.0 data are on our usual 30-arcsec global raster grid.  This is also the case for the data in PEATMAP_mask.nc4.  The only difference in PEATMAP_mask.nc4 is that the lat/lon values are in real*4, which is not appropriate for global lat/lon vectors are 30-arcsec resolution.  I'm pretty (but not absolutely) sure that we're not using the lat/long values from the nc4 file in our processing.  For now, I think we should just construct the lat/long values as:

```
dlat = 180./21600                                  (double precision)
lat = (-90+dlat/2):dlat:(90-dlat/2)                (double precision)
```

and similarly for longitude, then write the coords into the nc4 file.   (Might need to pay attention to the ordering of the lats and longs to match up the array orientation across the two files.  If this is wrong, it'll readily show when we test the v14 bcs generation against my data.)

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90
```diff
@@ -3761,28 +3777,106 @@ SUBROUTINE soil_para_hwsd (nx,ny, n_land, tile_pfs, tile_id)
...
+       case (1)
+          ! Legacy PEATMAP mask 
+          fname = trim(MAKE_BCS_INPUT_DIR)// '/land/soil/SOIL-DATA/PEATMAP_mask.nc4'
+       case (2)
+          
+          !  ! Preprocessed GPM 2.0 peat mask: 1 = peat; 0 = mineral (includes reclassified "2=peat in soil mosaic")
+          !  fname = trim(MAKE_BCS_INPUT_DIR) // '/land/soil/SOIL-DATA/v2/PEATMAP_from_GPA22_like_old_conservative.nc4'
+
+          ! Preprocessed GPM 2.0 peat mask: 1 = peat dominated, 0.5 = peat in soil mosaic, 0 = mineral)
+          !fname = trim(MAKE_BCS_INPUT_DIR) // '/land/soil/SOIL-DATA/v2/PEATMAP_from_GPA22_like_old_alpha0.50.nc4'
```
gmao-rreichle: The directory tree structure and contents of 
`trim(MAKE_BCS_INPUT_DIR) // '/land/soil/SOIL-DATA/v2/`
needs some revision.  We can discuss offline

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mod_process_hres_data.F90
```diff
@@ -3761,28 +3777,106 @@ SUBROUTINE soil_para_hwsd (nx,ny, n_land, tile_pfs, tile_id)
...
+          fname = '/discover/nobackup/rreichle/PEATMAP_from_GPA22_like_old_alpha0.50.nc4'
```
gmao-rreichle: Need to update with final file name / location:  `trim(MAKE_BCS_INPUT_DIR) // ...`

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/soil/soil_properties/v2/README_code_sequence
```diff
@@ -0,0 +1,275 @@
+Updated: Dec 2025
```
gmao-rreichle, line 1: adding a new v2 dir for soil preproc scripts is good, but we should probably need to add a "v1" dir for the existing scripts.

gmao-rreichle (commented):
@biljanaorescanin : See comments below.  We can discuss next week

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/soil/peat/v2/convert_peat_tiff_to_nc4.py
biljanaorescanin: No, it was more for convenience for people. I can remove all reference to that. Result is zero diff if you use tiff or nc4. I can also add nc4 if you want it, some people don't like tiff.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/soil/peat/v2/build_gpa22_peat_rasters_30arcsec.py
biljanaorescanin: Nope, same as previous comment. If you don't want it I will remove it.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/soil/peat/v2/convert_peat_tiff_to_nc4.py
gmao-rreichle: I think we should remove the nc4 file.  It's confusing.  Note that right now there's the following comment in the build_gpa22*.py script: 
```
# For strict reproduction of your current files, keep this as "nc" first.
GPA_SOURCE_KIND = "nc"
```
This comment is probably also outdated and should be removed.

I'm happy to keep the py script that converts tif to nc4, in case someone does want to have an nc4 version of the tif file in the future.  But we should add a note at the top of this tif2nc4 conversion script that it's meant as a separate tool and not required for the preprocessing workflow.

biljanaorescanin: I've kept converter and added comments to it. Removed from preprocessing nc4 usage (just tiff now).

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/soil/peat/v2/build_gpa22_peat_rasters_30arcsec.py
biljanaorescanin: Renamed file. Removed one not used.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/soil/peat/v2/build_gpa22_peat_rasters_30arcsec.py
biljanaorescanin: Updated preprocessing not to use PEATMAP file grid but as you suggested.

biljanaorescanin: Note that this change in lat/lon values changes the generation of the GPA nc4 peat mask on the 30 arcsec raster grid.  This was updated in the make_bcs_inputs dir.

gmao-rreichle (approved):

sdrabenh (approved):

PR #1198, opened 2026-01-30
v11: Fixed a memory leak in mkIAU.

atrayano:
This PR fixes a memory leak caused by wrongly passing a filename to MAPL_read_bundle instead of a file template. Then MAPL cannot properly close the netCDF after done reading it.


atrayano:
This PR needs MAPL v2.65.0. Otherwise, there will be run-time failures due to improper formatting issues of pFlogger

sdrabenh (approved):

PR #1201, opened 2026-02-04
v12 shocmf updates

narnold1:
SHOCMF updates.  Adds v2.0 of double gaussian PDF, removes CLASP and LTS options from EDMF, various stability improvements and code cleanup.  0-diff *with exception of turb_internal_rst*: new internals were added, used by SHOC only.


mathomp4:
@narnold1 GNU has an issue but I think it's been fixed in upstream v12. When ready, you can press the "Update branch" button and the CI should be happy.

mathomp4:
@narnold1 The build error seems to be saying something is missing in the call to `partition_dblgss`. There aren't enough arguments in the call.

The call does:
```
​            call partition_dblgss(fQi,          &
                                  TEn,          &
                                  QVn,          &
                                  QCn,          &
                                  0.0,          &
                                  ZL,           &
                                  PL*100.,      &
...
```
but the subroutine has:
```
subroutine partition_dblgss( fQi,          &  ! IN
                              tabs,         &
                              qwv,          &
                              qc,           &  ! OUT
                              qsat,         &
!                              qi,           &
                              omega,        &  ! IN
                              zl,           &
                              pval,         &
```

So it's either missing `qsat` or `omega`. Something is being passed in as `0.0` but I'm not sure what.

narnold1:
Yes, "qsn" should be passed in after QCn.  omega is passed in as 0.

sdrabenh:
My tests reverting the [negative protection change](https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/ed81cc92d723d61398313ec3f8f1f93e993ce7b4/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/Process_Library.F90#L2419) show this PR passes:
 1. All regression tests
 2. 0-inc replay test
 3. All restarts from AMIP and Replay modes are 0-difference EXCEPT:
    * `turb_import_checkpoint` long_name changed from "surface geopotential height" to "surface geopotential"
    * `turb_internal_checkpoint` added variables: `MFAW`, `QTFLXMF`, `SLFLXMF`

PR #1212, opened 2026-02-19
get issm exports in landice tile space

weiyuan-jiang:
Since ISSM's exports are on mesh-like locstream, the variables on landice-tile space cannot be exported. This PR uses internal state to pass the variables from ISSM gridcomp back to landice gridcomp and use landice's export for history output. @agstub , this PR is not done. I have left some place for you to implement. Please take a look and let me know your concerns.

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSlandice_GridComp/GEOS_LandIceGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSlandice_GridComp/GEOS_LandIceGridComp.F90
@@ -44,7 +44,9 @@ module GEOS_LandiceGridCompMod
   use GEOS_UtilsMod
   use DragCoefficientsMod
   use GEOS_IssmGridCompMod,   only : IssmSetServices  => SetServices
-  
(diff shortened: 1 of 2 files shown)


agstub:
so the "whatever" [here](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1212/changes#diff-4375805f88644e3d498571098c548ab865fdf6b89e8c92e7ffb62b8fbda8f98bR676) should have the (local) size of landice tiles? So the idea is I can transform between mesh and landice tiles like before (with ESMF_FieldRegrid and MAPL_LocStreamTransform), set the landice tile version in this new internal state, then recover these pointers via internal state in the landice gridcomp? Am I thinking about it right?

weiyuan-jiang:
> so the "whatever" [here](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1212/changes#diff-4375805f88644e3d498571098c548ab865fdf6b89e8c92e7ffb62b8fbda8f98bR676) should have the (local) size of landice tiles? So the idea is I can transform between mesh and landice tiles like before (with ESMF_FieldRegrid and MAPL_LocStreamTransform), set the landice tile version in this new internal state, then recover these pointers via internal state in the landice gridcomp? Am I thinking about it right?

Yes, that is local size

PR #1213, opened 2026-02-20
Mst Phy change that is in ADAS 5.43 and needed in 5.44

rtodling:
In updating the ADAS from v5.43 to v5.44 I am also updating the model from something close to v11.7.0 (plus) to v11.7.3. Since ADAS v5.43 includes a change to the moist physics to partly address the warming at low level temperatures, the same change needs to be brought what will become ADAS v5.44.

This is a one file update to GEOSgcm_GridComp v2.7.6 (which is in the GEOSgcm v11.7.3). 

This is aimed at a side-branch GEOS-FP-5_44 and ideally would be tagged v2.7.6.1.

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/Process_Library.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/Process_Library.F90
@@ -405,26 +405,16 @@ module GEOSmoist_Process_Library
      ! Sigmoidal functions like figure 6b/6c of Hu et al 2010, doi:10.1029/2009JD012384
       if (SRF_TYPE >= 2.0) then
         ! Over snow (SRF_TYPE == 2.0) and ice (SRF_TYPE == 3.0)
-        if (ICE_RADII_PARAM == 1) then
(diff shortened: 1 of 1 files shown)


mathomp4 (approved):

rtodling:
@mathomp4 Why is this saying one CI test not successful?! looking at the CI log this seems more like a CI issue than an actual error, doesn't it?

mathomp4:
> @mathomp4 Why is this saying one CI test not successful?! looking at the CI log this seems more like a CI issue than an actual error, doesn't it?

@rtodling Yeah. Something wonky is going on with Github it seems. It's saying it's unauthorized to do do something it's been doing for months. For now just ignore and I'll try and figure it out...

rtodling:
@mathomp4 anyway, this is now merged to GEOS-FP-5_44 - would you be able to tag it as I suggested above?

mathomp4:
> @mathomp4 anyway, this is now merged to GEOS-FP-5_44 - would you be able to tag it as I suggested above?

@rtodling I've tagged [v2.7.6.1](https://github.com/GEOS-ESM/GEOSgcm_GridComp/releases/tag/v2.7.6.1)

PR #1223, opened 2026-03-12
Handmerge develop into v12 2026-Mar-12

mathomp4:
This is a hand merge of `develop` into `feature/sdrabenh/gcm_v12` as of 2026-Mar-12

I see changes:

* AGCM from @FlorianDeconinck (should be safe)
* In Raster code so I summon @biljanaorescanin (I think this is the update from @atrayano to make it faster)
* Dataatm and CICE so I summon @zhaobin74 

The main things we need to know is: Should these changes go into v12?

The DataAtm and CICE ones are the ones I'm not sure about.

Also: Not sure how to label.


zhaobin74:
Thanks, Matt. Both the DataAtm and ice_in fix are needed in v12 as well as v11.

FlorianDeconinck:
Changes `AGCM_GridComp` are related to the experimental Fortran < - > Python bridge with MAPL support. This is _not_ on by default and will be a pass-through function call with no runtime impact without a bespoke build flag.

It is safe for v12.

PR #1231, opened 2026-03-25
[DSL] Moist physics: UW, GFDL_1M and GF2020

FlorianDeconinck:
This PR brings the work of @katrinafandrich and @CharlesKrop to mainline GEOS.

It cover the first DSL-powered version of the UW and GFDL_1M parametrization. All the files exists in a `pyMoist` directoy within the `GEOSmoist_GridComp` directory. The `pyMoist` is a self-sustaining package but is _not_ use like one for runs on Discover (see https://github.com/GEOS-ESM/GEOSgcm_App/pull/826 PR and https://github.com/GEOS-ESM/ESMA_env/pull/178).

The status of the parametrizations are:
- GFDL_1M: first SCM experiments are showing some small instability. Expect follow up fixing PR.
- UW: code is not up to DSL standards yet but is functional with good results on SCM experiments. Follow up PR to clean up and raise the code to DSL standards.
- GF2020: code is slowly drifting but within reason. Expect following fixing PR.

Continuous integration:
- A new github workflow `ndsl-checks` to run specific checks for `pyMoist` code on a `paths` hooks

Build:
- Changed `GEOSmoist_GridComp` to a shared library so the python dynamically open it and fetch fortran hooks
- Requires dependencies `fms_r4` to be spelled out for the previous change to work
- Install `pyMoist` python code in `lib/Python`

Change to the Fortran:
- MoistGridComp
  - A new `Finalize` function to allow for profiler data collection in the DSL
- GFDL_1M_Interface
  - Introducing a `USE_PYMOIST_GFD1M` flag to turn on/off the DSL code
  - Introducing custom code for convection tracers that move them to Structure of Arrays - fitting the DSL needs
  - A new `Finalize` function to allow for profiler data collection in the DSL
- UW_Interface
  - Introducing a `USE_PYMOIST_UW` flag to turn on/off the DSL code
  - A new `Finalize` function to allow for profiler data collection in the DSL

--- 

Closes #1232

PR #1232, opened 2026-03-26
v11: Remove FMS dependence in Moist

mathomp4:
PR #1231 from @FlorianDeconinck introduces a *library* dependence on FMS (due to moving to a shared library). This is not good for the rest of GEOS.

However, v12 no longer cares about FMS as the GFDL microphysics in there converted FMS and MPP calls to Fortran and MAPL. So we do the same here. If this works, then @FlorianDeconinck can use the same changes in his branch.

It is zero-diff. So I'll undraft but if these changes are put into #1231 by @FlorianDeconinck then this can be closed.


FlorianDeconinck:
I believe this can be closed because of https://github.com/FlorianDeconinck/GEOSgcm_GridComp/pull/13 flowing into #1231 - which will bring the dependency removal to bear

mathomp4:
> I believe this can be closed because of [FlorianDeconinck#13](https://github.com/FlorianDeconinck/GEOSgcm_GridComp/pull/13) flowing into #1231 - which will bring the dependency removal to bear

Very true!

Issue #1359, opened 2026-04-01
Deactivate non-ported components on release/MAPL-v3


tclune:
## Context

On the `release/MAPL-v3` branch we are preparing to flip `use MAPL` to import MAPL3 public entities. Components that have not yet been ported to MAPL3 will break at that point and will cause CI failures.

To protect CI we are temporarily deactivating non-ported components in CMake by commenting out their `add_subdirectory` / `SUBCOMPONENTS` entries. This makes it straightforward to re-activate each component as it is ported — just uncomment the relevant lines.

## Components currently ported (remain active)

- `GEOSgwd_GridComp` (GWD)
- `FVdycoreCubed_GridComp` (DynCore) — lives in its own repo, no change needed there

## Changes required in this repo

**`GEOSagcm_GridComp/GEOSphysics_GridComp/CMakeLists.txt`**

- Comment out non-ported subdirs from `alldirs`: `GEOSchem_GridComp`, `GEOSmoist_GridComp`, `GEOSsurface_GridComp`, `GEOSturbulence_GridComp`, `GEOSradiation_GridComp`
- Leave `GEOSgwd_GridComp` active
- Replace the `esma_add_library` / `else` block with a simple `esma_add_subdirectories(${alldirs})` call (the top-level `GEOS_PhysicsGridComp.F90` source `USE`s non-ported components, so we bypass it for now)
- Comment out the original `if/esma_add_library/else/esma_add_subdirectories/endif` block with a note explaining how to restore it

**`GEOSagcm_GridComp/GEOSsuperdyn_GridComp/CMakeLists.txt`**

- Comment out non-ported subdirs from `alldirs`: `FVdycore_GridComp`, `GEOSdatmodyn_GridComp`, `ARIESg3_GridComp`
- Leave `FVdycoreCubed_GridComp` active
- Replace the conditional library block with a simple `esma_add_subdirectories(${alldirs})` (bypass `GEOS_SuperdynGridComp.F90` which `USE`s non-ported components)
- Comment out the original block with a restore note

## Feature branch

`feature/deactivate-nonported-mapl3-components`

## Related issues

- GEOSchem_GridComp: (to be filed)
- GOCART: (to be filed)

## Reverting

To re-activate a component as it is ported, simply uncomment its entry in `alldirs` and, once all components in a given `CMakeLists.txt` are restored, uncomment the `esma_add_library` block and remove the temporary `esma_add_subdirectories` bypass.

[...]

PR #1360, opened 2026-04-01
Deactivate non-ported components on release/MAPL-v3

tclune:
## Summary

- Comments out all top-level siblings except `GEOSagcm_GridComp`; bypasses `GEOSgcm_GridComp` top-level library
- Bypasses `GEOSagcm_GridComp` top-level library
- `GEOSphysics_GridComp`: comments out all physics subdirs except `GEOSgwd_GridComp` and `GEOSchem_GridComp` (the latter restricted to GOCART-only)
- `GEOSsuperdyn_GridComp`: comments out all subdirs except `FVdycoreCubed_GridComp`; bypasses top-level library
- Removes spurious `GEOS_Shared` dependency from `GEOSgwd_GridComp` (no source file actually uses it)

## Context

Part of the MAPL3 flip work tracked in GEOS-ESM/GEOSgcm_GridComp#1359. Only components already ported to MAPL3 are compiled on `release/MAPL-v3`.

### DIFF
--- a/GEOSagcm_GridComp/GEOSsuperdyn_GridComp/CMakeLists.txt
+++ b/GEOSagcm_GridComp/GEOSsuperdyn_GridComp/CMakeLists.txt
@@ -2,28 +2,33 @@ esma_set_this ()
 
 set (alldirs
   FVdycoreCubed_GridComp
-  FVdycore_GridComp
-  GEOSdatmodyn_GridComp
-  ARIESg3_GridComp
+  # FVdycore_GridComp       # not yet ported to MAPL3 - restore when ported
+  # GEOSdatmodyn_GridComp   # not yet ported to MAPL3 - restore when ported
+  # ARIESg3_GridComp        # not yet ported to MAPL3 - restore when ported
   )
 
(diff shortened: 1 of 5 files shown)


GEOSagcm_GridComp/GEOSsuperdyn_GridComp/CMakeLists.txt
```diff
@@ -2,28 +2,33 @@ esma_set_this ()
 
 set (alldirs
   FVdycoreCubed_GridComp
-  FVdycore_GridComp
-  GEOSdatmodyn_GridComp
-  ARIESg3_GridComp
+  # FVdycore_GridComp       # not yet ported to MAPL3 - restore when ported
```
mathomp4, line 5: One note here for @tclune / Claude: I have a set of PRs to v12 to *remove* the Latlon and ARIES cores.

So it might not even be worth the effort to port.

Datmo we definitely need for the SCM.

tclune: Oh - I was definitely **not** going to port those.   My fear is that there are other less, er, infamous, components that I would not so recognize.

PR #1365, opened 2026-04-09
PBLH library for turbulence

narnold1:
This moves calculation of PBL height diagnostics from the TurbulenceGridComp into a new diagnostics module.


mathomp4:
I think the current CI failure will be fixed when the "Update branch" button is pressed

sdrabenh:
Adding do not approve label temporarily until after v12 is released

Issue #1369, opened 2026-04-13
GEOSgcm v12 fails with ifort Debug - 2026-Apr-13


mathomp4:
Test runs of GEOSgcm v12 with the latest Moist seems to throw an error with ifort Debug. For example at C24 with my usual 2000-04-14 21z restarts:

```
 Writing:    162 Slices to File:  stock-v12-2026Apr13-1day-c24-DEBUG.geosgcm_surf.20000415_1330z.nc4
 Writing:     48 Slices to File:  stock-v12-2026Apr13-1day-c24-DEBUG.geosgcm_ocn.20000415_1430z.nc4
 Writing:      8 Slices to File:  stock-v12-2026Apr13-1day-c24-DEBUG.geosgcm_budi.20000415_1500z.nc4
 Writing:     21 Slices to File:  stock-v12-2026Apr13-1day-c24-DEBUG.geosgcm_buda.20000415_1330z.nc4

 GCM Date: 2000/04/15  Time: 15:00:00  Throughput(days/day)[Avg Tot Run]:        120.0         76.1         80.5  TimeRemaining(Est) 000:02:59   18.5% :  34.4% Mem Comm:Used  Wallclock: 2026/04/13 13:24
 GCM Date: 2000/04/15  Time: 15:20:00  Throughput(days/day)[Avg Tot Run]:        120.7        166.5        168.3  TimeRemaining(Est) 000:02:49   18.3% :  34.4% Mem Comm:Used  Wallclock: 2026/04/13 13:24
 GCM Date: 2000/04/15  Time: 15:40:00  Throughput(days/day)[Avg Tot Run]:        121.2        166.3        168.0  TimeRemaining(Est) 000:02:38   18.3% :  34.4% Mem Comm:Used  Wallclock: 2026/04/13 13:24
forrtl: error (65): floating invalid
Image              PC                Routine            Line        Source
libpthread-2.31.s  0000149EB7D39910  Unknown               Unknown  Unknown
GEOSgcm.x          00000000086C5E23  uwshcu_mp_compute        2246  uwshcu.F90
GEOSgcm.x          0000000008655E49  uwshcu_mp_compute         313  uwshcu.F90
GEOSgcm.x          0000000008617408  geos_uw_interface         455  GEOS_UW_InterfaceMod.F90
GEOSgcm.x          00000000069849FE  geos_moistgridcom        5954  GEOS_MoistGridComp.F90
[...]

PR #1371, opened 2026-04-14
Feature/sdrabenh/rc23 mg3

sdrabenh:


mathomp4:
In my tests, this is zero-diff to v12 save for the removal of `ALH` from the `moist_import_checkpoint`:

```
Comparing NC4 moist_import_checkpoint using nccmp...
Failure!
Checking for data differences
Checking for metadata differences
DIFFER : VARIABLE : ALH : DOES NOT EXIST IN "mg3merge-v12-2026Apr14-1day-c24//scratch/moist_import_checkpoint"
```

Both BACM-L72 and GFDL-L181 show this.

Also, it seems to have passed `gcm_regress.j`

And, more weirdly, #1369 is "fixed" by this...making me think it's a memory issue. Yay.

sdrabenh:
My tests show similar results. The only non-zero-diff result is the removal of the `ALH` variable from the `moist_import` restart. This PR builds, runs, passes all CI, and regression tests for both L72 and L181 default configurations. Also builds in debug. Additionally, results are 0-diff for AMIP, Replay, and 0-inc modes except for the `ALH` variable mentioned above. Note MG was not tested itself.

PR #1372, opened 2026-04-15
v11: two phase initialization for CICE/CICEPlug

zhaobin74:
This PR added a second phase of initialization for CICEPlug and CICE6 such that the previous delayed part in thermo callback is not needed anymore.

Fixed https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/1197

Thanks, coauthor @atrayano

PR #1373, opened 2026-04-15
v12: two phase initialization for CICE/CICEPlug

zhaobin74:
Same as https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1372 but for V12.


gmao-rreichle (commented):
doesn't seem to touch the land or preprocessing, so ok with me

gmao-rreichle (approved):

PR #1374, opened 2026-04-15
v11: Fix bug in Turb GC (NZD)

mathomp4:
@mmanyin spotted what seems to be a bug in v11 Turb GC that has been there for...a long time. Like 7 years at least.

It looks like @wmputman or someone fixed it **_in v12_** in https://github.com/GEOS-ESM/GEOSgcm_GridComp/commit/7ec7c46d5a1ebe07583a7ebf603d4f64fa0e2b4d but not in v11 (or v10 or...). Gemini helped spot this with some comparison to v12 code.

This changes the state so very non-zero-diff. So, I am going to mention...everyone. This seems to touch surface and `SYNCTQ` and...

cc: @wmputman @narnold1 @atrayano @gmao-rreichle @amolod 

NOTE: maybe this is a very minor change on the whole, but, well, I'm not qualified to comment on how NZD something is. Gemini like most ML models tends to be a bit...expansive. It says "massive implications" when in reality it might just be "okay, it's like an extra leaf fell to the ground" or something.

---

Per Gemini:

## Fix for Turbulence Diffusion Diagnostics (VAFDIFFUSE, QAFDIFFUSE, etc.)

This PR fixes a bug in the `DIFFUSE` subroutine of `GEOS_TurbulenceGridComp.F90` where the post-diffusion diagnostic exports were being populated incorrectly or left as zero.

### The Problem

During the diffusion loop, variables are checked against their diffusion `TYPE` attribute (`'U'` for momentum, `'S'` for heat, `'Q'` for tracers) instead of their actual variable `NAME` (`'U'`, `'V'`, `'Q'`, etc.). 
The original code looked like this:
```fortran
! Fill exports of U,V and S after diffusion
      if( TYPE=='U' ) then
          if(associated(UAFDIFFUSE)) UAFDIFFUSE = SX
       endif
      if( TYPE=='V' ) then
          if(associated(VAFDIFFUSE)) VAFDIFFUSE = SX
       endif
       if( TYPE=='S' ) then 
          if(associated(SAFDIFFUSE)) SAFDIFFUSE = SX
       endif
       if( TYPE=='Q' ) then
          if(associated(QAFDIFFUSE)) QAFDIFFUSE = SX
       endif
```
Because `TYPE` is a general diffusion class rather than the specific variable name, it caused several major issues in the output:
1. **`VAFDIFFUSE` was always zero:** The variable `V` diffuses like momentum, so its `TYPE` is `'U'`. Therefore, `TYPE == 'V'` was never true, and `VAFDIFFUSE` was never populated.
2. **`UAFDIFFUSE` was overwritten by `V`:** Since both `U` and `V` have `TYPE == 'U'`, the loop would correctly populate `UAFDIFFUSE` when processing `U`, but then erroneously overwrite it with the `V` field when processing `V`.
3. **`QAFDIFFUSE` output was wrong:** Every tracer (e.g., aerosols, `QLLS`, `QILS`) has a `TYPE` of `'Q'`. Consequently, `QAFDIFFUSE` was continuously overwritten by every tracer in the loop, ultimately outputting the diffusion field of whichever tracer happened to be processed last, rather than specific humidity (`Q`).

### The Solution

The checks for filling the export diagnostics have been updated to use `trim(NAME)` instead of `TYPE` to ensure the correct variable is mapped to its respective diagnostic exactly once:
```fortran
! Fill exports of U,V and S after diffusion
      if( trim(NAME)=='U' ) then
          if(associated(UAFDIFFUSE)) UAFDIFFUSE = SX
       endif
      if( trim(NAME)=='V' ) then
          if(associated(VAFDIFFUSE)) VAFDIFFUSE = SX
       endif
       if( trim(NAME)=='S' ) then 
          if(associated(SAFDIFFUSE)) SAFDIFFUSE = SX
       endif
       if( trim(NAME)=='Q' ) then
          if(associated(QAFDIFFUSE)) QAFDIFFUSE = SX
       endif
```

### Physics Impact (Non-Zero Diff)

This bug had massive implications for the model physics because these exported `AFDIFFUSE` variables are critical inputs passed from the first turbulence stage to both the Surface Component (SURF) and the second stage of the Turbulence Component (TURBL).
Tracing the `AFDIFFUSE` exports into `GEOS_PhysicsGridComp.F90` (specifically around line 2626 in the `SYNCTQ` section), here is exactly what was happening:

#### 1. Corrupted Zonal Wind, Meridional Wind, and Wind Speed at the Surface

Because `UAFDIFFUSE` was being erroneously overwritten by the diffused `V` field, and `VAFDIFFUSE` was left as `0.0`:
```fortran
! From GEOS_PhysicsGridComp.F90
UFORSURF = UAFDIFFUSE(:,:,LM)  ! <--- Was erroneously getting V!
VFORSURF = VAFDIFFUSE(:,:,LM)  ! <--- Was erroneously getting 0.0!
```
This means the **zonal wind (`U`) passed to the surface was actually the meridional wind (`V`)**, and the **meridional wind passed to the surface was exactly zero**.
Furthermore, this propagated directly into the calculation of the surface wind speed (`SPD4SURF`), which drives all surface stress and flux calculations:
```fortran
SPD4SURF = SQRT( UFORSURF*UFORSURF + VFORSURF*VFORSURF )
```
Because `VFORSURF` was 0 and `UFORSURF` was actually `V`, the surface wind speed was effectively evaluating to `SQRT(V^2 + 0) = |V|`. The surface physics was completely ignoring the zonal wind speed!

#### 2. Corrupted Specific Humidity at the Surface

Because `QAFDIFFUSE` was continuously overwritten by every tracer (since they all share `TYPE == 'Q'`), the value passed to the surface for specific humidity was actually the diffused field of whichever tracer happened to be processed last in the loop (likely an aerosol or chemical species, which usually have values orders of magnitude smaller than water vapor).
```fortran
QFORSURF = QAFDIFFUSE(:,:,LM) ! <--- Was getting the last tracer in the bundle instead of specific humidity (QV)
```
This means the surface component was computing evaporation and latent heat fluxes using completely invalid near-surface moisture values. 

**Summary:** This fix is very much non-zero-diff. It restores the correct zonal wind, meridional wind, scalar wind speed, and specific humidity that get passed from the first turbulence stage into both the Surface component and the second Turbulence stage.


narnold1:
It looks like this was introduced about 3.5 years ago with the refactoring.  7857ea7

This would primarily impact the RUN2 methods, as RUN1 surf uses the AFMOIST variables.  Based on a comparison of Jason-4 and v11.8 AMIPs (before/after the buggy commit), climatological surface fluxes and surface layer variables seem to be minimally affected.

narnold1 (approved):

sdrabenh (approved):

PR #1377, opened 2026-04-22
Fix cdo sourcing on GCMv11 Topo develop

### LINKED ISSUE #1368: Topo package status
Current summary of issue I see: 

1. From October 14th to now g5_modules changed and now cdo module if loaded is not compatible with what cdo library says it is, so code fails on develop GCMv11. 

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/786f712f2f6333c21e2b3cfbbbfb19b4929d2d90/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/preproc/topography/utils_topo/make_topo.py#L106-L108

Gives this error: 
`cdo: /usr/local/other/gcc/12.3.0/lib64/libstdc++.so.6: version` `GLIBCXX_3.4.32' not found (required by cdo)`
`cdo: /usr/local/other/gcc/12.3.0/lib64/libstdc++.so.6: version` `GLIBCXX_3.4.32' not found (required by /usr/local/other/cdo/2.5.4/deps/lib64/libeccodes.so)`

**If I change modules to this:** 
`source bin/g5_modules.sh` ( this I need so I can have questionnaire)
`module purge` ( now I purge all and just add ones I need for run)
`module load comp/intel/2024.2.0`
`module load mpi/impi/2021.13`
`module load nco`
`module load cdo`
`module load python/GEOSpyD/25.3.1-0/3.13`

**Now develop runs and all is good.** 

BUT this brings us to second issue ...

2. @sdrabenh - During testing even though all connecting PRs say we need to merge Topo PR# 5 with others, we missed this one to merge. 

https://github.com/GEOS-ESM/Topo/pull/5

| Repo              | Original        | Current                                   |
|-------------------|-----------------|-------------------------------------------|
| GEOSgcm_GridComp  | (t) v2.8.1 (DH) | (b) develop                               |
| Topo              | (t) geos/v2.1.0 (DH) | (b) feature/borescan_cleanup_and_documentation |

**With this branch what we produced under official Topo BCs ` */bcs_shared/make_bcs_inputs/atmosphere/TOPO/v2/*` is zero diff for both regular and stretched grid.** 

3. Now we come to pi PR ... best for last. 

@gmao-rreichle Testing shows that introducing the PI cleanup in [PR #1364](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1364/changes#top) changes the result: with the PR, the code fails; without it, the code passes. This suggests the topo preprocessing geometry is highly sensitive to very small numerical perturbations. In particular, the updated PI value appears to expose an existing weakness in the corner-ordering/orientation logic for very small or nearly degenerate grid polygons, rather than introducing a fundamentally new algorithmic error. The PR is cleaner and more consistent numerically, but it likely requires additional safeguards so that small changes in constants do not flip polygon ordering or alter downstream behavior.

| Repo             | Original             | Current                                      |
|------------------|----------------------|----------------------------------------------|
| GEOSgcm_GridComp | (t) v2.8.1 (DH)      | (b) bugfix/rreichle/value_of_pi_in_topo_package |
| Topo             | (t) geos/v2.1.0 (DH) | (b) feature/borescan_cleanup_and_documentation |

**My questions for all are:**  
1-> Do I make these module changes on top of pi PR? Or do we want this to be fixed for develop in g5_modules? 

2-> Do we still want to proceed with pi changes, if yes,  should I add code changes on top of current pi PR and topo# 5? 

3-> PR #1364 is a reasonable cleanup, but it also highlights that this part of the topo preprocessing is numerically sensitive. If we standardize the value of `PI` in `GEOSgcm_GridComp`, we should make sure the `Topo` side uses the same convention as well (https://github.com/search?q=repo%3AGEOS-ESM%2FTopo+3.141592653589793&type=code ) ; otherwise, we risk introducing cross-repository inconsistencies in geometry calculations. The current behavior suggests that small numerical differences can change branch decisions in corner ordering or orientation for very small or nearly degenerate polygons. So the right long-term fix is probably not to avoid the cleanup, but to make the constants consistent across the code base and add safeguards where the geometry logic is threshold sensitive. Downside it takes time to do and redo testing.

biljanaorescanin:
Resolving last part of #1368 issue for topo, bullet one.

Using `baselibs cdo` resolves issue of library compatibility that started happening after sometime in January. 

This PR only affects topography preprocessing so it is trivially zero diff for testing.


biljanaorescanin:
@mathomp4 @weiyuan-jiang looks like this fixed the issue. 
I've did sanity test, one of each grids and all looked fine and zero diff. Thanks!

gmao-rreichle (approved):

sdrabenh (approved):

PR #1393, opened 2026-05-06
[DSL] remove unused constant `N_MODES`

romanc:
The constant and the associated field type are not used (anymore) in pyMoist.


FlorianDeconinck:
`0-diff` because this a DSL-only change, deactivated by default, no changes to the Fortran

FlorianDeconinck (approved):
`N_MODES` is a tied to the Aerosol Activation, which did port but not move to the 11.8.1

Fine for removal - when we do carry the AerActivation port over we will want to use `DataDimensionFields` to cover.

sdrabenh (approved):

PR #1394, opened 2026-05-06
[DSL] pyMoist update to v11.8.1 (+ pre-v11.10.1)

romanc:
## Description

- Added `docs`
- Make some translate test GPU ready
- Local optimziation pass
- GPU-ready timers

- UW shallow convection parameterization updated to v11.10.1, scientifically validated (not quite numerically identical to Fortran, but incredibly close.
- GFDL1M microphysics and GF2020 convection parameterizations are not yet updated. v11.10 brought significant changes to these schemes, so their updates will occur in separate (subsequent) PRs. Warnings have been added to notify users that they are running outdated code (which will likely crash as Fortran renamed numerous fields and constants in v11.10).

PR #1396, opened 2026-05-07
[DSL] ci: run ndsl-checks on PRs

romanc:
## Description

This PR updates the GitHub action workflow `ndsl-check.yml`. The idea of this workflow is that it runs conditionally on PRs upon changes to `pyMoist`. This PR makes sure that the workflow actually runs on PRs and doesn't fail "somewhere in the middle". In more details, the changes include

- change workflow to run on pull requests
- change paths to include changes to the workflow
- change python version to 3.12 (as on the HPC)
- fix `mkdocs build`
- fix data url
- make wget less verbose
- fix `pytest` invocation
- add NDSL to dependencies
- non-verbose `tar` extract
- pick a (random) experiment (why is this needed?!)
- fix `test_constants` by adding more excludes
- fix running translate test by temporarily exluding netCDFs that don't match a translate test anymore.

I took a rather pragmatic approach to "just make it run through". While this now runs through, I have two main questions

1. Are we sure about the data for the translate tests? I had to manually delete a bunch of netCDFs (the translate test system searches for translate tests based on the (name of the) netCDFs) to make it at least run. And even then, many that run actually fail. PR https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1394/ will help with the microphysics PRs that complain about missing arguments.
2. Can we have a default for the `EXP_NAME` environment variable? Since it's checked in `constants.py` (and everything imports `constants.py`), whatever attempts to run code needs this defined. For example, the step "Run regular unit tests" needs to define the environment variable, even if it only runs `test_constants.py`, which doesn't need to know how many tracers are defined.

I'm leaving this in a draft state until we decide on 1 and 2. 

Marked as `0-diff` because we don't change the Fortran in the PR.

### DIFF
--- a/.github/workflows/ndsl-checks.yml
+++ b/.github/workflows/ndsl-checks.yml
@@ -1,8 +1,10 @@
 name: NDSL Lint & Translate Tests
 
 on:
-  push:
-    paths: "./GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/pyMoist/**"
+  pull_request:
+    paths:
+      - '.github/workflows/ndsl-checks.yml'
+      - 'GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSmoist_GridComp/pyMoist/**'
   workflow_dispatch:
 
 # cancel running jobs if theres a newer push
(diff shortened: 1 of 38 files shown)


romanc:
closing as superseded by https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1431 and/or https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1394.

PR #1400, opened 2026-05-07
add TOPO_version_info to BCs package and route dir

biljanaorescanin:
Here we record in file what we decided BCs will use for topo version


biljanaorescanin:
@gmao-rreichle  Both fix for route dir creation, and new added file info were tested and works as we intended. 

I've run v14 BCs for EASE M03 and CF180.

gmao-rreichle (approved):

sdrabenh (approved):

PR #1405, opened 2026-05-11
Replace MAPL_DecomposeDim with mapl_GetPartition in RouteGridComp

tclune:
## Summary

- Replace legacy `MAPL_DecomposeDim` with v3 `mapl_GetPartition` in `GEOS_RouteGridComp.F90`
- Add `use mapl_Partition, only: mapl_GetPartition` to the module use list

## Notes

- This was the only call site of `MAPL_DecomposeDim` in GEOSgcm, enabling its removal in a follow-on MAPL PR.
- `mapl_GetPartition` returns a 1-based integer array of local sizes, fully compatible with the existing `ims(mype+1)` indexing.

Closes GEOS-ESM/GEOSgcm#1080

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOSroute_GridComp/GEOS_RouteGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOSroute_GridComp/GEOS_RouteGridComp.F90
@@ -25,6 +25,7 @@ module GEOS_RouteGridCompMod
 
   use ESMF
   use MAPL_Mod
+  use mapl_Partition, only: mapl_GetPartition
   use MAPL_Constants
   use ROUTING_MODEL,          ONLY:     &
        river_routing, ROUTE_DT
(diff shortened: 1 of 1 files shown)


gmao-rreichle:
Thanks, @tclune, for cleaning this up.  How confident are you that this will work?  I'm asking because the Routing GC is brand new, and we don't yet have a 0-diff test for it.  From looking at the changes, perhaps this could be considered trivially 0-diff even for the Routing GridComp (provided the in-progress build succeeds)?  

The change is trivially 0-diff for the GCM because Routing is so far only working in GEOSldas.

I see that the change is going into the MAPL-v3 branch, but it's probably better to merge it into develop.  Or is mapl_GetPartition() not yet available in current MAPL releases? 

To reduce the PR count for @sdrabenh, and assuming mapl_GetPartition() is available now, we could also add your change into @zyj8881357's forthcoming PR that will enable the Routing GC in the GCM.  Arguably, that would not be the fastest way to get rid off the MAPL_DecomposeDim() call.  It's difficult to predict how long it will take for @zyj8881357's forthcoming PR to be ready for merging.  A wild guess is probably 1-2 months.

@biljanaorescanin, @zyj8881357 : We should work on expanding the LDAS_GLOBAL/model test to include routing.  The config of the new test should be identical to what we have now, except that Routing is enabled and a "route" HISTORY collection file is written out. 

cc: @biljanaorescanin @zyj8881357

tclune:
@gmao-rreichle    I'm fairly confident it will work, but was hesitant about making this PR simply because this code is not even compiled on the mapl v3 branch for now.   But as it was the only external caller to a routine that I'm deleting, it seemed better to indicate the change in the v3 branch.   If you want to test the same change in your own branches, that will be great.  But in the grand scheme these are not  the things to worry about just now.

PR #1440, opened 2026-05-29
Make consistency btw default values of SS_FOUND in OpenWater and SeaiceInterface

sinakhani:
## Notes:
- Default values of SS_FOUND were different in ImportSpec from OpenWater and SeaiceInterface
- This makes inconsistency in MAPLE initialDefault
- This PR fixes this inconsistency

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSsaltwater_GridComp/GEOS_OpenWaterGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSsaltwater_GridComp/GEOS_OpenWaterGridComp.F90
@@ -1298,7 +1298,7 @@ module GEOS_OpenwaterGridCompMod
           UNITS              = 'PSU',                             &
           DIMS               = MAPL_DimsTileOnly,                 &
           VLOCATION          = MAPL_VLocationNone,                &
-          DEFAULT            = 30.0,                              &
+          DEFAULT            = 33.3333,                           &  !SK - Match the SS_FOUND in OpenWater and SeaiceInterface
           _RC)
 
      call MAPL_AddImportSpec(GC,                                  &
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSsaltwater_GridComp/GEOS_SimpleSeaiceGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSsaltwater_GridComp/GEOS_SimpleSeaiceGridComp.F90
@@ -1071,7 +1071,7 @@ module GEOS_SimpleSeaiceGridCompMod
         UNITS              = 'psu',                               &
         DIMS               = MAPL_DimsTileOnly,                   &
         VLOCATION          = MAPL_VLocationNone,                  &
-        DEFAULT            = 30.0,                                &
+        DEFAULT            = 33.3333,                             &     !SK - Match the SS_FOUND inSimpleSeaice with that from OpenWater and SeaiceInterface
         _RC  )
 
    call MAPL_AddImportSpec(GC,                                  &


zhaobin74 (dismissed):
Thanks for the fix. Just a note here:  OpenWater does not really use SS_FOUND other than exporting it.

mathomp4:
@zhaobin74 @sinakhani So this doesn't affect AMIP at all? And would it affect my nightly c12 MOM6 runs?

sinakhani:
I think it probably will not affect AMIP. MAPLE initializes SS_FOUND with a default value. Currently, this default value is different for OpenWater and SeaiceInterface GridComps (one is 30. the other 33.3333). With this, an inconsistency in ImportSpec occurs that causes geos to die with an error in MAPLE_generic. The PR just sets SS_FOUND with the same default value in ImportSpec.

zhaobin74:
> @zhaobin74 @sinakhani So this doesn't affect AMIP at all? And would it affect my nightly c12 MOM6 runs?

@mathomp4, it will only affect runs bootstrapping `saltwater_import_rst`. If your runs always have this restart, there is no issue.

mathomp4:
> > @zhaobin74 @sinakhani So this doesn't affect AMIP at all? And would it affect my nightly c12 MOM6 runs?
> 
> @mathomp4, it will only affect runs bootstrapping `saltwater_import_rst`. If your runs always have this restart, there is no issue.

Hmm. So it *might*. I do actually bootstrap `saltwater_import_rst` when I do my test runs. I only keep the 8 required restarts. So this might be...kind of non-zero-diff? I'll try and test tomorrow.

zhaobin74:
> > > @zhaobin74 @sinakhani So this doesn't affect AMIP at all? And would it affect my nightly c12 MOM6 runs?
> > 
> > 
> > @mathomp4, it will only affect runs bootstrapping `saltwater_import_rst`. If your runs always have this restart, there is no issue.
> 
> Hmm. So it _might_. I do actually bootstrap `saltwater_import_rst` when I do my test runs. I only keep the 8 required restarts. So this might be...kind of non-zero-diff? I'll try and test tomorrow.

Then it should be added as a required restart when running with CICE6. If it is not available for a particular resolution, it can be regridded.

mathomp4:
> Then it should be added as a required restart when running with CICE6. If it is not available for a particular resolution, it can be regridded.

Ahh. I do see I have a `saltwater_internal_rst` in my MOM6 c12 testing restarts! Okay. 

Now, @sinakhani thinks this PR is needed for the CI failure I see here:

https://github.com/GEOS-ESM/GEOS_Util/pull/234

I'm hoping to test this on discover "soon" but yeesh, it does not want me to have nodes!

mathomp4:
@sinakhani @zhaobin74 Should you also change the value in SimpleSeaIce as well?

In there I see:

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/028bbb33165c1c8ec8ea3864ed1d5a67a1ab652a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSsaltwater_GridComp/GEOS_SimpleSeaiceGridComp.F90#L1068-L1075

zhaobin74:
> @sinakhani @zhaobin74 Should you also change the value in SimpleSeaIce as well?
> 
> In there I see:
> 
> https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/028bbb33165c1c8ec8ea3864ed1d5a67a1ab652a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSsaltwater_GridComp/GEOS_SimpleSeaiceGridComp.F90#L1068-L1075

Maybe change SeaInceInterface value to 30, to be consistent with Openwater and SimpleSeaice? The value does not matter at all. Just don't want to trigger these failures.

BTW, `saltwater_import_rst` is part of the suite of restarts that can be handled by remap_restart utility. There is no reason anyone needs to bootstrap it.

mathomp4:
Can confirm with the update to simple seaice, it runs and is zero-diff! Yay!

I'll go re-run the CI in the GEOS_Util PR to double check

sinakhani:
I have updated my branch with a change to SimpleSeaice. I think I missed @zhaobin74 's suggestion to only change the other one -- Sorry!

zhaobin74:
> I have updated my branch with a change to SimpleSeaice. I think I missed @zhaobin74 's suggestion to only change the other one -- Sorry!

@sinakhani, that's ok. Either way works to stop the CI error. Thanks.

mathomp4:
> > I have updated my branch with a change to SimpleSeaice. I think I missed @zhaobin74 's suggestion to only change the other one -- Sorry!
> 
> @sinakhani, that's ok. Either way works to stop the CI error. Thanks.

Wasn't just a CI issue. It broke dataice (or whatever we call it). But now it's all happy! Huzzah! 🎉 

Thanks @sinakhani

sdrabenh (approved):

mathomp4 (approved):

PR #1445, opened 2026-06-04
bugfix: Replace non-standard fstat with standard Fortran INQUIRE in mkMITAquaRaster.F90

mathomp4:
GCC 16 tightened enforcement of the `fstat` intrinsic signature, requiring a 13-element array but the code declared only 12, causing a compile error:

```
Error: Size of 'values' argument of 'fstat' intrinsic at (1) too small (12/13)
```

`fstat` is a non-standard GNU extension. This fix replaces both calls with the standard Fortran 2003 `INQUIRE(file=..., size=...)` intrinsic, which directly returns the file size without needing an intermediate status array. This is portable across compilers and was already applied in the Flang fixes in #1175.

Additional cleanup:
- `integer(REAL64)` → `integer(INT64)` for `filesize` and `Length` — using `REAL64` as a kind parameter for an integer variable was technically accidental (both happen to be kind=8 on most platforms); `INT64` is semantically correct.
- `STATARRAY(12)` declaration removed as it is no longer needed.
- `INT64` added to the `iso_fortran_env` import.

All my testing shows this is zero-diff but then, the model doesn't run this code. This is part of the makebcs code...

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkMITAquaRaster.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/mkMITAquaRaster.F90
@@ -5,7 +5,7 @@
 
       use LogRectRasterizeMod,     ONLY: LRRasterize
       use MAPL_ExceptionHandling
-      use, intrinsic :: iso_fortran_env, only: REAL64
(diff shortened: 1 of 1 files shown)


sdrabenh (approved):

mathomp4:
This is very weird. The CI keeps trying to checkout this branch in MAPL:

```
Checking out branch bugfix/fixes-for-gcc16 in cmake
Checking out branch bugfix/fixes-for-gcc16 in MAPL
```
But that branch does not exist. At all.

PR #1449, opened 2026-06-08
Bugfix: stabilize GEOSldas offline route restart comparisons

biljanaorescanin:
This PR addresses an intermittent **route-only restart comparison failure** in the LDAS nightly tests, where small `WSTREAM`/`WRIVER` differences/fails moved between **Intel** `GLOBAL` and **GNU** `GNUGLOBAL`.

### GEOSldas Nightly test behavior before fix

| Date | Configuration | Result vs current baseline | Notes |
|---|---|---:|---|
| 2026-06-07 | `GNUGLOBAL` | `pass/FAIL` | Route-only comparison failure|
| 2026-06-06 | all tests | `pass` | No route comparison failure reported |
| 2026-06-05 | `GLOBAL` & `GNUGLOBAL` | `pass/FAIL` | Route-only comparison failure|
| 2026-06-04 | `GLOBAL` & `GNUGLOBAL`| `pass/FAIL` | Route-only comparison failure|
| 2026-06-03 | `GLOBAL` & `GNUGLOBAL` | `pass/FAIL` | Route-only comparison failure |
| 2026-06-02 | `GNUGLOBAL` | `pass/FAIL` | Route-only comparison failure|
| 2026-05-31 | `GLOBAL` | `pass/FAIL` | Route-only comparison failure |

Diagnostics ruled out `RIVER_ROUTING_HYD` itself as the source of the run to run instability when `QRUNOFF` was forced to deterministic values. The remaining sensitivity pointed to the tile to Pfafstetter remap used to compute `QRUNOFF`, especially in the EASE/Pfaf route path.

The fix promotes the route remap source/destination fields and runtime remap pointer from `R4` to `R8` before casting the mapped runoff back to the route array. This stabilizes the EASE/Pfaf runoff remap and removed the same-executable `route_internal_rst` differences in repeated Intel and GNU LDAS route tests, including the previously sensitive `WSTREAM` and `WRIVER` fields. 

`This change is zero-diff for GCM configurations, since routing is currently used only in offline GEOSldas mode.`

**Validation:** 
The full regression framework still reports model comparison differences against the current baseline because the baseline does not yet include this route remap change and will need to be rebased. However, standalone Intel and GNU LDAS route reruns were zero-diff for `route_internal_rst` when compared with the corresponding regression run outputs generated from this PR. Also full Regression package was run 2 times to confirm is reproduces itself. 

@gmao-rreichle @weiyuan-jiang  @mathomp4 @zyj8881357

One question remains for all, we implemented change to [FOPT2](https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/develop/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSroute_GridComp/CMakeLists.txt) for Route but if this PR is the fix we needed I am not sure is that change still necessary...?


zyj8881357:
I have checked both route outputs and RS. The results look good for me.

mathomp4:
> One question remains for all, we implemented change to [FOPT2](https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/develop/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSroute_GridComp/CMakeLists.txt?rgh-link-date=2026-06-08T14%3A42%3A24Z) for Route but if this PR is the fix we needed I am not sure is that change still necessary...?

@biljanaorescanin Probably? I don't know. For now, I'd say get your current fix in. Then, when @weiyuan-jiang is back from leave, he can take a look to see if FOPT2 is still needed. But I mean, working code is better than perfect code to start (and in the end, it's not like FOPT3 was making the code 10x faster or something)

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSroute_GridComp/GEOS_RouteGridComp.F90
```diff
@@ -947,7 +947,7 @@ subroutine RUN (GC,IMPORT, EXPORT, CLOCK, RC )
 
     integer                            :: nt_global, nt_local
 
-    real,                  pointer     :: arrayPtr(:)
+    real(kind=8),          pointer     :: arrayPtr8(:)
```
mathomp4: @biljanaorescanin Could you change from `kind=8` to `kind=REAL64`? `kind=8` isn't as portable (NAG for instance uses `kind=2` for 64-bit reals).

You'll need to add:
```fortran
use, intrinsic :: iso_fortran_env, only: REAL64
```
at the top.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSroute_GridComp/GEOS_RouteGridComp.F90
```diff
@@ -1028,21 +1028,36 @@ subroutine RUN (GC,IMPORT, EXPORT, CLOCK, RC )
...
-       ! redistribute runoff from tile space of GEOS_LandGridComp to Pfafstetter catchment space of GEOS_RouteGridComp
-       call ESMF_FieldGet(route%field_src, farrayPtr=arrayPtr, rc=status)
+       ! Clear destination route/Pfaf field before remapping.
+       call ESMF_FieldGet(route%field, farrayPtr=arrayPtr8, rc=status)
        VERIFY_(STATUS)
-       ArrayPtr = route%runoff_acc(:)
-       call ESMF_FieldSMM(srcField=route%field_src, dstField=route%Field, &
-            routeHandle=route%routeHandle, rc=rc)
-       call ESMF_FieldGet(route%field, farrayPtr=arrayPtr, rc=status)
+       arrayPtr8 = 0.0_8
```
mathomp4: This should be:
```fortran
       arrayPtr8 = 0.0_REAL64
```
as well.

biljanaorescanin: I pushed the update if you wanna check but I think I changed all. I will test again once you confirm it looks ok

mathomp4 (approved):

gmao-rreichle (approved):

sdrabenh (approved):

PR #1457, opened 2026-06-17
Hotfix for M21C: Read and apply climatological precipitation rescaling factors

gmao-qliu:
REPLACED BY PR #1468 (MERGED)

---

The modification implements Rolf's suggestion as follows. @gmao-rreichle @mathomp4 

"The M21C corrected precip (and thus the land variables) are bad in the Tropics (20S-20N) after the switch from IMERG-Final to IMERG-Late on 1 Oct 2025, and the problem is with the IMERG-Late data. 

The best fix that I can think of right now is to add a new option in the GCM executable.  This new option rescales the model precip to the IMERG-Final climatology using seasonally varying scaling parameters (as opposed to replacing it with the precip from the corrected precip files).  This option requires new (scaling) files and a new DAS tag, which is a tall order.  But the fix addresses the two most important things: (1) It removes IMERG-Late from the system entirely and (2) avoids a step change in the M21C soil moisture, ET, etc climatology in Oct 2025."

**Update on AGCM Testing:** @sdrabenh @gmao-rreichle 
AGCM testing was completed using the current m21c precipitation setting versus the new climatological factor setting.

Current (PRECIP_FILE):
`ExtData/PCP/precip_corr_R21C/diag/Y%y4/M%m2/R21C.tavg_1hr_prcpcorr_c360_sfc.%y4%m2%d2_%h230.nc4`
New (PRECIP_FILE_CLIMSCALE): 
`/home/qliu/pclim_scale/M21CtoIMGv7BF.precip_clim_scale.%y4%m2%d2.nc4`

Beginning with data-day 1 Oct 2025 (after the end of IMERG-Final data), the M21C config needs to be changed.  The rc variable PRECIP_FILE needs to be commented out in the config.  A new rc variable PRECIP_FILE_CLIMSCALE needs to be added that points to the new "precip_clim_scale" files.

**Testing Results:**
When using the IMERG-Final based corrections ("PRECIP_FILE") in the AGCM, the output PREC[xxx] and PREC[xxx]CORR variables show zero differences (0-diff) when compared against the GEOSgcm R21 branch. 
When using the climatological rescaling ("PRECIP_FILE_CLIMSCALE") in the AGCM, the new setting works as intended.

**Additional Testing To Be Done:**
The 0-diff test for PRECIP_FILE should be done with the M21C ADAS.
The new PRECIP_FILE_CLIMSCALE functionality needs to be verified with the M21C ADAS

### DIFF
--- a/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
+++ b/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
@@ -5136,8 +5136,8 @@ module GEOS_SurfaceGridCompMod
     type (ESMF_Field)                   :: Field
     type (ESMF_Grid)                    :: GRID
     type (ESMF_Time)                    :: CurrentTime
-    character(len=ESMF_MAXSTR)          :: PRECIP_FILE
-    character(len=ESMF_MAXSTR)          :: PRECIP_SCALE_FILE
+    character(len=ESMF_MAXPATHLEN)      :: PRECIP_FILE
+    character(len=ESMF_MAXPATHLEN)      :: PRECIP_FILE_CLIMSCALE
 
     type (T_SURFACE_STATE), pointer     :: surf_internal_state 
     type (SURF_wrap)                    :: wrap
@@ -6230,7 +6230,7 @@ module GEOS_SurfaceGridCompMod
 
     ! for now, do not allow the combination of precip replacement and clim rescaling
     
-    _ASSERT( .not. (PRECIP_FILE /= "null" .and. PRECIP_FILE_CLIMSCALE /= "null"), &
+    _ASSERT( .not. (trim(PRECIP_FILE) /= "null" .and. trim(PRECIP_FILE_CLIMSCALE) /= "null"), &
          "only one of PRECIP_FILE *or* PRECIP_FILE_CLIMSCALE can be set" )
     
 ! These exports are the rainfalls and total snowfall that
@@ -6265,7 +6265,7 @@ module GEOS_SurfaceGridCompMod
     ICE = ICEFL
     FRZR= FRZRFL
 
-    REPLACE_PRECIP: if(PRECIP_FILE /= "null" .or. PRECIP_FILE_CLIMSCALE /= "null") then
+    REPLACE_PRECIP: if(trim(PRECIP_FILE) /= "null" .or. trim(PRECIP_FILE_CLIMSCALE) /= "null") then
 
        bundle = ESMF_FieldBundleCreate (NAME='PRECIP', RC=STATUS)
        VERIFY_(STATUS)
@@ -6287,7 +6287,7 @@ module GEOS_SurfaceGridCompMod
                           RC=STATUS )
        VERIFY_(STATUS)
        
-       if(PRECIP_FILE /= "null") then
+       if( trim(PRECIP_FILE) /= "null") then
           
           ! read corrected precip (PTTe) directly from (hourly) file if file exists (crashes otherwise?)
           
@@ -7431,7 +7431,7 @@ module GEOS_SurfaceGridCompMod
        VERIFY_(STATUS)
 
        ! Do not correct the precip over ocean tiles
-       if(Precip_File /= "null") then
+       if( trim(Precip_File) /= "null"  .or. trim(PRECIP_FILE_CLIMSCALE) /= "null" ) then
 
           call MAPL_LocStreamTransform( LOCSTREAM, TMPTILE  , PCU,     RC=STATUS); VERIFY_(STATUS)
           where(tiletype == MAPL_OCEAN)  PCUTILE = TMPTILE
@@ -7451,7 +7451,7 @@ module GEOS_SurfaceGridCompMod
        end if
 
        ! Adjust the discharge going to the ocean
-       if(Precip_File /= "null" .and. DischargeAdjustFile /= "null") then
+       if( (trim(Precip_File) /= "null"  .or. trim(PRECIP_FILE_CLIMSCALE) /= "null") .and. DischargeAdjustFile /= "null") then
 
           call MAPL_GetPointer(INTERNAL, DISCHARGE_ADJUST, 'DISCHARGE_ADJUST',  RC=STATUS)
           VERIFY_(STATUS)
@@ -7490,7 +7490,7 @@ module GEOS_SurfaceGridCompMod
     call MAPL_GetPointer(EXPORT, CN_PRCP, 'CN_PRCP', ALLOC=.true., RC=STATUS)
     VERIFY_(STATUS)
 
-    if(PRECIP_FILE /= "null") then
+    if( trim(PRECIP_FILE) /= "null" .or. trim(PRECIP_FILE_CLIMSCALE) /= "null" ) then
        TMPTILE = PCUTILE
        call MAPL_LocStreamTransform( LOCSTREAM, CN_PRCP, TMPTILE, RC=STATUS)
        VERIFY_(STATUS)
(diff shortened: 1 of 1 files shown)


mathomp4:
@gmao-qliu I updated the base of your PR to `R21C`. It was going to `develop` and was saying you had a zillion changes!

mathomp4:
@gmao-qliu I see one possible issue. You added:
```fortran
! Read in precip scale data as a second option to get "corrected" precip 
!------------------------------------------------------
    
    call MAPL_GetResource(MAPL,PRECIP_SCALE_FILE,LABEL="PRECIP_SCALE_FILE:",default="null", RC=STATUS)
    VERIFY_(STATUS)

    call ESMF_ClockGet(CLOCK, currTime=CurrentTime, rc=STATUS)     
    VERIFY_(STATUS)
    call ESMF_TimeGet (currentTime,               &                
                       YY=YEAR, MM=MONTH, DD=DAY, &
                       H=HR,    M=MN,     S=SE,   &
                                        RC=STATUS ) 
    VERIFY_(STATUS)  
    ! daily climatology, only need month/day information                   
    call ESMF_TimeSet (currentTime,               &
                       YY=8888, MM=MONTH, DD=DAY, & 
                       H=0,    M =0,    S = 0,  & 
                                        RC=STATUS ) 
    VERIFY_(STATUS)
```

In this you are "reusing" the `currentTime` variable that the old code was using:
```fortran
! Read in precip data. This is used in 'coupled' replay
!------------------------------------------------------

    call MAPL_GetResource(MAPL,PRECIP_FILE,LABEL="PRECIP_FILE:",default="null", RC=STATUS)
    VERIFY_(STATUS)

    call ESMF_ClockGet(CLOCK, currTime=CurrentTime, rc=STATUS)
    VERIFY_(STATUS)
    call ESMF_TimeGet (currentTime,               &
                       YY=YEAR, MM=MONTH, DD=DAY, &
                       H=HR,    M=MN,     S=SE,   &
                                        RC=STATUS )
    VERIFY_(STATUS)
    call ESMF_TimeSet (currentTime,               &
                       YY=YEAR, MM=MONTH, DD=DAY, &
                       H=HR,    M =30,    S = 0,  &
                                        RC=STATUS )
    VERIFY_(STATUS)
```

For safety's sake, you might want to use a different `ESMF_Time` variable for the scale case.

Or I suppose put if-blocks around the `ESMF_TimeSet` call where "if old, do HH:30:00, if new use 8888-MM-DD 00:00:00"

bena-nasa:
Other than the issues @mathomp4 raised this looks good to me

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
```diff
@@ -6263,20 +6284,38 @@ subroutine RUN2 ( GC, IMPORT, EXPORT, CLOCK, RC )
 !  a "corrected" precip according to Rolf et al. This was used in MERRA-2 and
 !  would typically be done one during reanalysis or replay. Also, OGCM-coupled
 !  replays require special treatment.
+! Option to read a scale factor and compute the "corrected" precipiation to 
+! address IMERG-Late v7B quality issues.
 !-----------------------------------------------------------------------------
 
-    REPLACE_PRECIP: if(PRECIP_FILE /= "null") then
+    REPLACE_PRECIP: if(PRECIP_FILE /= "null" or PRECIP_FILE /= "null") then
```
mathomp4: I'm guessing this should be referring to `PRECIP_SCALE_FILE` in the new code?

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
```diff
@@ -6263,20 +6284,38 @@ subroutine RUN2 ( GC, IMPORT, EXPORT, CLOCK, RC )
...
 
+       PRECSUM = RCU+RLS+SNO+ICE+FRZR
+
+       if(PRECIP_FILE /= "null") then
+          ! read corrected precip (PTTe) directly from file if file exists
+          call MAPL_read_bundle( Bundle,PRECIP_FILE, CurrentTime, regrid_method=REGRID_METHOD_CONSERVE, RC=status)
+          VERIFY_(STATUS)
+          call ESMFL_BundleGetPointerToData(Bundle,'PRECTOT',PTTe, RC=STATUS)
+          VERIFY_(STATUS)
+       else
```
mathomp4, line 6305: This might need to be looked at. You have if/else, but what if a user has *neither* a `PRECIP_FILE` or a `PRECIP_SCALE_FILE`. 

As it is now, the else block will run if even if `PRECIP_SCALE_FILE` is `null`

mathomp4: And, well, `MAPL_read_bundle` on `null` will probably be very unhappy. (cc @bena-nasa )

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
gmao-qliu: correct

gmao-qliu:
> @gmao-qliu I see one possible issue. You added:

> In this you are "reusing" the `currentTime` variable that the old code was using:

> For safety's sake, you might want to use a different `ESMF_Time` variable for the scale case.
> 
> Or I suppose put if-blocks around the `ESMF_TimeSet` call where "if old, do HH:30:00, if new use 8888-MM-DD 00:00:00"

Makes sense. please modify.

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
gmao-qliu: if neither, the whole block will skip according to the outer if condition ...

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
gmao-rreichle: I think this comment is obsolete after my latest commit

GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOS_SurfaceGridComp.F90
gmao-rreichle: I believe Qing is right.  The "else" condition necessarily implies that PRECIP_FILE_CLIMSCALE has a non-null value.

gmao-rreichle (commented):
@gmao-qliu @mathomp4 @bena-nasa : I tried to fix the issues pointed out in the comments and also clean up the logic a little bit more.  It's late in the day and I might well failed to address the existing issues or created new problems.  Please double-check.
Here's my commit: de6a76a4307c0e6413e2625cc248215d8a0adf26
But it's probably easier to just look at the "Files Changed" tab as it is now.
Thanks! 

cc: @elakkraoui

mathomp4:
@gmao-rreichle I think it looks good. 

One thing I noticed was:
```fortran
    character(len=ESMF_MAXSTR)          :: PRECIP_FILE
    character(len=ESMF_MAXSTR)          :: PRECIP_SCALE_FILE
```

I might change these to `ESMF_MAXPATHLEN`, which is 1024, over `ESMF_MAXSTR` which is 256. With a deep enough directory tree sometimes, you can reach 256 pretty fast. Both @bena-nasa and myself have hit this at times and it is annoying to debug.

My only other thought was in places where you do like:
```fortran
PRECIP_FILE /= "null"
```
I wonder if it might be better to do:
```fortran
trim(PRECIP_FILE) /= "null"
```

? I'm not sure it's necessary as it didn't have `trim()` before and was working. (@tclune is the one that really knows these when-to-trim-or-not realities of Fortran...)

gmao-rreichle:
Thanks, @mathomp4!  I added your suggestions and also fixed a few other things 05766558ef80f9083733cdccd4ced1bbb0510e4a.  The key changes are in Lines 7493 and 7506, otherwise the diagnostic output under the PRECIP_FILE_CLIMSCALE option would have been wrong, I think.  But I'm not sure, and it would be great if someone else can double-check. 
For M21C, I don't think "DISCHARGETILE" kicks in, but I added the changes nevertheless for completeness.

gmao-rreichle:
This PR was replaced by #1468 and is no longer needed, I think... Closing

cc: @mathomp4 @sdrabenh @gmao-qliu @elakkraoui

PR #1461, opened 2026-06-18
Ability to run coupled model at different dt

atrayano:
This add ability to couple the atmosphere and the ocean at different time intervals, and to run the ocean at different time-step from the global clock's heartbeat.

Note that this PR requires also the changes to the GEOS_OceanGridComp repository on the feature/oceanDt.rc31 branch. 

These changes are 0-diff against standard AMIP and coupled configurations

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ if (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/GEOS_GcmGridComp.F90)
 
  ecbuild_declare_project()
   esma_add_library(${this}
-    SRCS GEOS_GcmGridComp.F90
+    SRCS GEOS_GcmGridComp.F90 gc2gc.F90
     SUBCOMPONENTS ${alldirs}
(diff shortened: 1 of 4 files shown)


mathomp4:
@atrayano I think the GCC error is because you've split a macro across multiple lines:
```
[100%] Building Fortran object src/Components/@GEOSgcm_GridComp/CMakeFiles/GEOSgcm_GridComp.dir/gc2gc.F90.o
/__w/GEOSgcm_GridComp/GEOSgcm_GridComp/src/Components/@GEOSgcm_GridComp/gc2gc.F90:85:45:

   85 |        _ASSERT(tk==ESMF_TYPEKIND_R4, &
      |                                             1
Error: Syntax error in argument list at (1)
gmake[2]: *** [src/Components/@GEOSgcm_GridComp/CMakeFiles/GEOSgcm_GridComp.dir/build.make:88: src/Components/@GEOSgcm_GridComp/CMakeFiles/GEOSgcm_GridComp.dir/gc2gc.F90.o] Error 1

```

GNU doesn't like that. It has to be one long line.

PR #1471, opened 2026-07-09
Feature/donifan/rc31cpl add gf2 m

dbarahon:
Addded 2M microphysics to GF. PySR regressions instead of neural networks, and corrected bugs in hystpdf


sdrabenh:
I believe this PR is also outdated and will be followed up with a new one soon

PR #1203, opened 2026-02-05
Add ISSM ice sheet model GridComp to Landice

agstub:
This PR adds a new gridcomp that runs [ISSM](https://github.com/ISSMteam/ISSM), the Ice-sheet and Sea-level System Model. The ISSM gridcomp is a child of the Landice gridcomp.

- The branch works with the GEOS-ESM/ISSM tag [geos/v1.0.3](https://github.com/GEOS-ESM/ISSM/tree/geos/v1.0.3). Note: on NCCS the updated module is in [env v5.25.0](https://github.com/GEOS-ESM/ESMA_env/releases/tag/v5.25.0) 
- Aaron Stubblefield (@agstub) verified these [zero-diff tests](https://github.com/GEOS-ESM/GEOSgcm_GridComp/wiki/Zero-diff-changes) on 05/05/2026. 

**Overview of the new ISSM gridcomp:**
Imports: 
- ICESMB_ISSM (time-averaged ice surface mass "ICESMB" balance) from landice, described below. (Note this is imported through a private internal state because ISSM GridComp has a new mesh tile space)

Exports: 
- ICESURF (ice surface elevation)
- ICETHICK (ice thickness)
- ICEVEL (ice flow speed)
- ICEVX (ice velocity in local projection x-direction)
- ICEVY (ice velocity in local projection y-direction)

Internals:
- ALL EXPORTS (to persist across job segments in case ISSM does not run)
- IMLS (ice-mask level set --> ice front/terminus position)
- OMLS (ocean-mask level set --> grounding line position)
- ISSM_NSTEPS (number of landice timesteps since last ISSM run; see description below)
- RS_NODEIDS (global node IDs corresponding to ordering in the restart variables; see below)

Configuration parameters (in GEOS_SurfaceGridComp.rc):
- DO_ISSM: Turn on/off ISSM (default 0 / off)
- ISSM_DT: ISSM timestep (default semiweekly)

*Example mesh output (via 1d mesh tilespace):*
combinedres

*Example gridded output (via landice tile export --> grid):*

globalthick_grid

ISSM gridcomp main features:
1. The gridcomp will run any number of glaciers (binary input files) that are provided as GEOS BCs (see below for how to create BCs). The current test configuration runs Antarctica and Greenland, for example. 
[...]

PR #1476, opened 2026-07-13
Fix ISSM type

mathomp4:
In #1203, ISSM was added as a gridcomp. But in that code we have:

```fortran
type(ISSM_TILE_WRAP) :: issm_tile_wrap
```

While Intel seems to allow this, GNU (and Flang) does not like doing:
```fortran
type(foo) :: foo
```
where both the type and the variable have the same name. I think this is a violation of the Fortran Standards.

So this PR changes the type to be `T_ISSM_TILE_WRAP` to echo the `T_ISSM_TILE_STATE` type.

---

Note 1: Please look at this with "no whitespace". My editor did a number on these files cleaning up unnecessary whitespace.

Note 2: The reason this passed in CI is that my CI images do not have ISSM in them. They only have the minimal Baselibs. It might be a bit of work getting ISSM into them...I'll try and take a look.


agstub (approved):

lcandre2 (approved):

mathomp4:
Thanks for the approvals. I've merged this in.

Issue #1466, opened 2026-06-29
Intel and GNU L181 Failures with latest Moist GC


mathomp4:
My nightly runs are throwing failures with the current GEOSgcm_GridComp `develop`. I'm getting runtime failures with ifort Debug and ifx Debug and GNU Debug. BACM L72 runs with ifort, ifx, and GNU seem happy.

---

## ifort Debug

This is c24 L181 MERRA-2 Ocean AMIP

```
 Writing:     48 Slices to File:  1day.geosgcm_ocn.20000415_1030z.nc4

 GCM Date: 2000/04/15  Time: 11:00:00  Throughput(days/day)[Avg Tot Run]:        120.9         78.6         79.9  TimeRemaining(Est) 000:04:57   20.7% :  37.8% Mem Comm:Used  Wallclock: 202
6/06/29 02:10
forrtl: error (65): floating invalid
Image              PC                Routine            Line        Source
libpthread-2.31.s  000015192908A910  Unknown               Unknown  Unknown
libGEOSmoist_Grid  00001519546E511C  uwshcu_mp_compute        2273  uwshcu.F90
libGEOSmoist_Grid  0000151954679F27  uwshcu_mp_compute         365  uwshcu.F90
libiomp5.so        000015194B136C43  __kmp_invoke_micr     Unknown  Unknown
libiomp5.so        000015194B0BEFD1  Unknown               Unknown  Unknown
libiomp5.so        000015194B0BD716  __kmp_fork_call       Unknown  Unknown
libiomp5.so        000015194B07FC88  __kmpc_fork_call      Unknown  Unknown
libGEOSmoist_Grid  000015195466EBBE  uwshcu_mp_compute         284  uwshcu.F90
libGEOSmoist_Grid  000015195461392D  geos_uw_interface         548  GEOS_UW_InterfaceMod.F90
libGEOSmoist_Grid  000015195489AC29  geos_moistgridcom        6037  GEOS_MoistGridComp.F90
...
```

## ifx Debug

This is c24 L181 MERRA-2 Ocean AMIP

```
 Writing:     48 Slices to File:  1day.geosgcm_ocn.20000415_0630z.nc4

 GCM Date: 2000/04/15  Time: 07:00:00  Throughput(days/day)[Avg Tot Run]:        132.0         82.8         84.0  TimeRemaining(Est) 000:06:21   18.9% :  36.8% Mem Comm:Used  Wallclock: 202
6/06/29 05:42
[...]

PR #1483, opened 2026-07-17
Fixes #1482: spurious source of water

### LINKED ISSUE #1482: Spurious water sources in the Moist physics
The column integrated total water budget has a residual indicating an unphysical source of water within the MoistGridComp.  One part of this residual was traced to the clipping of ENT and SUB condensate tendencies after UW.  A second part is traced to GF, which is apparently generating negative total precipitation in some columns - when clipped to zero, this results in a spurious "source" of water, though not enough to fully account for the budget residual.

A fix is needed for the UW tendencies, the GF negative precipitation, and the remaining GF residual.

narnold1:
This initial commit fixes the spurious water from UW by simply allowing the SUB and ENT tendencies to create negative condensate, which is then "filled" by the FILLQ2ZERO routines immediately after.
Alternately we could modify the SUB and ENT tendencies as needed to prevent negative condensate, but would then need to modify QV to conserve the total water.


mathomp4:
Not sure if this is what it was intended for, but with this PR, I can now run in Debug (see #1466 )

ETA: Well, with Intel. I'll try GNU now.

ETA2: Apparently GNU is working without this. I'll try ifx.

ETA3: Never mind. ifx failed. It might have just been coincidence with ifort. Dang!

narnold1:
Spurious water from GF was traced to two errors:  (1) excessive precip evaporation, leading to negative surface precip, and (2) vertical smoothing of the final GF tendencies.  A fix for the excessive evaporation was added to the branch. The smoothing can be disabled by setting USE_SMOOTH_TEND: 0, but this needs further testing.

PR #1485, opened 2026-07-20
Fix UW shallow convection mumin NaNs

mathomp4:
This PR was from me asking GPT 5.6 Terra to look at the issues in #1466 

It first saw an OpenMP issue that, while a bug, is not the cause of #1466. It then looked at `compute_mumin2` and thought it could have odd behavior in some cases. So, I asked it for a `compute_mumin3` and this is what it came up with.

I can say that ifx passed Debug with this.

---

This draft addresses the UW shallow convection `mu` solve that could produce NaNs in edge cases.

Why this change
- `compute_mumin2` uses a Newton iteration starting from `mulow`.
- In some columns the iteration can move into a region where `exp(-mu**2)/erfc(mu)` becomes numerically unsafe, so the old solve can trigger NaN behavior instead of converging cleanly.
- That makes the limiter path fragile even though the intent is only to enforce the existing `ufrc`/`cbmf` constraints.

What changed
- Added `compute_mumin3`, a bounded bisection solve on `[mulow, 3]`.
- Replaced the risky solve path with the bounded solver when the `ufrclcl` limit needs to be enforced.
- Kept the same physical constraint logic; this is a numerical robustness fix, not a change to the underlying shallow convection formulation.
- Also includes the OpenMP tracer-workspace fix in `uwshcu.F90` so threaded execution uses private workspace arrays.

How `mumin3` differs from `mumin2`
- `mumin2`: unbounded Newton iteration, faster when well-behaved, but it can overshoot into a numerically unsafe region and produce NaNs.
- `mumin3`: bounded bisection, slower but more conservative, and it stays inside a finite interval so it cannot chase the solve into the underflow/0-0 regime.
- In normal cases the two should give very similar `mu` values.
- The difference should mainly show up in pathological columns where the old solver was unstable or non-convergent.
- In those cases, `mumin3` should return a finite bounded result instead of failing the limiter solve.

Validation note
- The main expected outcome is that NaN-producing cases now return a bounded, finite `mu` instead of failing the solve.

Acknowledgement
- This PR was developed with assistance from LLM-based tools, including GPT-5.6 Terra and GPT-5.4-mini.


narnold1 (approved):
Can confirm that this fixes the "compute_mumin" issue, and has minimal effect on science output (based on BOMEX SCM case and a short AMIP exp).

mathomp4:
As @narnold1 approves, I'll un-draft and @sdrabenh can test for his next update.

PR #1493, opened 2026-08-07
init do_shmidt to avoid crash

weiyuan-jiang:
It may be set to true by default, which leads to crash


gmao-rreichle (approved):

weiyuan-jiang:
it is replaced by PR #1494

PR #1495, opened 2026-08-10
produce tile nc4 file while skipping land params generation

weiyuan-jiang:


GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_latlon.py
```diff
@@ -25,10 +25,13 @@
     bin/mkLatLonRaster.x -x {NX} -y {NY} -b DE -p PE -t 0 {IMO} {JMO} >/dev/null
     bin/CombineRasters.x -f 0 -t {NT} DE{IMO}xPE{JMO} Pfafstetter >/dev/null
     bin/CombineRasters.x -t {NT} DC{IM}xPC{JM} DE{IMO}xPE{JMO}-Pfafstetter
-    setenv OMP_NUM_THREADS 1
```
weiyuan-jiang, line 28: This may have duplicated the computation

weiyuan-jiang:
There is nonzero diff for elevation variable. The orignal Nc4 file get raster_id from the memory but the TileFile_ASCII_to_Nc4 reads it from raster file, which leads to nonzero diff

biljanaorescanin:
For me for v14 for these resolutions 
`CF0180x6C_DE1440xPE0720`
`CF0180x6C_M6TP1440x1080 `
`CF0540x6C-SG001_CF0540x6C`  
`DC0576xPC0361_DE0360xPE0180`
`EASEv2_M09`
just tile option, so all under geometry, is **zero diff.** 

I also tried option of full run and that works as before. 

@weiyuan-jiang did you try to run `NL3` and got differences?  
Those are different due to compiler when they were created. They are quite old. We can't reproduce those. We even have note about that in the package. 

https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/899cb835825b9cc6b79a30db53992ef16f28dcb2/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/Raster/makebcs/make_bcs_questionary.py#L182-L191

weiyuan-jiang:
@biljanaorescanin @gmao-rreichle I have moved the lake_type to MAPL. So the PR needs this one from [ MAPL](https://github.com/GEOS-ESM/MAPL/pull/5323) . Also, please note that I changed -9999 to MAPL_UNDEFINED_INTEGER to make it consistent with other integer types. Would this change be acceptable ? @lcandre2

PR #1512, opened 2026-09-01
Fix GCC 15 use-association name collisions

mathomp4:
## Summary
GFortran 15 rejects component-local variables whose names collide with public entities use-associated from GEOSmoist_Process_Library. Intel compilers accepted these nonconforming declarations, but the collisions cause GCC to reject the local declarations and emit cascading implicit-type errors.

Rename the component-local configuration variables in the MGB2 interface and moist grid component while preserving their resource labels, defaults, and all runtime behavior.

## Verification
- make GEOSmoist_GridComp in build-Debug with GCC 15


dbarahon:
This works for now but needs to be fixed in the long run. If another component changes PDFSHAPE for example, MGB has no way of knowing it, unless it comes directly from AGCM.rc. That may prevent us from having a coupled configuration in the future.

dbarahon (approved):
Works but not optimal

PR #1513, opened 2026-09-02
Add LIS_GridComp: build LISF's LIS library natively under CMake

pchakraborty:
> [!NOTE]  
> Not to be merged at the moment, here only for comparison purposes.

## Summary

Introduces `GEOSlis_GridComp`, a MAPL/ESMF GridComp plugin that couples GEOS directly to LISF's core Fortran API (`LIS_domain_init`, `LIS_surfaceModel_run`, ...), bypassing LISF's own NUOPC coupling layer.

- `GEOS_LisGridComp.F90`: `SetServices`/`Initialize`/`Run`/`Finalize`/`Record`/`Refresh` entry points for the plugin.
- `LisFieldCatalog.F90`: self-contained static copy of `LIS_NUOPC_Gluecode`'s field stdName/stateName/units metadata and role (adImport/adExport/directConn) classification, so the plugin avoids building/linking LISF's NUOPC cap just for this static data.
- `LisFields.F90`: sorts the field catalog into import/export/internal `LIS_FieldSpec` arrays for `SetServices`.
- `LISF_cmake/CMakeLists.txt`: ports LISF's `LIS` library to build natively under CMake (replacing the previous `make nuopc` approach), parsing `Filepath`/`plugins.py` output into CMake source lists.

`GEOSlis_GridComp` is opt-in: `GEOSphysics_GridComp/CMakeLists.txt` only adds it to `alldirs` when configured with `-DBUILD_LIS=ON`, so existing builds are unaffected by default.

## Test

- Verified builds with `-DBUILD_LIS=ON` under `ifort-stack`.


GEOSagcm_GridComp/GEOSphysics_GridComp/LIS_GridComp/LISF_cmake/CMakeLists.txt
```diff
@@ -0,0 +1,208 @@
...
+#
+# This is a first-pass port and has NOT been build-verified in this
+# environment (no Fortran compiler available here) -- treat it as a
+# starting point to iterate on once it's actually compiled.
+#
+# BASEDIR must point to a Baselibs install (e.g. -DBASEDIR=$BASEDIR/Linux,
+# matching how Baselibs is found elsewhere in this build) -- it is used
+# here to locate HDF4/HDF-EOS2, which (unlike NetCDF/HDF5/ESMF) do not
+# already have CMake targets set up by @cmake/external_libraries/FindBaselibs.cmake.
+if (NOT BASEDIR)
```
mathomp4: Hmm. This will break in our Spack builds of GEOS.

mathomp4:
@pchakraborty You'll also want to add a `.gitignore` in `GEOSlis_GridComp/` so that no one accidentally commits `@LISF/` a la:

```gitignore
/@LISF
/LISF
/LISF@
```

pchakraborty:
> @pchakraborty You'll also want to add a `.gitignore` in `GEOSlis_GridComp/` so that no one accidentally commits `@LISF/` a la:
> 
> ```gitignore
> /@LISF
> /LISF
> /LISF@
> ```

Done https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/1513/commits/ff50b3f8c854c82d5deb5f0ef89c16eee020a3da

GEOSagcm_GridComp/GEOSphysics_GridComp/LIS_GridComp/LISF_cmake/CMakeLists.txt
pchakraborty: Not sure what to do here. The good news is that by default LIS is not built, and the integration at the moment is only to enable the LIS developers to try out running as part of GEOS. Maybe we can tackle this issue later?

mathomp4: I mean, I think you can remove this bit. I mean, ESMA_cmake should handle Baselibs vs Spack for everyone.

PR #1515, opened 2026-09-03
[DSL | Fix] Pin `ndsl` in pyproject to 2026.08.00 for CI

FlorianDeconinck:
Fix `ci` which was pulling from top of `develop` triggering failure on deprecated stencil name


twicki (approved):
Makes sense this way

Issue #1522, opened 2026-09-17
Possible uninitialized array in ISSM GridComp


mathomp4:
In looking at a weird "failure" in the nightly LDAS tests when using `ifx` when `ICETHICK` changed answers, @biljanaorescanin suggested maybe there might be an uninitialized array.

So I had GPT 5.6 Terra take a look and it found one thing (see below). @agstub could you take a look? It's possible we never hit this path so it's moot, but just in case...

## From Terra

In `GEOS_ISSMGridComp.F90`, the `ZEROS` buffer is allocated for the no-valid-restart bootstrap path:

```fortran
allocate(ZEROS(num_nodes))
...
call RunISSM(ISSM_DT_DP, c_loc(ZEROS), c_loc(GEOS_RESTARTS))
```

The buffer is documented as “zero input for bootstrapping,” but Fortran allocate does not initialize numeric storage. Consequently, `ZEROS` contains indeterminate memory unless it is explicitly assigned.

This matters because the ISSM C++ bridge reads every value of the passed forcing array and uses it as the surface mass-balance forcing:

```c++
smb_forcing[iv] = gcm_forcings[lid + f * total_nodes];
element->AddInput(SmbMassBalanceEnum, smb_forcing, P1Enum);
```

Therefore, on the bootstrap path, ISSM receives arbitrary SMB values rather than zero forcing. The ISSM solve then produces outputs, including thickness, which GEOS immediately unpacks and exports as `ICETHICK`. The arbitrary forcing can thus cause run-to-run differences in initial ice thickness, depending on the contents of newly allocated memory.

Initialize the buffer immediately after allocation:

```fortran
allocate(ZEROS(num_nodes)
ZEROS(:) = 0.0_dp
```

This makes the actual behavior match the existing comment and ensures that bootstrap results are deterministic with respect to the intended zero-forcing input.

@gmao-rreichle @lcandre2

PR #1523, opened 2026-09-17
fix ISSM initialization

biljanaorescanin:
Addresses https://github.com/GEOS-ESM/GEOSgcm_GridComp/issues/1522  by explicitly initializing ZEROS after allocation, ensuring zero forcing is passed to Run ISSM on the ISSM bootstrap path.

Zero-diff for GCM configuration because we don't run with ISSM.


gmao-rreichle (approved):
Thanks, @mathomp4 and @biljanaorescanin, for tracking this down.  Seems like this PR is needed even if it doesn't fix the LDAS test problem (although it looks like it would). 
@sdrabenh:  How soon could we get this merged into develop?  We'll need a new GCM GC release eventually; for now a merge into develop would be most helpful.  Thanks!

mathomp4 (approved):

mathomp4:
> Thanks, @mathomp4 and @biljanaorescanin, for tracking this down. Seems like this PR is needed even if it doesn't fix the LDAS test problem (although it looks like it would). @sdrabenh: How soon could we get this merged into develop? We'll need a new GCM GC release eventually; for now a merge into develop would be most helpful. Thanks!

I think this is pretty safe to pull in even now. As @biljanaorescanin says it doesn't affect GEOSgcm and it seems a good fix.

I'll merge as it looks like @sdrabenh is out of office today. 

I guess we'll see what happens with the LDAS tests tonight!

gmao-rreichle:
Thanks for the prompt merge, @mathomp4.  It may make sense to run the ifx test now and rebase.  We don't know whether the existing baseline was using true zeros or random bits.  By rebasing now, we should see a successful comparison tomorrow.   If we don't rebase, the comparison may still fail tonight.

PR #1526, opened 2026-09-23
undef values if not issm tiles

weiyuan-jiang:
This PR addresses the issue https://github.com/GEOS-ESM/GEOSldas_GridComp/issues/207


biljanaorescanin:
I tested this change with `LDAS_GLOBALCS`.

The `tavg24_1d_issm_Nt` collection is zero diff before/after the PR as is now, as expected. In the `tavg24_2d_glc_Nx` collection, only one `ICETHICK` point changes (`max_absdiff = 0.25 m`); `ICESURF` and `ICEVEL` are zero diff.

I also checked the Himalaya region from issue # 207 (`70–105E, 25–40N`). 
Before the PR, `ICESURF` has 2201 grid points in this region: 858 are missing and 1343 are finite. 
All 1343 finite values are exactly zero.

The result is unchanged after this PR, so the zero valued `ICESURF` points outside the ISSM domain are still present.

The allocation change here still looks needed, but based on this test it does not appear to be sufficient by itself to resolve # 207.

biljanaorescanin:
I added one more commit to complete the fix 

Initializing the destination field to `MAPL_UNDEF` was not sufficient because `ESMF_FieldRegrid` defaults to `zeroregion=ESMF_REGION_TOTAL`, which resets the destination field to zero before regridding. I changed the regrid call to use `zeroregion=ESMF_REGION_SELECT` so unmapped points retain MAPL_UNDEF.

I validated this over the Himalaya region (70–105E, 25–40N). Before the fix, 1343 points had finite ICESURF values and all 1343 were exactly zero. After the change, there are no finite values in that region; all of those points are now missing as expected.

This now resolves the behavior reported in [issue 207]( https://github.com/GEOS-ESM/GEOSldas_GridComp/issues/207)

lcandre2:
@biljanaorescanin This looks really good! I think the only issue is what is going on with the two islands down by Antarctica. Are they ISSM tiles (e.g. a problem with the original ISSM grid) or are they not being masked properly? Or is it just the plotting?

biljanaorescanin:
I checked the two islands separately. They are not a plotting artifact. The blue points in the plot are glc ICESURF locations that changed from valid to missing, while the ISSM mesh coordinates are taken from tavg24_1d_issm_Nt. There are no ISSM mesh points in this region (60–80°E, 58–45°S: 0 ISSM points).

ICESURF_islands_around_70E

The two affected clusters are:

Cluster 15: 36 points | 67.5–71.5°E, 48.5–50.5°S | ISSM points nearby = 0
Cluster 14: 20 points | 72.0–74.5°E, 52.5–54.0°S | ISSM points nearby = 0

These correspond to the Kerguelen and Heard Island regions. **They are LandIce locations outside the ISSM mesh**, so the previous zero valued ICESURF values there were not valid ISSM data. With the fix, those points are now changed to missing as intended.

biljanaorescanin:
I think I finally understand what you were asking, you meant the two islands that are still visible in the `After `panel, not why those points were removed?

I checked this more closely. The island outlines are still visible because the coastline is plotted independently of the ICESURF data. The islands themselves are not retaining ICESURF values after the fix; the corresponding data points are missing/white in the After field.

So this appears to be a plotting effect from the coastline overlay, rather than those locations being retained as valid ISSM data.

ICESURF_south_kept_removed_issm

